PR #28, opened 2020-03-20
Nonlinear Reservoir with Unit Tests

jdmattern-noaa:
This is the Nonlinear Reservoir with its associated unit tests.

## Additions

-

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

1.

- [ ] Linux


models/kernels/Nonlinear_Reservoir.hpp
```diff
@@ -0,0 +1,153 @@
...
+        this->outlets = outlets;
+	  
+        //Ensure that reservoir outlet activation thresholds are sorted from least to greatest height
+        sort(this->outlets.begin(), this->outlets.end(), [](const Reservoir_Outlet &left, const Reservoir_Outlet &right)
+        {              
+            return left.activation_threshold_meters < right.activation_threshold_meters;
+        });
+
+        //Ensure that the activation threshold is less than the maximum storage
+        for (auto& outlet : this->outlets) //pointer to outlets
```
hellkite500: Since this is sorted now, you can simply look at at the last outlet and see if it is greater than max.  If you want to know more than one, just loop backwards until you are less than max.

models/kernels/Nonlinear_Reservoir.hpp
```diff
@@ -0,0 +1,153 @@
...
+        state.current_storage_height_meters += in_flux_meters_per_second * delta_time_seconds;
+	
+        for (auto& outlet : this->outlets) //pointer to outlets
+        {
+            //Update storage from outlet velocity multiplied by delta time.
+            state.current_storage_height_meters -= outlet.velocity_meters_per_second(parameters, state) * delta_time_seconds;
+        }
+
+        //If storage is less than minimum storage, correct to minimum storage.
+        if (state.current_storage_height_meters < parameters.minimum_storage_meters)
```
hellkite500: Is this possible given that velocity_meters_per_second won't take water if it isn't available (i.e. storage > activation).  Also if this does happen and we do decide to correct, then we must keep track of how we corrected by, since this is arbitrary water created.  If we don't have it then we cannot properly mass balance.

models/kernels/Nonlinear_Reservoir.hpp
```diff
@@ -0,0 +1,153 @@
...
+            state.current_storage_height_meters = parameters.minimum_storage_meters;
+
+        //If storage is greater than maximum storage, correct to maximum storage.
+        if (state.current_storage_height_meters > parameters.maximum_storage_meters)
+            state.current_storage_height_meters = parameters.maximum_storage_meters;   
+            //TODO: Return extra storage to caller
```
hellkite500: this would be (current - max)/dt

I think passing in_flux by reference and taking from the flux and returning to it is the most straight forward way to handle this.  The caller must be prepared to check pass a flux, and then check that it has been exhausted (i.e. is 0) or has some left (i.e. > 0) and do something with it.

@donaldwj what do you think about this method?

models/kernels/Nonlinear_Reservoir.hpp
```diff
@@ -0,0 +1,153 @@
...
+
+//This class is for a single reservior outlet that holds the parameters a, b, and the activation threhold, which is
+//the height in meters the bottom of the outlet is above the bottom of the reservoir. This class also contains a 
+//function to return the velocity in meters per second of the discharge through the outlet. This function will only
+//run if the reservoir storage passed in is above the activation threshold.
+class Reservoir_Outlet
+{
+    public:
+
+    double activation_threshold_meters;
```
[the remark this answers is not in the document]
jdmattern-noaa: Okay I will change. Thanks.

models/kernels/Nonlinear_Reservoir.hpp
jdmattern-noaa: Ok thanks. Will change this too. @jdmattern-noaa changing.

models/kernels/Nonlinear_Reservoir.hpp
jdmattern-noaa: This check is after the reservoir calls the response on the vector of outlets, so it is checking the resulting storage after the response. I was trying various inputs to the unit tests and did come up with a negative storage. I don't know if the calculations should allow this to be possible though. I will go back and check through those now. I agree that the mass balance needs to be corrected from this. Fred mentioned some scenarios where it might be necessary to pass a flag out of the function in certain edge cases to go back and recalculate at a smaller timestep to account for these differences. I think you mentioned a similar type of idea previously. I think this would be a case where that would be helpful.

models/kernels/Nonlinear_Reservoir.hpp
jdmattern-noaa: Oh right thanks. I will change to (current - max)/dt. @jdmattern-noaa fixing.

models/kernels/Nonlinear_Reservoir.hpp
jdmattern-noaa: After looking into this further. I found the only time that the unit tests result in a negative storage is if the coefficient 'a' is > 1. I will email Fred to double check on this and include you, Nels. We might be able to throw out this below 0 check in the code if the algorithm doesn't allow it.

jdmattern-noaa: Fred says that the timestep is too long in this case. Back to the original plan then.

PR #34, opened 2020-03-26
Nonlinear Reservoir HYMOD Integration

jdmattern-noaa:
Integration of nonlinear reservoir into HYMOD.

## Additions

-

## Removals

-

## Changes
#	modified:   models/hymod/include/Hymod.h
#	modified:   models/kernels/Nonlinear_Reservoir.hpp

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

- [X ] PR has an informative and human-readable title

1.

- [ ] Linux


models/hymod/include/Hymod.h
```diff
@@ -127,12 +128,17 @@ class hymod_kernel
         // calculate et
         double et = calc_et(soil_m, et_params);
 
+        double groundwater_excess;  //excess water from groundwater reservoir that can be positive or negative
+        double excess_water;        // excess water from reservoir that can be positive or negative
+
         // get the slow flow output for this time - ks
-        double slow_flow = groundwater.response(slow, dt);
+        double slow_flow = groundwater.response_storage_meters(slow, dt, groundwater_excess);
+        slow_flow += groundwater_excess / dt;
```
hellkite500: I'm not sure I like this.  The groundwater reservoir response is slow flow, but the excess isn't nessicarly slow flow, it really should go back to whatever produced slow.  Also, what happens when excess < 0?  At some point, someone has to reconcile the supply/demand mismatch.

models/hymod/include/Hymod.h
```diff
@@ -127,12 +128,17 @@ class hymod_kernel
...
 
         for(unsigned long int i = 0; i < nash_cascade.size(); ++i)
         {
-            runoff = nash_cascade[i].response(runoff, dt);
+            runoff = nash_cascade[i].response_storage_meters(runoff, dt, excess_water);
+            runoff += excess_water / dt;
```
hellkite500: Similar to GW comment.

models/kernels/Nonlinear_Reservoir.hpp
```diff
@@ -107,14 +107,25 @@ class Nonlinear_Reservoir
...
 	
+        //Loop through reservoir outlets.
         for (auto& outlet : this->outlets) //pointer to outlets
         {
+            //Calculate outlet velocity.
+            outlet_velocity_meters_per_second = outlet.velocity_meters_per_second(parameters, state);
+
             //Update storage from outlet velocity multiplied by delta time.
-            state.current_storage_height_meters -= outlet.velocity_meters_per_second(parameters, state) * delta_time_seconds;
+            state.current_storage_height_meters -= outlet_velocity_meters_per_second * delta_time_seconds;
```
hellkite500: Really like the units in these variables, makes dimensional analysis pretty quick!

hellkite500:
Also, in a bit of haste I squashed and merged your old PR, and now I'm seeing the same commits present again.  To avoid having to squash and merge again, you should sync upstream and make a fresh branch for the new changes so we only have the new commits with new changes.

jdmattern-noaa:
> Also, in a bit of haste I squashed and merged your old PR, and now I'm seeing the same commits present again. To avoid having to squash and merge again, you should sync upstream and make a fresh branch for the new changes so we only have the new commits with new changes.

Yeah good point thanks. You showed that to me before, and I will do next time.

jdmattern-noaa:
> Would like to iterate on this just a little bit. Will get @donaldwj to take a look as well. I think the same time consideration we discussed in the LinearReservoir may apply here as well. May need to incorporate the same type of logic for the "internal dt as related to alpha (a)" in the outlet coefficients. In the general case with >1 outlet, I start to wonder if those should/need to be related, or if they can be independent. Thoughts?

Yeah Donald and I spoke about this earlier. He said that passing dt in at response nullifies the need for the internal dt. I will let him clarify.

models/hymod/include/Hymod.h
jdmattern-noaa: When excess < 0, then it would be subtracted. Do you have any suggestions on what to do with excess?

hellkite500:
What I'm not sure about is it we are adding/subtracting from the correct
place.

models/hymod/include/Hymod.h
ADHydro-devel: I think that if something happens where the storage goes negative, then it indicates a solution overshoot.   The standard solution is to check if the change in storage this time step is negative and greater than the current storage, then we would make the change in storage equal to the current storage, current storage would be come zero, and we probably want to throw a warning that it happened.

hellkite500:
Replaced by #35.  Will close once that PR is resolved.

models/hymod/include/Hymod.h
jdmattern-noaa: Thanks Fred. I accidentally passed over this comment when you made it. We added boundary checks for min/max and warnings and adjustments to storage to zero if below min and max if exceeding max. We drain the reservoir to min storage in that case by sending the remaining storage to output. We send to excess water back to the flux when over max.

hellkite500:
Done in #35.

PR #36, opened 2020-03-26
Geojson

christophertubbs:
Added parsing capabilities for Geojson documents

## Additions

- Handling for Geojson documents

## Testing

1. Run the test_geojson test or test_all to test the functionality

## Notes

- Handling for GeometryCollections is not complete
- Build issues were keeping everything from working if the implementation code wasn't in the header. I believe those issues have been fixed, so the implementation code will need to migrate from .hpp to .cpp

## Todos

- Complete the documentation

PR #37, opened 2020-03-27
Rebased geojson

robertbartel:
Added parsing capabilities for Geojson documents (this is a rebased version of the changes proposed in PR #36 along with a few additional directly related improvements).

## Additions

- Handling for Geojson documents

## Testing

1. Run the `test_geojson`, `test_unit`, or `test_all` target to test the functionality.  Keep in mind however that, except for `test_geojson`, other unrelated tests will also run.

## Notes

- Handling for GeometryCollections is not complete
- In PR #36, build issues were keeping everything from working if the implementation code wasn't in the header. That should be addressed for the most part in this PR.

## Todos

- Complete the documentation


src/geojson/Makefile
```diff
@@ -0,0 +1,332 @@
+# CMAKE generated file: DO NOT EDIT!
```
hellkite500: Generated files should not be indexed and versioned.

hellkite500: Similarly,  src/geojson/libgeojson.a should get removed.

robertbartel:
Indeed, I didn't notice those had snuck in.  I'll remove them.  I think I also want to fix a couple other things also ...

robertbartel:
Ok @hellkite500 and @christophertubbs:  I have those cleaned up, and have actually moved code definitions to code files for the rest of the classes.  I may not have been entirely consistent about how I did this with constructors, so those may need to be modified (as part of this PR, or later).

include/geojson/Feature.hpp
```diff
@@ -0,0 +1,72 @@
...
+            );
+
+            Feature(std::vector<JSONGeometry> new_geometry_collection,
+                std::vector<double> new_bounding_box = std::vector<double>(),
+                property_map new_properties = property_map()
+            );
+
+            Feature(boost::property_tree::ptree &tree);
+
+            Feature(const Feature &feature);
```
hellkite500, line 43: If we are going to have a copy constructor, I would advocate for a swap function as well.  Copy Swap is a nice pattern for implementing various semantics, and might be a good idea to put in as many places as possible.  If performance issues arise we can profile and tackle them on an individual basis.

include/geojson/JSONGeometry.hpp
```diff
@@ -0,0 +1,203 @@
...
+            static JSONGeometry of_point(pt::ptree &tree);
+
+            static JSONGeometry of_linestring(two_dimensional_coordinates &coordinates);
+
+            static JSONGeometry of_linestring(pt::ptree &tree);
+
+            static JSONGeometry of_polygon(three_dimensional_coordinates &coordinates);
+
+            /**
+             * Generate a polyginal geometry from the details of coordinates passed in shaped like:
```
hellkite500, line 138: polygonal?

src/geojson/Feature.cpp
```diff
@@ -0,0 +1,209 @@
...
+            if (property.second.data() == "true") {
+                properties.emplace(key, JSONProperty(key, true));
+                continue;
+            }
+            else if (property.second.data() == "false") {
+                properties.emplace(key, JSONProperty(key, false));
+                continue;
+            }
+
+            for(int character_index = 0; character_index < value.length(); character_index++) {
```
hellkite500, line 110: I would be inclined to move this character parsing to a private function by itself.

include/geojson/JSONGeometry.hpp
```diff
@@ -0,0 +1,203 @@
...
+
+            bg::model::multi_linestring<linestring_t> as_multilinestring() const;
+
+            bg::model::multi_polygon<polygon_t> as_multipolygon() const;
+        private:
+            JSONGeometryType type;
+            linestring_t linestring;
+            coordinate_t point;
+            polygon_t polygon;
+            bg::model::multi_point<coordinate_t> multi_point;
```
hellkite500, line 197: Need to think about a better way to dynamically type the boost geom model.

hellkite500:
@christophertubbs @robertbartel I wanted to give you a chance to see the comments in the review I posted.  I want to merge this as a starting point to iterate on unless you think there is anything we should address out of those comments

Issue #48, opened 2020-04-10
Build tests failed.


ZacharyWills:
Short description explaining the high-level reason for the new issue.

## Current behavior
Build tests exits Code 2

## Expected behavior
All Tests: Build and Pass

## Steps to replicate behavior (include URLs)

1.https://github.com/ZacharyWills/ngen/runs/577718969

hellkite500:
@jdmattern-noaa looks like a few references to refactored variables didn't get updated.

jdmattern-noaa:
Okay I will fix thanks. @jdmattern-noaa

jdmattern-noaa:
I will fix over the weekend or Monday morning.

ZacharyWills:
failed new merge also, could be the same cause:
https://github.com/ZacharyWills/ngen/actions/runs/77295822

ZacharyWills:
test badge:
![C/C++ CI](https://github.com/ZacharyWills/ngen/workflows/C/C++%20CI/badge.svg)

ZacharyWills:
Appears to be passing now. We can close.
https://github.com/ZacharyWills/ngen/runs/583343310

Also badge seems to work.

PR #51, opened 2020-04-13
Fixing minor Hymod.h bugs

robertbartel:
Looks like some minor issues were introduced into Hymod.h at some point, causing it to fail to compile.  These look like they were related to variable name refactoring.  

Several were obvious.  One was slightly less clear, as it introduced two usages (i.e. declaration and assignment) of the same variable name.  This was the `slow_flow_meters_per_second` variable in the `run` function.  I believe I have fixed this correctly, and unit tests now pass, but it is worth perhaps more careful consideration during review.


hellkite500:
Closes #48

PR #58, opened 2020-04-17
Updates to the GeoJSON library

christophertubbs:
I updated some of the interface for the geojson library to keep things a little simpler (oddly enough), added simpler graphing capabilities for the features and feature collections, fixed issues with instance lifespans, added support for the visitor pattern, and replaced the JSONGeometry class with a boost variant of the different geometries.

## Notes

- Code had been separated out to the proper source files as I was working on it. To get everything to mesh with my current development, I had to move code back up to the header, to the code base's detriment. I'm going to move the code back down into the .cpp files, I just needed to get this in and going first.

## Todos

- Separate the code out from the headers and back to the source
- Add documentation
- Like the JSONGeometry class, JSONProperty probably needs to be replaced by a variant as well

1. test_geojson should test all of the functionality


robertbartel:
@christophertubbs, I think you need to rebase your _master_ branch.  This PR as-is includes commits for which the contents are already in the main repo's _master_.  E.g., your "Initial Geojson work" commit, `60d0466`.

christophertubbs:
Any tips for the rebase? It's not in my normal workflow and, when I performed it, it made no difference.

robertbartel:
@christophertubbs, have a look at the [Keeping Forks Up to Date](docs/GIT_USAGE.md#keeping-forks-up-to-date) section in the `docs/GIT_USAGE.md` file.  I suspect the problem is that you haven't fetched the recent upstream changes.  

The rebase itself may be messy at this point with several conflicts, but for that you'll just have to work through manually resolving things.  I suggest using the `-i` flag to do the rebase interactively.  Depending on how the rebase goes, you may want to drop certain commits with their changes already upstream, but you'll just have to see.  You also may need to end up starting and aborting the rebase a few times to figure out the best way to clean up stuff. 

Once you have your local copy in a good state, I think you'll probably also see Git complain when you try to push to your fork.  I expect the best way to deal with that is to use `-f` to force your fork's current branch to be overwritten.  In its current state, I don't think there is any way around that.

christophertubbs:
👍

christophertubbs:
Alright, rebasing was a miserable process, but I think we're good.

robertbartel:
As we are likely to use other workflows/jobs, I suggest either changing the file name of `main.yml` to reflect that it is specifically is related to building documentation, or renaming the workflow to be more general.

christophertubbs:
Renamed. It was only named that so I could work through my automation experiments.

include/geojson/features/FeatureBase.hpp
```diff
@@ -0,0 +1,471 @@
...
+            FeatureType type;
+            ::geojson::geometry geom;
+            std::vector<::geojson::geometry> geometry_collection;
+
+            PropertyMap properties;
+            std::vector<double> bounding_box;
+            PropertyMap foreign_members;
+            std::string id;
+
+            std::vector<FeatureBase*> upstream;
```
hellkite500, line 465: Are upstream and downstream ordered in the vector (i.e. downstream[0] is first downstream, downstream[1] is two hops away?  If not, a map might be a better alternative, the lookups in get_(up/down)stream would be much quicker.  If order is important though, the vector might be the best.

src/geojson/CMakeLists.txt
```diff
@@ -1,11 +1,12 @@
...
-
+add_library(geojson STATIC
+        JSONGeometry.cpp
+        JSONProperty.cpp
+        FeatureCollection.cpp
+        )
 add_library(NGen::geojson ALIAS geojson)
 target_include_directories(geojson PUBLIC
-        ${PROJECT_SOURCE_DIR}/include/geojson
+        ../../include/geojson
```
hellkite500, line 9: This should probably go back to working from ${PROJECT_SOURCE_DIR}

src/geojson/Makefile
```diff
@@ -0,0 +1,332 @@
+# CMAKE generated file: DO NOT EDIT!
```
hellkite500, line 1: Generated files shouldn't be indexed and versioned.

test/geojson/JSONGeometry_Test.cpp
```diff
@@ -174,30 +122,23 @@ TEST_F(JSONGeometry_Test, polygon_test) {
...
     boost::property_tree::json_parser::read_json(stream, tree);
-    geojson::JSONGeometry geometry = geojson::JSONGeometry::of_polygon(tree);
+    auto read_geometry = geojson::build_geometry(tree);
+    auto geometry = geojson::get_shape<geojson::polygon_t>(read_geometry);
 
-    ASSERT_EQ(geometry.get_type(), geojson::JSONGeometryType::Polygon);
-    ASSERT_EQ(geometry.as_polygon().outer().size(), 5);
-    ASSERT_EQ(geometry.as_polygon().inners().size(), 1);
-    ASSERT_EQ(geometry.as_polygon().inners()[0].size(), 4);
+    ASSERT_EQ(geometry.outer().size(), 5);
```
hellkite500, line 128: These calling interfaces are SOOOO much simpler, good work!

include/geojson/features/FeatureBase.hpp
christophertubbs: The order isn't really important; if you want to get a feature that's two hops away, you'd go to downstream[0]->downstream[0].

src/geojson/CMakeLists.txt
christophertubbs: This was a merge issue; I'll address it.

src/geojson/Makefile
christophertubbs: Yet ANOTHER merge issue.

doc/programming_standards.md
```diff
@@ -93,7 +93,15 @@ int px;
 ```
 
 Control structures should _also_ have descriptive names, even though variables like
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
christophertubbs: Welp. That's a merge issue.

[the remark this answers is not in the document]
christophertubbs: Should be fixed now; I have no idea how a merge conflict got repeated and NESTED in there.

src/geojson/CMakeLists.txt
christophertubbs: Alright, should be fixed

src/geojson/Makefile
christophertubbs: Should be gone. Also removed the binaries.

include/geojson/features/FeatureBase.hpp
hellkite500: So its link-list like.  a map of id -> feature would be appropriate.

PR #61, opened 2020-04-22
IN PROGRESS: Tshirt/create class

robertbartel:
Current progress creating the class for the T-shirt model, as a PR for public accessibility/visibility.

Work is very much incomplete and still in progress.  This should almost certainly not be merged/approved as-is.


robertbartel:
It isn't perfect, but the bulk of the class has been created.  

One particular thing that needs attention still is that the groundwater reservoir `Qgw` value is being calculated raw from the equation given in the _doc/T-shirt_model_description.pdf_ document.  It is specifically:

`double Qgw = params.Cgw * ( exp(params.expon * state.Sgw / params.Sgwmax) - 1 );`

The other reservoir flow equations are of the form:

`Q = C * (S - S_fc) / (S_max - S_fc)`

The latter lends itself to inferring the constructor parameters for our reservoir and reservoir outlet classes, but it is not apparent to me how to do this with the former.

include/realizations/catchment/Simple_Lumped_Model_Realization.hpp
```diff
@@ -4,6 +4,7 @@
 #include <HY_CatchmentArea.hpp>
 #include "LinearReservoir.hpp"
 #include "hymod/include/Hymod.h"
+#include "tshirt/include/Tshirt.h"
```
donaldwj, line 7: You should not include tshirt code in this file. Make a class like this don't use this file.

robertbartel: Thanks, @donaldwj.  Seems I did that several weeks ago, and I'm not exactly sure why.

models/tshirt/include/Tshirt.h
```diff
@@ -0,0 +1,284 @@
...
+            // init subsurface percolation flow outlet
+            subsurface_outlets[perc_outlet_index] = Reservoir_Outlet(params.satdk * params.slope, 1.0, Sfc, max_perc_flow);
+
+            Nonlinear_Reservoir subsurface_reservoir(0.0, params.depth, state.Ss, subsurface_outlets);
+
+            double subsurface_excess;
+            subsurface_reservoir.response_meters_per_second(subsurface_infiltration_flux, dt, subsurface_excess);
+
+            // lateral subsurface flow
+            double Qlf = subsurface_reservoir.velocity_meters_per_second_for_outlet(lf_outlet_index);
```
jdmattern-noaa, line 215: All of the usage of the subsurface reservoir look good to me up until this point. The only concern here is that velocity_meters_per_second_for_outlet uses the updated state of current_storage_height_meters after the response_meters_per_second is called. Therefore, the velocities that are returned reflect the end velocity after the drainage and not the velocity during the drainage. The velocity after drainage might actually be what is needed though. I am a unsure from the diagram and this implementation which velocity is actually needed.

If the velocity during the drainage is needed, I can create a small separate function that returns the velocities that were used in the drainage/response_meters_per_second, and the Reservoir_Outlet class would also maintain that as a state variable also.

jdmattern-noaa:
> It isn't perfect, but the bulk of the class has been created.
> 
> One particular thing that needs attention still is that the groundwater reservoir `Qgw` value is being calculated raw from the equation given in the _doc/T-shirt_model_description.pdf_ document. It is specifically:
> 
> `double Qgw = params.Cgw * ( exp(params.expon * state.Sgw / params.Sgwmax) - 1 );`
> 
> The other reservoir flow equations are of the form:
> 
> `Q = C * (S - S_fc) / (S_max - S_fc)`
> 
> The latter lends itself to inferring the constructor parameters for our reservoir and reservoir outlet classes, but it is not apparent to me how to do this with the former.

I would check with Fred on this one. If the equation for groundwater is indeed the former, then I could subclass the Reservoir_Outlet class into the 2 separate implementations.

models/tshirt/include/Tshirt.h
```diff
@@ -0,0 +1,284 @@
...
+            //double Qgw = params.Cgw * ( exp(params.expon * state.Sgw / params.Sgwmax) - 1 );
+
+            // Given the equation:
+            //      double Qgw = params.Cgw * ( exp(params.expon * state.Sgw / params.Sgwmax) - 1 );
+            // The max value should be when Sgw == Sgwmax, or ...
+            double max_gw_velocity = params.Cgw * ( exp(params.expon) - 1 );
+            // TODO: verify activation threshold
+            Nonlinear_Reservoir groundwater_res(0, params.Sgwmax, state.Sgw, params.Cgw, 1, 0, max_gw_velocity);
+            // TODO: what needs to be done with this value?
+            double excess_gw_water;
```
jdmattern-noaa, line 250: After calling groundwater_res, the excess_gw_water should be added to the fluxes.surface_runoff.

We added the excess groundwater to runoff in hymod, and I assume it would be added to runoff here too.
https://github.com/NOAA-OWP/ngen/blob/master/models/hymod/include/Hymod.h#L149

        //TODO: Review issues with dt and internal timestep
        runoff_meters_per_second += groundwater_excess_meters / dt;

The excess is in meters so be sure to divide by dt.

hellkite500: Yes, essentially if there is any excess it has to go back up the stack and get added to the bucket of water that provided the original water to that call.

hellkite500:
@robertbartel Showing conflicts in test/CMakeLists.txt.  This may be due to a recent merge of other code. Can you take a look?

models/tshirt/include/Tshirt.h
robertbartel: @jdmattern-noaa, I don't see how that fits into things in the case of T-shirt, whereas it's more clear in Hymod, but it seems sensible.  I'll make the change, but I'll need to confirm whether that belongs in the runoff amounts before or after GIUH calculations are performed.

hellkite500: @fred-ogden what to do with potentially excess water from the groundwater reservoir?

models/tshirt/include/Tshirt.h
```diff
@@ -0,0 +1,289 @@
...
+
+    /*!
+     * Tshirt state structure
+     *
+     * This structure provides storage for the state used by Tshirt hydrological model at a particular time step.
+     */
+    struct tshirt_state
+    {
+        // TODO: confirm this is correct
+        double Ss;           //!< current water storage in soil column nonlinear reservoir
```
hellkite500, line 72: ultimately would like to see these follow a more verbose convention, i.e. soil_storage_meters, ground_water_storage_meters, nash_cascade_storeage_meters (or whaterver appropriate units these variables are modeling)

models/tshirt/include/Tshirt.h
```diff
@@ -0,0 +1,289 @@
...
+    /*!
+     * Tshirt flux structure
+     *
+     * This structure provides storage for the fluxes generated by Tshirt at any time step
+     */
+    struct tshirt_fluxes
+    {
+        double surface_runoff;  //!< Direct surface runoff
+        double Qgw;             //!< Groundwater flow from groundwater reservoir to channel flow
+        double Qperc;           //!< Percolation flow from subsurface to groundwater reservoir
```
hellkite500, line 91: units need documenting, minimally in the comments (is Qperc m/s -> meters_per_sec)

models/tshirt/include/Tshirt.h
```diff
@@ -0,0 +1,289 @@
...
+        tshirt_fluxes(double q_gw, double q_perc, double q_lf, double runoff, double et_loss) : Qgw(q_gw),
+                                                                                                Qperc(q_perc),
+                                                                                                Qlf(q_lf),
+                                                                                                surface_runoff(runoff),
+                                                                                                et_loss(et_loss) {
+
+        }
+    };
+
+    // TODO: consider combining with or differentiating from similar hymod enum
```
hellkite500, line 104: common error represntations, i.e. mass_balance_error, can be reused and probably should be.  Unique conditions per model can can be enumerated in the model namespace

hellkite500: This can be done later

models/tshirt/include/Tshirt.h
```diff
@@ -0,0 +1,289 @@
...
+        /*!
+         * Calculate the height above water table based on the given Tshirt parameters
+         *
+         * Calculate the height above water table based on known constants and the particular `alpha` value given for
+         * relative soil suction head, as provided in the Tshirt parameters.
+         *
+         * @param params
+         * @return
+         */
+        static double calc_Hwt(const tshirt_params& params)
```
hellkite500, line 145: function names should be more descriptive -> height_above_water_table_meters ???

models/tshirt/include/Tshirt.h
```diff
@@ -0,0 +1,289 @@
...
+        {
+            // H_wt = alpha_fc * H_atm ; H_atm = P_atm / gamma ; P_atm = 101,300 [Pa] ; gamma = 9,810 [N m^-3] ('very nearly')
+            return params.alpha_fc * (101300 / 9810);
```
hellkite500, line 148: Should define a constants header and macro these constants

models/tshirt/include/Tshirt.h
```diff
@@ -0,0 +1,289 @@
...
+        //! run one time step of tshirt
+        static int run(
+                double dt,
+                tshirt_params params,        //!< static parameters for tshirt
+                tshirt_state state,          //!< model state
+                tshirt_state& new_state,     //!< model state struct to hold new model state
+                tshirt_fluxes& fluxes,       //!< model flux object to hold calculated fluxes
+                double input_flux_meters,          //!< the amount water entering the system this time step
+                // TODO: should/can this be a smart pointer?
+                giuh_kernel* giuh_obj,       //!< kernel object for calculating GIUH runoff from subsurface lateral flow
```
hellkite500, line 184: The catchment realization should hold the giuh kernel (as a smart pointer) and manage the giuh memory.  The realization is what composes these kernels/models together into a realized model.

models/tshirt/include/Tshirt.h
```diff
@@ -0,0 +1,289 @@
...
+
+            // Note this surface runoff value has not yet performed GIUH calculations
+            double surface_runoff, subsurface_infiltration_flux;
+
+            Schaake_partitioning_scheme(dt, calc_Cschaake(params), column_total_soil_moisture_deficit, input_flux_meters,
+                    &surface_runoff, &subsurface_infiltration_flux);
+
+            double Sfc = calc_Sfc(params, state);
+
+            vector<Reservoir_Outlet> subsurface_outlets;
```
hellkite500, line 197: Does this need to be recreated each time step?  Or can the outlets be defined, initialized, and reused?

models/tshirt/include/Tshirt.h
```diff
@@ -0,0 +1,289 @@
...
+
+            // Keep track of the indexes of the specific outlets for later access
+            int lf_outlet_index = 0;
+            int perc_outlet_index = 1;
+
+            // TODO: confirm these are correct
+            // Max transmissivity
+            double max_lateral_flow = params.satdk * params.multiplier * params.Ssmax;
```
hellkite500, line 205: again, anything derived strictly from params should be relatively static, can be defined at initilization and resued by the instance of the model class

models/tshirt/include/Tshirt.h
robertbartel: We had a discussion about this at the end of the standup.  If a later reservoir becomes or approaches being full, that should restrict the flow from the earlier source.  Here, e.g., if the GW reservoir is full, that impose a limit on the max allowed `Qperc`.  If both the GW and the subsurface reservoirs are full, then this limits the influx into the subsurface reservoir, and the excess water gets attributed to surface runoff.

hellkite500: Hmm....a stack of water!

models/tshirt/include/Tshirt.h
robertbartel: @hellkite500, so only the "raw" surface runoff needs to be present and utilized inside the Tshirt model kernel, with the GIUH calculations happening completely separately from `tshirt_kernel.run(...)`?  

This seems sensible, but the card for creating the Tshirt class suggests having the GIUH kernel directly included (and, for that matter, `tshirt_kernel` _requires_ reservoir kernel class objects).  I.e., in some respects, the separation of concerns between the Tshirt model kernel class versus the Tshirt model realization is unclear.

More practically, it would also hinder a spot I see to employ some concurrency inside the `run` function in the future, though that could be made up for in various ways and may not be a priority.

models/tshirt/include/Tshirt.h
robertbartel: This again raises the question of how things should be separated in the design.  Should reservoirs and reservoir outlets be created in/owned by the Tshirt realization object, then just passed to a call `tshirt_kernel.run(...)`?

jdmattern-noaa: Reservoirs and outlets can be defined, initialized, and reused. With my limited understanding, I would vote for created in/owned by the Tshirt realization object as I think it would make things more modular.

models/tshirt/include/Tshirt.h
jdmattern-noaa: I am still caught up on this thought. Seems to me that things could get kind of complicated with having to go back and recalculate the Qperc to reconcile the difference with at which exact point in the timestep the GW reservoir overflows. Could we just keep Qperc as originally calculated and update the subsurface reservoir state storage height with the GW excess? If the subsurface reservoir storage height then exceeds max after the update, then that subsurface reservoir's excess could be added to the runoff.

models/tshirt/include/Tshirt.h
jdmattern-noaa: After further discussion, we concluded it is the velocity during the drainage. I am updating the nonlinear reservoir code to appropriately return this. @jdmattern-noaa

models/tshirt/include/Tshirt.h
robertbartel: I've created `include/utilities/Constant.h` for this purpose.  It can be moved or renamed if necessary.

models/tshirt/include/Tshirt.h
robertbartel: I have done this with this and a few other such derived values.

models/tshirt/include/Tshirt.h
robertbartel: I've removed a few functions like this that were initially placeholders but turned out to be one-liners used just once.  I have refactored the others with clearer names.

models/tshirt/include/Tshirt.h
robertbartel: I've renamed these members, with new names now including the relevant units.

models/tshirt/include/Tshirt.h
robertbartel: I have renamed most of these, but there are a handful (e.g., `Klf`) I haven't.  I think it makes sense for certain constant parameters to be left like this and/or reflect external conventions, but I won't mark this conversation resolved just yet.

robertbartel:
@hellkite500, FYI, I have changes we talked about yesterday added to the branch for this PR, including the realization implementation.

robertbartel:
Once PR #72 gets approved, I'll need to rebase the changes here and then update the Tshirt class(es) to incorporate using the new reservoir outlet types.

hellkite500:
@robertbartel just merged #72.

include/core/catchment/giuh/GIUH.hpp
```diff
@@ -0,0 +1,128 @@
...
+                cdf_ordinate_times_seconds.push_back(time_for_ordinate);
+
+                // Find index 'i' of largest CDF time less than the time for the current ordinate
+                // Start by getting the index of the first time greater than time_for_ordinate
+                int cdf_times_index_for_iteration = 0;
+                while (this->cdf_times[cdf_times_index_for_iteration] < cdf_ordinate_times_seconds.back()) {
+                    cdf_times_index_for_iteration++;
+                }
+                // With the index of the first larger, back up one to get the last smaller
+                cdf_times_index_for_iteration--;
```
hellkite500, line 62: add some checks for cdf_times_for_iteration >= 0, < cdf_times.size()

include/core/catchment/giuh/GIUH.hpp
```diff
@@ -0,0 +1,128 @@
...
+
+                // Finally, increment the ordinate time based on the regularity
+                time_for_ordinate = cdf_ordinate_times_seconds.back() + cdf_regularity_seconds;
+            }
+
+            // Finally, the last ordinate time gets set to have everything
+            cdf_ordinate_times_seconds.push_back(time_for_ordinate);
+            regularized_cdf_ordinates.push_back(1.0);
+
+            // Also, now calculate the incremental values between each ordinate step
```
hellkite500, line 82: Might be worth pulling the interpolation into a publicly exposed function so the GIUH could switch time bases dynamically.

models/kernels/Nonlinear_Reservoir.hpp
```diff
@@ -1,3 +1,6 @@
+#ifndef NONLINEAR_RESERVOIR_HPP
```
hellkite500, line 1: I think this was merged in a prior PR so this is causing a merge conflict.

hellkite500: @robertbartel can you check on this conflict then I'll merge

robertbartel: I've done some rebasing to get the exponential reservoir outlet changes incorporated, which should have also taken care of this.

PR #62, opened 2020-04-23
GeoJSON Interface additions

christophertubbs:
I added some helpful changes to FeatureBase and FeatureCollection and disabled the documentation workflow.

It looks like a bunch of redundant commits, but I was having a bunch of rebasing/pushing/pulling issues


christophertubbs:
The action is failing at the moment because github-actions[bot] doesn't have the right to push the documentation to gh-pages. I'll probably take out the push action, but we won't be able to automagically update the docs. I think a PR can be set up, though. I'll need to play with it.

PR #73, opened 2020-05-18
Updated GeoJSON Interface

christophertubbs:
I added some standard collection functions, fixed missing logic for assigning IDs, and added some equality overloads in order to make the GeoJSON interface a little easier to work with.

### DIFF
--- a/src/geojson/FeatureCollection.cpp
+++ b/src/geojson/FeatureCollection.cpp
@@ -7,15 +7,75 @@ Feature FeatureCollection::get_feature(int index) const {
     return features[index];
 }
 
+int FeatureCollection::find(Feature feature) {
+    for(int feature_index = 0; feature_index < this->get_size(); feature_index++) {
+        if (*this->features[feature_index] == *feature) {
+            return feature_index;
+        }
+    }
+
+    return -1;
+}
+
+int FeatureCollection::find(std::string ID) {
+    for (int feature_index = 0; feature_index < this->get_size(); feature_index++) {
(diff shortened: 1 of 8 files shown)


christophertubbs:
Disregard for a bit; tried rebasing and it went waaaay wrong.

src/geojson/FeatureCollection.cpp
```diff
@@ -7,15 +7,75 @@ Feature FeatureCollection::get_feature(int index) const {
     return features[index];
 }
 
+int FeatureCollection::find(Feature feature) {
```
hellkite500, line 10: This could be written as
```
FeatureList::iterator it = std::find(features.begin(), feature.end(), feature)
return std::distance(feature.begin(), it);
```

src/geojson/FeatureCollection.cpp
```diff
@@ -7,15 +7,75 @@ Feature FeatureCollection::get_feature(int index) const {
...
+    for(int feature_index = 0; feature_index < this->get_size(); feature_index++) {
+        if (*this->features[feature_index] == *feature) {
+            return feature_index;
+        }
+    }
+
+    return -1;
+}
+
+int FeatureCollection::find(std::string ID) {
```
hellkite500, line 20: This would be a good place to keep an `unorded_map<std::string, int> index_by_id` and then just look up the vector index by ID.
If the memory overhead of the map is significant, a refresh of the overall storage mechanics might be in order.

Issue #80, opened 2020-05-20
Improve CSV Reader Class Handling of Missing Files


robertbartel:
Improve how the CSV reader class (and associated usages of it) handle cases when an expected file doesn't exist.  Right now the CSV reader itself doesn't fail, but in such cases it doesn't read anything either, leading to inconsistent behavior downstream in program execution.

A particular example was seen in unit testing for forcings.  If the unit test executable is run from the main directory, rather than when inside the build directory, then a relative path needed for the `ForcingTest.TestForcingDataRead` test will not be correct, leading to a crash of the unit test executable.

jdmattern-noaa:
The Forcing PR #83 was just integrated and addresses this with an error being thrown and halting the program if a CSV file is missing. This does not solve the part of that issue of the forcing unit test failing if the directory that the tests are called from is not the build directory and its dependencies on relative paths. A upcoming solution to this part of the issue would also address issue #69, where the problems with these relative paths is discussed.

jdmattern-noaa:
Issue #69 has been fixed and therefore fixes the rest of this issue.

Issue #69, opened 2020-05-13
Tests failing for last merge


ZacharyWills:
Short description explaining the high-level reason for the new issue.

## Current behavior
Tests run but fail 

[----------] 1 test from ForcingTest
80
[ RUN      ] ForcingTest.TestForcingDataRead
81
/home/Z/actions-runner/_work/_temp/1ac2add5-d50d-471a-994c-51763b638d59.sh: line 1:  5313 Segmentation fault      (core dumped) ./cmake_build/test/test_all
82
##[error]Process completed with exit code 139.

## Expected behavior
Tests should pass

## Steps to replicate behavior (include URLs)

https://github.com/ZacharyWills/ngen/runs/671741082

ZacharyWills:
Still getting seg fault

https://github.com/ZacharyWills/ngen/runs/686388461?check_suite_focus=true

ZacharyWills:
<img width="1793" alt="Screen Shot 2020-05-15 at 1 09 05 PM" src="https://user-images.githubusercontent.com/60660437/82247665-64526480-990c-11ea-93e9-c7e763c2d64b.png">

ZacharyWills:
<img width="1321" alt="Screen Shot 2020-05-18 at 1 34 23 PM" src="https://user-images.githubusercontent.com/60660437/82247692-72a08080-990c-11ea-9d49-887fdc5e2fc6.png">

ZacharyWills:
Tests are now failing to build, and do not make it to the run stage.

https://github.com/ZacharyWills/ngen/runs/691120396?check_suite_focus=true

hellkite500:
@ZacharyWills @jdmattern-noaa The issue is most likely the test data path is coded relative to the binary, so if the build doesn't make the binary in `ngen/build/` or some other single subdir, the file isn't found, which causes a bug in Forcing.h to index a vector out of bounds.

ZacharyWills:
I changed the path to the root, which works manually but on attempting to get it to run automatedly I run into issues with include/core/catchment/giuh/GIUH.hpp

Not sure why and I'm going to keep dealing with the automation to try and get the tests to work at least. Any location on your caches locally? That would help

ZacharyWills:
Or post your local step-by-step and I can try it with no temp dir dancing.

ZacharyWills:
Tests are working macOS but not on Linux, I'll reference this here:https://github.com/NOAA-OWP/ngen/pull/106#event-3497860029

[...]

PR #83, opened 2020-05-21
Forcing Bounds Checking and UTC Time Fix

jdmattern-noaa:
Added Bounds Checking and UTC Time Fix to Forcing Class

## Additions
Forcing class and unit test bounds checking. Fixed UTC time read from Forcing data.

## Issues
This PR addresses issue #80 with an error being thrown and halting the program if a CSV file is missing. This does not solve the part of that issue of the forcing unit test failing if the directory that the tests are called from is not the build directory and its dependencies on relative paths. A upcoming solution to this part of the issue would also address issue #69, where the problems with these relative paths is discussed. 

## Testing

1. Updated Unit Test

## Checklist

1.

- [x] Linux

### DIFF
--- a/include/forcing/Forcing.h
+++ b/include/forcing/Forcing.h
@@ -102,7 +102,6 @@ class Forcing
      */
     double get_current_hourly_precipitation_meters_per_second()
     {
-        //return precipitation_rate_meters_per_second_vector[forcing_vector_index_ptr];
         return precipitation_rate_meters_per_second_vector.at(forcing_vector_index_ptr);
     }
 
--- a/include/forcing/Forcing.h
+++ b/include/forcing/Forcing.h
@@ -102,7 +102,8 @@ class Forcing
      */
     double get_current_hourly_precipitation_meters_per_second()
     {
-        return precipitation_rate_meters_per_second_vector[forcing_vector_index_ptr];
+        //return precipitation_rate_meters_per_second_vector[forcing_vector_index_ptr];
+        return precipitation_rate_meters_per_second_vector.at(forcing_vector_index_ptr);
     }
 
     /**
@@ -114,8 +115,10 @@ class Forcing
      */
     double get_next_hourly_precipitation_meters_per_second()
     {
-        //Increment forcing index
-        forcing_vector_index_ptr = forcing_vector_index_ptr + 1;
+        //Check forcing vector bounds before incrementing forcing index
+        if (forcing_vector_index_ptr < forcing_vector_size - 1)
+            //Increment forcing index
+            forcing_vector_index_ptr = forcing_vector_index_ptr + 1;
 
         //Increment current time by 1 hour
         current_date_time_epoch = current_date_time_epoch + 3600;
@@ -161,34 +164,34 @@ class Forcing
                 //Row vector
                 std::vector<std::string>& vec = data_list[i];
 
-                //Declare pointer to struct for the current row date-time
-                struct tm *current_row_date_time;
+                //Declare pointer to struct for the current row date-time utc
+                struct tm *current_row_date_time_utc;
 
-                //Allocate memory to struct for the current row date-time
-                current_row_date_time = new tm();
+                //Allocate memory to struct for the current row date-time utc
+                current_row_date_time_utc = new tm();
 
                 //Year
                 string year_str = vec[0];
                 int year = stoi(year_str);
-                current_row_date_time->tm_year = year - 1900;
+                current_row_date_time_utc->tm_year = year - 1900;
 
                 //Month
                 string month_str = vec[1];
                 int month = stoi(month_str);
-                current_row_date_time->tm_mon = month - 1;
+                current_row_date_time_utc->tm_mon = month - 1;
 
                 //Day
                 string day_str = vec[2];
                 int day = stoi(day_str);
-                current_row_date_time->tm_mday = day;
+                current_row_date_time_utc->tm_mday = day;
 
                 //Hour
                 string hour_str = vec[3];
                 int hour = stoi(hour_str);
-                current_row_date_time->tm_hour = hour;
+                current_row_date_time_utc->tm_hour = hour;
 
-                //Convert current row date-time to epoch time
-                time_t current_row_date_time_epoch = mktime(current_row_date_time);
+                //Convert current row date-time utc to epoch time
+                time_t current_row_date_time_epoch = timegm(current_row_date_time_utc);
 
                 //If the current row date-time is within the model date-time range, then add precipitation to vector
                 if (start_date_time_epoch <= current_row_date_time_epoch && current_row_date_time_epoch <= end_date_time_epoch)
@@ -204,7 +207,7 @@ class Forcing
                 }
 
                 //Free memory from struct
-                delete current_row_date_time;
+                delete current_row_date_time_utc;
         }
     }
 
@@ -228,20 +231,20 @@ class Forcing
                 //Row vector
                 std::vector<std::string>& vec = data_list[i];
 
-                //Declare struct for the current row date-time
-                struct tm current_row_date_time;
+                //Declare struct for the current row date-time 
+                struct tm current_row_date_time_utc;
 
                 //Allocate memory to struct for the current row date-time
-                current_row_date_time = tm();
+                current_row_date_time_utc = tm();
 
                 //Grab time string from first column
                 string time_str = vec[0];
 
                 //Convert time string to time struct
-                strptime(time_str.c_str(), "%Y-%m-%d %H:%M:%S", &current_row_date_time);
+                strptime(time_str.c_str(), "%Y-%m-%d %H:%M:%S", &current_row_date_time_utc);
 
-                //Convert current row date-time to epoch time
-                time_t current_row_date_time_epoch = mktime(&current_row_date_time);
+                //Convert current row date-time UTC to epoch time
+                time_t current_row_date_time_epoch = timegm(&current_row_date_time_utc);
 
                 //If the current row date-time is within the model date-time range, then add precipitation to vector
                 if (start_date_time_epoch <= current_row_date_time_epoch && current_row_date_time_epoch <= end_date_time_epoch)
@@ -281,8 +284,11 @@ class Forcing
                 }
 
                 //Free memory from struct
-                //delete current_row_date_time;
+                //delete current_row_date_time_utc;
         }
+
+        //Get size of forcing precipitation rate vector
+        forcing_vector_size = precipitation_rate_meters_per_second_vector.size();
     }
 
     vector<double> APCP_surface_kg_per_meters_squared_vector;
@@ -295,6 +301,7 @@ class Forcing
     vector<double> VGRD_10maboveground_meters_per_second_vector;
     vector<double> precipitation_rate_meters_per_second_vector;
     int forcing_vector_index_ptr;
+    int forcing_vector_size;
     double precipitation_rate_meters_per_second;
     double air_temperature_fahrenheit;
     int basin_id;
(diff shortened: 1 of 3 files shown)


include/forcing/Forcing.h
```diff
@@ -102,7 +102,7 @@ class Forcing
      */
     double get_current_hourly_precipitation_meters_per_second()
     {
-        return precipitation_rate_meters_per_second_vector[forcing_vector_index_ptr];
+        return precipitation_rate_meters_per_second_vector.at(forcing_vector_index_ptr);
```
jdmattern-noaa, line 105: The .at() is less efficient. Do we want this on top of having the bounds check in the get_next function?

jdmattern-noaa:
Looking back at my notes, I need to add a warning when beyond the last forcing value and a check for negative index. I will add these shortly.

include/forcing/Forcing.h
```diff
@@ -228,20 +242,32 @@ class Forcing
...
-                //Convert current row date-time to epoch time
-                time_t current_row_date_time_epoch = mktime(&current_row_date_time);
+                //Convert current row date-time UTC to epoch time
+                time_t current_row_date_time_epoch = timegm(&current_row_date_time_utc);
+
+                //Ensure that forcing data covers the entire model period. Otherwise, throw an error.
+                if (i == 1 && start_date_time_epoch < current_row_date_time_epoch)
+                    /// \todo TODO: Return appropriate error
+                    //cout << "WARNING: Forcing data begins after the model start time." << endl;
+                    throw std::runtime_error("Error: Forcing data begins after the model start time.");
```
jdmattern-noaa, line 264: Warning or error for this?

include/forcing/Forcing.h
```diff
@@ -228,20 +242,32 @@ class Forcing
...
+
+
+                else if (i == data_list.size() - 1 && current_row_date_time_epoch < end_date_time_epoch)
+                    /// \todo TODO: Return appropriate error
+                    cout << "WARNING: Forcing data ends before the model end time." << endl;
```
jdmattern-noaa, line 269: Warning or error for this?

include/utilities/CSV_Reader.h
```diff
@@ -30,6 +30,13 @@ inline std::vector<std::vector<std::string> > CSVReader::getData()
 {
 	std::ifstream file(fileName);
 
+        if(file.fail()){
+            /// \todo TODO: Return appropriate error
+            throw std::runtime_error("Error: Forcing file " + fileName + " does not exist.");
+
+            /// \todo Potentially only output warning and fill array with sentinel values.
```
jdmattern-noaa, line 37: So we do want to go with sentinel values instead of just killing the program here?

include/forcing/Forcing.h
```diff
@@ -257,46 +356,61 @@ class Forcing
...
+        double UGRD_10maboveground_meters_per_second; //U-Component of Wind (m/s)
+        double VGRD_10maboveground_meters_per_second; //V-Component of Wind (m/s)
+    };
+
+    vector<AORC_data> AORC_vector;
+
     vector<double> precipitation_rate_meters_per_second_vector;
+    vector<time_t> time_epoch_vector; 
     int forcing_vector_index_ptr;
+    int forcing_vector_size;
```
hellkite500, line 409: This is tracked in the vector `precipitation_rate_meters_per_second_vector.size()`
It is probably better to use this anywhere you need size in case future work allows this vector to change size during the lifetime of a forcing object.

include/forcing/Forcing.h
```diff
@@ -257,46 +356,61 @@ class Forcing
...
+        double TMP_2maboveground_K; //Temperature (K)
+        double UGRD_10maboveground_meters_per_second; //U-Component of Wind (m/s)
+        double VGRD_10maboveground_meters_per_second; //V-Component of Wind (m/s)
+    };
+
+    vector<AORC_data> AORC_vector;
+
     vector<double> precipitation_rate_meters_per_second_vector;
+    vector<time_t> time_epoch_vector; 
     int forcing_vector_index_ptr;
```
hellkite500, line 408: Thinking about this a bit, might be worth keeping this as an iterator?  Not important for the short term, but something to think about.

include/forcing/Forcing.h
```diff
@@ -114,11 +122,14 @@ class Forcing
      */
     double get_next_hourly_precipitation_meters_per_second()
     {
-        //Increment forcing index
-        forcing_vector_index_ptr = forcing_vector_index_ptr + 1;
+        //Check forcing vector bounds before incrementing forcing index
+        if (forcing_vector_index_ptr < forcing_vector_size - 1)
+            //Increment forcing index
+            forcing_vector_index_ptr = forcing_vector_index_ptr + 1;
```
hellkite500, line 128: This function side effects by incrementing `forcing_vector_index_ptr` which other get functions rely on, which is unclear.  Need to rethink this eventually...

include/forcing/Forcing.h
hellkite500: I don't think it is important, but you would have to check the entire range and handle both side, `< 0` and `> .size()`

include/forcing/Forcing.h
```diff
@@ -133,13 +144,89 @@ class Forcing
...
+
+        /// \todo: Sort out using local versus UTC time
         current_date_time = localtime(&current_date_time_epoch);
 
         current_day_of_year = current_date_time->tm_yday;
 
         return current_day_of_year;
     }
 
+    /**
```
hellkite500, line 157: All of these accessors need to share similar bound checking logic on the index.

include/forcing/Forcing.h
hellkite500: An error is fine for now, I think.

include/forcing/Forcing.h
hellkite500: A warning for now, and proper semantics/runtime warning when we run out of forcing.

include/forcing/Forcing.h
```diff
@@ -257,46 +356,61 @@ class Forcing
...
+        double SPFH_2maboveground_kg_per_kg; //Specific Humidity (kg/kg)
+        double TMP_2maboveground_K; //Temperature (K)
+        double UGRD_10maboveground_meters_per_second; //U-Component of Wind (m/s)
+        double VGRD_10maboveground_meters_per_second; //V-Component of Wind (m/s)
+    };
+
+    vector<AORC_data> AORC_vector;
+
     vector<double> precipitation_rate_meters_per_second_vector;
+    vector<time_t> time_epoch_vector;
```
hellkite500, line 407: Provide an accessor to this vector?

include/forcing/Forcing.h
```diff
@@ -257,46 +356,61 @@ class Forcing
...
+        double DLWRF_surface_W_per_meters_squared; //Downward Long-Wave Rad. (Flux W/m^2)
+        double DSWRF_surface_W_per_meters_squared; //Downward Short-Wave Radiation (Flux W/m^2)
+        double PRES_surface_Pa; //Pressure (Pa)
+        double SPFH_2maboveground_kg_per_kg; //Specific Humidity (kg/kg)
+        double TMP_2maboveground_K; //Temperature (K)
+        double UGRD_10maboveground_meters_per_second; //U-Component of Wind (m/s)
+        double VGRD_10maboveground_meters_per_second; //V-Component of Wind (m/s)
+    };
+
+    vector<AORC_data> AORC_vector;
```
hellkite500, line 404: I would provide an accessor to this vector and allow callers to get the entire structure in one call if they want.

include/utilities/CSV_Reader.h
hellkite500: I think this is an appropriate fatal error

include/forcing/Forcing.h
jdmattern-noaa: Okay thanks. I will leave the .at() in there.

include/forcing/Forcing.h
jdmattern-noaa: Okay. I have that in the get next function.

include/forcing/Forcing.h
jdmattern-noaa: Ok thanks. I will replace that.

include/forcing/Forcing.h
jdmattern-noaa: I will put that as a todo to consider.

include/forcing/Forcing.h
jdmattern-noaa: Ok, I will put a todo for this too.

include/forcing/Forcing.h
jdmattern-noaa: I will put that in.

include/forcing/Forcing.h
jdmattern-noaa: Will do.

include/forcing/Forcing.h
jdmattern-noaa: Will do.

include/forcing/Forcing.h
hellkite500: Its probably worth writing a function that does the bound checking and calling it from each accessor.

include/forcing/Forcing.h
```diff
@@ -42,6 +42,20 @@ struct forcing_params
     }
 };
 
+//AORC Forcing Data Struct
```
jdmattern-noaa, line 45: I had trouble with an accessor to this struct inside of the class, so I moved it outside of the class.

include/forcing/Forcing.h
```diff
@@ -90,9 +104,32 @@ class Forcing
...
 
         //Initialize forcing vector index to 0;
-        forcing_vector_index_ptr = 0;
+        forcing_vector_index = 0;
     }
 
+    /**
+     * @brief Checks forcing vector index bounds and adjusts index if out of vector bounds
+     */
+    void check_forcing_vector_index_bounds()
```
jdmattern-noaa, line 113: Modifying the index in this bounds check.

include/forcing/Forcing.h
```diff
@@ -102,23 +139,30 @@ class Forcing
...
+     * /// \todo: Reconsider incrementing the forcing_vector_index because other functions rely on this, and it
+     *            could have side effects
      * @return the current hourly precipitation in meters per second after pointer is incremented by one timestep
      */
     double get_next_hourly_precipitation_meters_per_second()
     {
-        //Increment forcing index
-        forcing_vector_index_ptr = forcing_vector_index_ptr + 1;
+        //Check forcing vector bounds before incrementing forcing index
+        if (forcing_vector_index < precipitation_rate_meters_per_second_vector.size() - 1)
```
hellkite500, line 160: This is off by 1. Should be `forcing_vector_index < precipitation_rate_meters_per_second_vector.size()` if using strict inequality.

include/forcing/Forcing.h
hellkite500, line 114: This works, but it creates an assumption that all forcing vectors are guaranteed to be equally sized (the size of the precip vector).  This is currently a valid assumption, but I'm not sure how strong the guarantee is.  Something to keep an eye one in the future.

jdmattern-noaa: Good point. I added this todo for that. I am unclear why this line didn't show up here.
https://github.com/NOAA-OWP/ngen/blob/f72b754aff80f0186d0222fdd87784e27c1081b0/include/forcing/Forcing.h#L112

include/forcing/Forcing.h
jdmattern-noaa: This is actually a check before incrementing the index as a bounds check to what would be after incrementing. Do you think this is correct given that case then?

hellkite500: Yup, you are correct.  Disregard.

include/forcing/Forcing.h
hellkite500: Kinda weird.  I'll merge and if it still isn't there we can add it in.

PR #85, opened 2020-05-21
Improvements to GIUH Logic and Incorporating into Realization and Main Exec

robertbartel:
## Additions

- Added GIUH sample data files
- Added `giuh_kernel` unit tests
- Added class for reading GIUH data files and catchment-id-to-COMID mappings (since the GIUH data is keyed by the latter) - `GiuhJsonReader`

## Changes

- Renamed members in `giuh_kernel` for clarity
- Separated interpolation logic from `giuh_kernel` constructor to allow for dynamic recalculation
- Fixed bug in GIUH calculations
- Incorporated `giuh_kernel` and new `GiuhJsonReader` more appropriately into `TshirtRealization`,
and then updated `NGen.cpp` as needed for these changes (with some hard-coded values)

## Testing

1. Unit tests were added, and currently pass.

## Checklist


robertbartel:
One particular thing to note is the hard-coding of the GIUH data JSON file and the id-to-COMID mapping JSON file.  Since I wasn't sure exactly where the executable would be run from, I've coded it to expect `./GIUH.json` and `./crosswalk.json` (the latter not even being the same file basename as the current example data file).  

Obviously eventually these should get parameterized, but these may or may not be the specific hard-coded values we want to use for the demo.

src/NGen.cpp
```diff
@@ -146,6 +147,21 @@ int main(int argc, char *argv[]) {
     double t = 0;
     time_step_t dt = 3600; //tshirt time step
 
+    // TODO: parameterize these values, rather than hard-code
+    // TODO: current mapping file shows "wat-*" rather than "cat-*" for some reason ... figure out if this has further implications
```
hellkite500, line 151: waterbodies and catchments are independent entities that are related.  comids associate with waterbodies, catchments realize waterbodies.

src/realizations/catchment/Tshirt_Realization.cpp
```diff
@@ -6,13 +6,15 @@ Tshirt_Realization::Tshirt_Realization(
...
+        giuh::GiuhJsonReader &giuh_json_reader,
         tshirt::tshirt_params params,
         const vector<double> &nash_storage,
-        Tshirt_Realization::time_step_t t
-        )
-    : HY_CatchmentArea(forcing_config), giuh_kernel(move(giuh_kernel)), params(params), dt(t)
+        time_step_t t)
+    : HY_CatchmentArea(forcing_config), catchment_id(catchment_id), params(params), dt(t)
 {
+    giuh_kernel = giuh_json_reader.get_giuh_kernel_for_id(this->catchment_id);
```
hellkite500, line 16: What to do when `giuh_kernel == nullptr`?

include/core/catchment/giuh/GIUH.hpp
```diff
@@ -28,70 +28,38 @@ namespace giuh {
...
+            interpolation_regularity_seconds = 60;
+
+            // TODO: have this be called by constructor, but consider later handling this concurrently
+            interpolate_regularized_cdf();
         }
 
         /**
           * Destructor
           */
         virtual ~giuh_kernel()= default;;
```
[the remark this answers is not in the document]
robertbartel: Not quite sure why/when that happened, but I can address that.

src/NGen.cpp
robertbartel: I made some inferences about the nature of the objects represented by the JSON data and how relationships between things work.  Given the mapping was between those identifiers (i.e., `wat-*`) and the identifiers/keys of the JSON objects holding GIUH data,  I expected all these things were particular to catchment representations. 

If something like `wat-87` is actually a waterbody id, that make sense, but is it not the case that, e.g., object _"9731248"_ in `GIUH.json` describes data pertaining to a catchment (as opposed to a waterbody)?  If that is indeed the case, where is the relationship to a catchment id (in the form that NGen seems to start with; i.e., `cat-*`) defined?

PR #120, opened 2020-07-21
Add docker build of test executable

ZacharyWills:

## Additions
RHEL 8 Dockerfile and related GitHub Actions
-

## Removals

-

## Changes

-

## Testing

1. https://github.com/ZacharyWills/ngen/actions
## Notes

-

## Todos

-

## Checklist

1.

- [x] Linux


ZacharyWills:
Tests not successful were the tests from my fork that are 403'd due to beta access to the security scanning. Can remove if needed.

ZacharyWills:
Ancillary files removed.

I don't mind whatever ENV we use for building (RHEL vs CentOS) as eventually I would like it to be multistage where we only copy over the executable and maybe some test data to the image for running, I'm testing a CentOS based image now as well. It's worth noting that the RHEL UBI is publicly available, just unable to 'yum update'. 
@hellkite500 
What do you think?

.github/workflows/anchor_containerized_test.yml
```diff
@@ -0,0 +1,32 @@
...
+
+on: push
+
+jobs:
+  Anchore-Build-Scan:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout the code
+      uses: actions/checkout@v2
+    - name: Change working dir
```
hellkite500, line 19: I don't think this is named correctly, `pwd && ls` doesn't do any directory changes

ZacharyWills: Removed debug step and re-commited the file

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+RUN git clone https://github.com/NOAA-OWP/ngen.git
+
+WORKDIR ngen
+
+ENV CXX=/usr/bin/g++
+
+RUN git submodule update --init --recursive -- test/googletest
+
+RUn curl -L -O https://dl.bintray.com/boostorg/release/1.72.0/source/boost_1_72_0.tar.bz2
```
hellkite500, line 13: RUN to be consistent

ZacharyWills:
How does everyone feel about the RHEL UBI vs CentOS image?

hellkite500:
It seems to me like we might want multiple Dockerfiles for various custom environments.  We can create a new subdir `docker` and use a naming convention to help identify the env, i.e. `Dockerfile_rhel8`.

Also, this PR is good candidate for a Squash and Merge :)

ZacharyWills:
Certainly should be squashed! 

Want to call the dir docker or env?

hellkite500:
If it is holding dockerfiles, I think docker is more logical

ZacharyWills:
Added docker folder and moved docker file into that folder
@hellkite500 What do we want to name env dockefiles? I usually would go with something like RHEL.dockerfile and centos.dockerfile

ZacharyWills:
> * I don't think the _Testing&Validation.yml_ file needs to be added like this.  It seems to just be a copy of the existing _test_and_validate.yml_ file.
Removed
> * The CodeQL check is failing in the PR checks.
Removed until working
> * I see a few things in the _Dockerfile_:
>   
>   * Very minor, but there is a small typo on line 13 where `RUn` is used.  I can't remember if it functionally matters, though it is at least messing with Github's syntax highlighting.
Addressed
>   * Per [this documentation](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_atomic_host/7/html-single/getting_started_with_containers/index#getting_images_from_registries), the `registry.access.redhat.com` registry used in the Dockerfile is said to be deprecated, so it is likely better to find an alternative registry (which possibly relates to the next item below).
Added CentOS docker file, supporting both isn't hard, should point automation to CentOS so people can reproduce.
>   * I have a few minor concerns about the use of RHEL here, mostly in the context of the want/need for community involvement and any licensing implications.  To be clear, I am not asserting there are any problems with this, but I am curious and uncertain whether there are.  I want to at least bring it up, though it may have already been considered and deemed not a blocker for rolling this out.
Hopefully addressed in the previous comment.

Issue #124, opened 2020-07-24
Bugs with excess calculation in Reservoir::response_meters_per_second


robertbartel:
The _response_meters_per_second_ function in the _Reservoir_ class does not handle calculating excess water properly in all cases.

## Current behavior

- In some cases, the function does not initialize the `excess_water_meters` parameter (passed in as a reference and used to indirectly return the excess amount), which it appears to be designed to do, even if the value should be `0.0`.
- In cases when there is excess - i.e., when the reservoir's `state.current_storage_height_meters` is greater than `parameters.maximum_storage_meters` - the function calculates the excess amount after resetting `state.current_storage_height_meters` to be equal to the max, which will thus always make the excess value `0.0`.

## Expected behavior
The pass excess reference should always be initialized to either `0.0` or the amount of the reservoir's current storage that is greater than its max capacity, at the end of of each call to _response_meters_per_second_.

PR #125, opened 2020-07-24
Fix bugs with excess calculation in Reservoir::response_meters_per_second

robertbartel:
Fix for issue #124, where excess is not always initialized or calculated correctly in the Reservoir class response function.

### DIFF
--- a/src/models/kernels/reservoir/Reservoir.cpp
+++ b/src/models/kernels/reservoir/Reservoir.cpp
@@ -105,6 +105,7 @@ double Reservoir::response_meters_per_second(double in_flux_meters_per_second, i
 {
     double outlet_velocity_meters_per_second = 0;
     double sum_of_outlet_velocities_meters_per_second = 0;
+    excess_water_meters = 0;
 
     //Update current storage from influx multiplied by delta time.
     state.current_storage_height_meters += in_flux_meters_per_second * delta_time_seconds;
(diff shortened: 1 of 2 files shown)


src/models/kernels/reservoir/Reservoir.cpp
```diff
@@ -143,14 +144,11 @@ double Reservoir::response_meters_per_second(double in_flux_meters_per_second, i
     }
 
     //If storage is greater than maximum storage, set to maximum storage and return excess water.
-    if (state.current_storage_height_meters > parameters.maximum_storage_meters)
-    {
+    if (state.current_storage_height_meters > parameters.maximum_storage_meters) {
         /// \todo TODO: Return appropriate warning
         cout << "WARNING: Reservoir calculated a storage above the maximum storage."  << endl;
-
+        excess_water_meters = state.current_storage_height_meters - parameters.maximum_storage_meters;
```
hellkite500, line 150: @jdmattern-noaa 😄 The little things!  Good catch @robertbartel

src/models/kernels/reservoir/Reservoir.cpp
```diff
@@ -105,6 +105,7 @@ double Reservoir::response_meters_per_second(double in_flux_meters_per_second, i
 {
     double outlet_velocity_meters_per_second = 0;
     double sum_of_outlet_velocities_meters_per_second = 0;
+    excess_water_meters = 0;
```
hellkite500, line 108: I don't see an immediate problem doing this, but the documentation should be very clear that this variable is reset per call.  A caller could choose to hold all excess in the variable and expect an accumulation into it (which we weren't doing.)

src/models/kernels/reservoir/Reservoir.cpp
jdmattern-noaa: Yes, my mistake here. Thanks @robertbartel!

src/models/kernels/reservoir/Reservoir.cpp
robertbartel: That typ of usage would have been a little dangerous even before these changes (at least in some situations, any previous value was already going to be lost), but I see your point.  I have pushed another commit with an update to the docstring that expands on how the reference param is handled, and that previous contents will be lost.

jdmattern-noaa: I think this is fine too.

Issue #128, opened 2020-07-24
Adjustments getting reservoir response in Tshirt model


robertbartel:
The usage of the _Reservoir::response_meters_per_second_ function in the _tshirt_model_ class is not entirely correct/consistent, especially regarding parameters obtained from Schaake partitioning.  These need to be corrected.

More generally, it should be considered whether the function's parameter for how much new water is being introduced should be a depth-type value in the form of meters (thereby implicitly meters per timestep), or a volumetric value with units of meters per second (as is currently the case).

jdmattern-noaa:
@robertbartel PR #130 would change this from response_meters_per_second  to response_meters in only the depth-type value as you suggest. We could further discuss if we want to fully replace the existing reservoir class with the timeless reservoir.

jdmattern-noaa:
@robertbartel Do you agree that issue can now be closed?

PR #129, opened 2020-07-24
Making adjustments getting reservoir response in Tshirt model

robertbartel:
Changes for issue #128.

Note in particular that while initially the idea was to change `response_meters_per_second` to have the first parameter be something in meter rather than meters per second, I have removed this and instead make that adjustment within the logic of _tshirt_model_.  That was the only place known thus far to try to pass meters-based value, rather than in the expected meters-per-second.  Conversely, there are many places that are passing in meters-per-second values, so it seems to make more sense to convert once, fixing the Tshirt logic, than to change the _Reservoir_ class and then have to adjust many other things depending on it.  However, this can be discussed further if needed.

### DIFF
--- a/src/models/kernels/reservoir/Reservoir.cpp
+++ b/src/models/kernels/reservoir/Reservoir.cpp
@@ -118,6 +118,22 @@ namespace Reservoir{
             //Update current storage from influx multiplied by delta time.
             state.current_storage_height_meters += in_flux_meters_per_second * delta_time_seconds;
 
+            //If storage is greater than maximum storage, set to maximum storage and return excess water.
+            if (state.current_storage_height_meters > parameters.maximum_storage_meters)
+            {
+                /// \todo TODO: Return appropriate warning
+                cout << "WARNING: Reservoir calculated a storage above the maximum storage."  << endl;
+
+                state.current_storage_height_meters = parameters.maximum_storage_meters;
+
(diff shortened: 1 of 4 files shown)


robertbartel:
@christophertubbs, it looks like this is failing some of the unit tests for the _Realization_Config_ and  _Realization_Config_Reader_ classes, specifically that make an assertion about the response value from an invocation of Tshirt.  Are the involved values just taken directly from what Tshirt was outputting before, such that after applying corrections to the logic, we should adjust these values?  Or do we know those test values are correct, and so must infer there is a problem being introduced by the changes in this PR?

christophertubbs:
Sorry for the delay. The values are just the ones they were before. It was just checking for consistencies sake.

hellkite500:
Rebase and see what reservoir changes still exist.

robertbartel:
@hellkite500, changes applied.  I also updated the new timeless reservoir class with the analogous functional changes, as that class was added after this change was originally put together.  I didn't apply the variable name refactoring, but I don't see that as a big deal.

robertbartel:
FWIW, the failed test is failing due to lack of some required hymod parameters in the current example config: basically completely unrelated to this change.  That'll need to be fixed, but separately from here.

hellkite500 (approved):

Issue #122, opened 2020-07-22
More Reservoir Functional Unit Testing Needed


jdmattern-noaa:
More reservoir unit testing needed to directly test particular reservoir functions.

## Expected behavior

We should have one or more TEST_F(ReservoirKernelTest, Test****) for each function in a class.  More than one function should be used to test different asusmptions/code paths.  i.e.  `TEST_F(ReservoirKernelTest, TestReservoirAddOutlet_1) ` might test that all assertions are good when reservoirs are added in an already sorted order. TEST_F(ReservoirKernelTest, TestReservoirAddOutlet_2) might test that the outlet states are all valid when an outlet with too large of activation is added...

We might want to refactor the testing at some point so we have something like
TEST_F(ReservoirKernelTest, TestReservoirAddOutlet) 
{   
    NoOutletReservoir2->add_outlet(ReservoirLinearOutlet);
    NoOutletReservoir2->add_outlet(0.3, 0.5, 0.0, 100.0);
    ASSERT_TRUE(true);
}

@jdmattern-noaa will address.

PR #151, opened 2020-08-24
More extensive reservoir unit tests and timeless reservoir unit tests

jdmattern-noaa:
More extensive reservoir unit tests and timeless reservoir unit tests

## Testing

1. Passes all unit tests

## Notes

Fixes issue #122 

## Checklist

1.

- [ ] Linux

### DIFF
--- a/src/models/kernels/reservoir/Reservoir.cpp
+++ b/src/models/kernels/reservoir/Reservoir.cpp
@@ -166,6 +166,7 @@ namespace Reservoir{
                 cerr
                     << "ERROR: excess_water_meters from the reservoir is calculated to be less than zero."
                     << endl;
+                    exit(-1);
             }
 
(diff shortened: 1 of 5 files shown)


jdmattern-noaa:
@hellkite500 I added more extensive unit tests for the various functions. Each type of constructor is called in the setup parts of the tests and then subsequent tests are run on those constructed reservoirs. Should each there be a separate test specifically for each constructor?

hellkite500:
It seems to me if each constructor is called as part of the test, then they are at least indirectly tested as a failed construction would likely cause a seg fault in the test that uses that constructed object.  I'm not 100% sure about the best paradigm for testing constructors.  Trivial constructors that don't have any logic but just set members probably don't NEED explicit tests, but any constructor with some logic in it should have those logical paths tested explicitly.  We can think about this some as we go and add to these tests suites.

PR #162, opened 2020-08-28
Initial Realization Implementation Using C-Style Tshirt Formulation

robertbartel:
Initial version of a C-style implementation of Tshirt model formulation (adapted from Fred's initial C version) and a utilizing realization class, along with unit tests verifying consistent behavior of the adaptation with the original.


src/models/tshirt/tshirt_c.cpp
```diff
@@ -0,0 +1,809 @@
...
+    */
+
+    // local variables
+    //-------------------------------------------------------------------
+    int i;
+    int tstep;
+    double upper_lim;
+    double lower_lim;
+    double diff;
+    char theString[513];   // dangerously hard coded string size... TODO fixme.
```
hellkite500, line 311: looks like this can go away, was used in the original implementation to read the AORC csv lines from file, this data is now passed into this function, so this shouldn't be needed.

hellkite500 (approved):
This can be merged to help continue work.  I think a few of the comments in this review and left in the code do need to be addressed in the near future, but we need to be able to continue making progress on multiple fronts and this code should help us do that.

robertbartel:
I've made a few more changes, including a somewhat more complete `create_formulation` implementation (though I'd still like to move away from that to a factory or something).  

@christophertubbs, in particular, since it's been a while since you went over how the new realization/formulation manager stuff works, is there anything else this needs to be fully incorporated with that?

hellkite500 (approved):

hellkite500 (approved):

hellkite500:
@robertbartel are you ready to merge this?

robertbartel:
@hellkite500, yes, it's good, so I went ahead and rebased/merged it in.

PR #193, opened 2020-10-19
Added the BMI header

christophertubbs:
I added the BMI header to the `include` directory, then added it to the .gitignore to prevent further modification.

As far as I understand, this is a reasonable way to do it? If not, I can play around with it some more.

### DIFF
--- a/include/bmi.hpp
+++ b/include/bmi.hpp
@@ -1,0 +1,85 @@
+// The Basic Model Interface (BMI) C++ specification.
+//
+// This language specification is derived from the Scientific
+// Interface Definition Language (SIDL) file bmi.sidl located at
+// https://github.com/csdms/bmi.
+
+#ifndef BMI_HPP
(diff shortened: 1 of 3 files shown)


include/bmi.hpp
```diff
@@ -0,0 +1,85 @@
+// The Basic Model Interface (BMI) C++ specification.
+//
+// This language specification is derived from the Scientific
+// Interface Definition Language (SIDL) file bmi.sidl located at
+// https://github.com/csdms/bmi.
+
```
hellkite500, line 6: I would add the version of the BMI this corresponds to as either a macro or a comment (maybe both???) just to be clear.

hellkite500 (changes requested):

christophertubbs (commented):
Yeah, that makes perfect sense

hellkite500 (approved):

PR #197, opened 2020-10-21
Improvements to output from driver execution

robertbartel:
Updating the design to supply better (and more model-specific) output files after driver execution.

## Additions

- Updated Formulation interfaces with functions for getting per-time-step output data, allowing implementations to control the format of what is output from them in a way consistent with the output fields they produce.
- Added implementations of aforementioned new functions to Formulation sub-classes.
- Updated Formulation implementation for Tshirt_C with additional output meta functions for getting the number of output fields, the output field names, and an array of all currently-calculated values for a specified output field.
- Added new unit tests for new functionality.

## Removals

-

## Changes

- Updating main driver to use new output functions from Formulation instances to build output files.
- Fixing bug for Tshirt C-based model, uncovered while verifying new format and content of produced output files, that resulted from models `run` function expecting input in millimeters per time step, while everything else dealt with meters.
- Made corrections to unit tests necessary after making Tshirt_C expect input in meters.

## Checklist

1.

- [ ] Linux

### DIFF
--- a/include/realizations/catchment/Formulation.hpp
+++ b/include/realizations/catchment/Formulation.hpp
@@ -59,7 +59,7 @@ namespace realization {
              *
              * @return An appropriate header line for this type.
              */
-            virtual std::string get_output_header_line(std::string delimiter=",") =0;
+            virtual std::string get_output_header_line(std::string delimiter=DEFAULT_FORMULATION_OUTPUT_DELIMITER) =0;
 
             /**
              * Get a formatted line of output values for the given time step as a delimited string.
@@ -78,7 +78,8 @@ namespace realization {
              * @param delimiter The value delimiter for the string.
              * @return A delimited string with all the output variable values for the given time step.
              */
-            virtual std::string get_output_line_for_timestep(int timestep, std::string delimiter=",") = 0;
+            virtual std::string get_output_line_for_timestep(int timestep,
+                                                             std::string delimiter = DEFAULT_FORMULATION_OUTPUT_DELIMITER) = 0;
             
             virtual void create_formulation(boost::property_tree::ptree &config, geojson::PropertyMap *global = nullptr) = 0;
             virtual void create_formulation(geojson::PropertyMap properties) = 0;
--- a/include/realizations/catchment/Formulation.hpp
+++ b/include/realizations/catchment/Formulation.hpp
@@ -14,6 +14,8 @@
 #include <boost/property_tree/ptree.hpp>
 #include <boost/algorithm/string.hpp>
 
+#define DEFAULT_FORMULATION_OUTPUT_DELIMITER ","
+
 namespace realization {
 
     class Formulation : public virtual Et_Aware {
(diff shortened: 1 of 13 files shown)


stcui007:
Does the current code produce test.out file as tshirt c-code? Thanks.

include/realizations/catchment/Formulation.hpp
```diff
@@ -48,6 +50,35 @@ namespace realization {
...
+
+            /**
+             * Get a header line appropriate for a file made up of entries from this type's implementation of
+             * ``get_output_line_for_timestep``.
+             *
+             * Note that like the output generating function, this line does not include anything for time step.
+             *
+             * @return An appropriate header line for this type.
+             */
+            virtual std::string get_output_header_line(std::string delimiter=",") =0;
```
hellkite500, line 62: Why not use the macro defined? `DEFAULT_FORMULATION_OUTPUT_DELIMITER`

src/NGen.cpp
```diff
@@ -228,7 +228,15 @@ int main(int argc, char *argv[]) {
         double response = formulation_pair.second->get_response(time_step, 3600.0);
         //dump the output
         std::cout<<"\tCatchment "<<formulation_pair.first<<" contributing "<<response<<" m/s to "<<catchment_to_nexus[formulation_pair.first]<<std::endl;
-        catchment_outfiles[formulation_pair.first] << time_step <<", "<<response<<std::endl;
+        // If the timestep is 0, also write the header line to the file
+        // TODO: add command line or config option to have this be omitted
+        if (time_step == 0) {
+            // Append "Time Step" to header string provided by formulation, since we'll also add time step to output
+            std::string header_str = formulation_pair.second->get_output_header_line();
+            catchment_outfiles[formulation_pair.first] << "Time Step," << header_str <<std::endl;
```
hellkite500, line 236: Really should pipe this output to the internal stream handler of the realization.

hellkite500: Note this is stored via a StreamHandler in the HY_CatchmentArea as `output_stream` in a protected member.  Might need to move that and make the access a little simpler.

src/models/tshirt/tshirt_c.cpp
```diff
@@ -446,7 +446,10 @@ extern int run(NWM_soil_parameters& NWM_soil_params,
     // Rather than read in forcing or rainfall data from file (as in original) function receives directly as param
     if(yes_aorc == TRUE) {
         for (i = 0; i < num_timesteps; i++) {
-            rain_rate[i] = (double) aorc_data[i].precip_kg_per_m2;  // assumed 1000 kg/m3 density of water.  This result is mm/h;
+            // assumed 1000 kg/m3 density of water.  The AORC data result is mm/h, so also convert to m/h.
+            // TODO: test to confirm the conversion to meters works out correctly
+            rain_rate[i] = ((double) aorc_data[i].precip_kg_per_m2) / 1000;
```
hellkite500, line 451: I will note that the AORC input read from the internal Forcing object and provided as the input flux is different than this value (it is pre-processed m/s and adjusted for densitity/temperature relationship in the pre-processing.)

stcui007:
I have compiled  and run this PR and it seems to work fine.

include/realizations/catchment/Tshirt_C_Realization.hpp
```diff
@@ -8,7 +8,17 @@
 #include "GiuhJsonReader.h"
 #include "tshirt_c.h"
 #include <memory>
+#include <set>
```
hellkite500, line 11: I didn't see anything in this change set that used `set`, if it is not used, should be be removed.

hellkite500 (changes requested):
The only significant change needed at this time is removing unused import.

include/realizations/catchment/Formulation.hpp
robertbartel: Must have been having a moment there.  Fixed in `2a1184e3d542`

include/realizations/catchment/Tshirt_C_Realization.hpp
robertbartel: Yeah, I think I started to do something in the changes that used it, but then later went with a different approach.  Looks like I just forgot to take that out.  Fixed in `066bad4f3701`.

src/NGen.cpp
robertbartel: Yeah, I agree this should be done, but I wasn't super familiar with the new StreamHandler type.   Also, the primary goal was to have the contents of the output be more complete and useful, so I was going to leave adjusting the overall output implementation for the driver to a separate issue.  I think your assessment was also to leave this until later, @hellkite500, but let me know if I have that wrong.

src/models/tshirt/tshirt_c.cpp
robertbartel: The Tshirt C code uses a custom struct for holding AORC data, which got copied straight from Fred's original code.  It may just be incompatible with our preprocessed AORC data format (or, rather, needing a more complex adapter object to transfer between forms).  

I don't think it matters though (at least for now), since the _Tshirt_C_Realization_ class always handles reading the forcing data and then just passes in the raw rain rates.  I have opened #198 to keep track of either verifying this works (in case it is ever used in the future) or just taking it out.

hellkite500 (approved):

Issue #144, opened 2020-08-18
Request to remove attributes from GIUH.json file


BrianAvant:
Request to simplify GIUH output file.

## Current behavior
The GIUH.json file contains attributes that may not be used in ngen and removing them could greatly reduce file size. Current attributes include cumulative frequency distribution ('CumulativeFreq'),  incremental runoff per hour ('hrHydro') and incremental runoff per minute ('minHydro') time series. It looks like only 'CumulativeFreq' is being used in the demo. Do we need these other attributes? We should be able to derive incremental runoff for any time step from the cumulative frequency time series. @hellkite500 @robertbartel 

## Expected behavior
Remove redundant attributes from GIUH.json

## Steps to replicate behavior (include URLs)

1.

PR #200, opened 2020-10-23
Fix GIUH Logic for Tshirt C

robertbartel:
Fixing the logic for initializing the correct GIUH ordinates over the appropriate interpolation regularity in the logic for creating Tshirt_C realizations.

### DIFF
--- a/src/core/catchment/giuh/GiuhJsonReader.cpp
+++ b/src/core/catchment/giuh/GiuhJsonReader.cpp
@@ -30,8 +30,10 @@ std::vector<double> GiuhJsonReader::extract_cumulative_frequency_ordinates(ptree
     if (catchment_data_node.get_child_optional("CDF.CumulativeFreq") != boost::none) {
         freq_node_name = "CDF.CumulativeFreq";
     }
+    // TODO: later apply logic to be able to handle if there is a 'CDF.minHydro.Runoff', though that may also belong
+    //  elsewhere since those are not cumulative frequencies.
     else {
-        freq_node_name = "CDF.minHydro.Runoff";
+        throw std::runtime_error("Unable to find GIUH cumulative frequencies data node in parsed GIUH JSON");
     }
 
     for (ptree::value_type freqs : catchment_data_node.get_child(freq_node_name)) {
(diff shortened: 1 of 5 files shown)


jdmattern-noaa (commented):
Looks good to me.

stcui007:
This PR produces the following runtime error:
$ ./cmake-build-dir/ngen ~/t-shirt/catchment_data.geojson cat-87
~/t-shirt/nexus_data.geojson nex-94 ~/t-shirt/realization_config.json
Hello there 0.1.0
Building Nexus collection
Building Catchment collection
terminate called after throwing an instance of 'std::runtime_error'
  what():  ERROR: No simulation time period defined.
Aborted (core dumped)

This might be inherited from an earlier merge.

src/core/catchment/giuh/GiuhJsonReader.cpp
```diff
@@ -26,9 +26,23 @@ std::vector<double> GiuhJsonReader::extract_cumulative_frequency_ordinates(ptree
     std::vector<double> cumulative_freqs;
 
     // TODO: account for error condition of unmatching JSON structure for freqs
-    for (ptree::value_type freqs : catchment_data_node.get_child("CDF.CumulativeFreq")) {
+    std::string freq_node_name;
+    if (catchment_data_node.get_child_optional("CDF.CumulativeFreq") != boost::none) {
+        freq_node_name = "CDF.CumulativeFreq";
+    }
+    else {
+        freq_node_name = "CDF.minHydro.Runoff";
```
hellkite500, line 34: I don't believe this key in the GIUH is a cumulative freq, I think it is a computed incremental run off.  My suspicion is that if this else case is ever encountered, the runtime_error below will always trigger.

hellkite500 (changes requested):

hellkite500:
#144 is relevant here especially in understanding the keys used in GIUH.json.  Based on Brians comment in that issue, we definitely should remove the fallback to `CDF.minHydro.Runoff` in the frequency reading function.

src/core/catchment/giuh/GiuhJsonReader.cpp
hellkite500: Just throw a runtime error here for now "GIUH doesn't contain CumlativeFreq key"

hellkite500:
Merge once above comment is resolved

src/core/catchment/giuh/GiuhJsonReader.cpp
robertbartel: Change applied.

hellkite500 (approved):

Issue #201, opened 2020-10-26
Broken model run tests 


ZacharyWills:
I think the gmake file wasn't updated somewhere along the way, so the new NGen.cpp fails to build. 

```
[ 97%] Building CXX object CMakeFiles/ngen.dir/src/NGen.cpp.o
714
/ngen/src/NGen.cpp: In function 'int main(int, char**)':
715
/ngen/src/NGen.cpp:241:55: error: 'time_step' was not declared in this scope
716
         catchment_outfiles[formulation_pair.first] << time_step << "," << current_timestamp << "," << output_str << std::endl;
717
                                                       ^~~~~~~~~
718
/ngen/src/NGen.cpp:241:55: note: suggested alternative: 'time_step_t'
719
         catchment_outfiles[formulation_pair.first] << time_step << "," << current_timestamp << "," << output_str << std::endl;
720
                                                       ^~~~~~~~~
721
                                                       time_step_t
722
gmake[3]: *** [CMakeFiles/ngen.dir/build.make:63: CMakeFiles/ngen.dir/src/NGen.cpp.o] Error 1
723
gmake[2]: *** [CMakeFiles/Makefile2:75: CMakeFiles/ngen.dir/all] Error 2
724
gmake[1]: *** [CMakeFiles/Makefile2:87: CMakeFiles/ngen.dir/rule] Error 2
725
gmake: *** [Makefile:175: ngen] Error 2
726
The command '/bin/sh -c cmake --build /ngen --target ngen' returned a non-zero code: 2
727
```

## Current behavior
Model runs are failing at the build, and need to be fixed.

## Expected behavior
Every time the action is triggered with new PRs the model should run

## Steps to replicate behavior (include URLs)

1. https://github.com/NOAA-OWP/ngen/runs/1299970026?check_suite_focus=true

## Screenshots
<img width="1674" alt="Screen Shot 2020-10-26 at 7 52 49 AM" src="https://user-images.githubusercontent.com/60660437/97174596-4eb95480-1760-11eb-9d3b-96f502bddb94.png">

hellkite500:
This happens because the CI container is building from `master` (see line 579 below):
<img width="1208" alt="Screen Shot 2020-11-10 at 4 31 49 PM" src="https://user-images.githubusercontent.com/5186647/98741511-6acd1080-2372-11eb-8b13-c437fed33a40.png">

[...]

PR #202, opened 2020-10-26
Fix Units in C-based Tshirt Formulation

robertbartel:
The units for input data were not entirely consistent between the formulation and model code for the C-based Tshirt model.  This corrects that by having the formulation do a conversion before calling the model code.  It also updates unit tests as appropriate.

### DIFF
--- a/src/realizations/catchment/Tshirt_C_Realization.cpp
+++ b/src/realizations/catchment/Tshirt_C_Realization.cpp
@@ -405,10 +405,13 @@ const std::vector<std::string>& Tshirt_C_Realization::get_required_parameters() 
 double Tshirt_C_Realization::get_response(time_step_t t_index, time_step_t t_delta_s) {
     // TODO: check that t_delta_s is of approprate size
 
-    // TODO: this is problematic, because what happens if the wrong t_index is passed?
-    double precip = this->forcing.get_next_hourly_precipitation_meters_per_second();
-    int response_result = run_formulation_for_timestep(precip);
+    // TODO: add some logic for ensuring the right precip data is gathered.  This will need to include considerations
+    //  for the cases when the dt is larger, smaller and equal to a specific, discrete data point available in the
+    //  forcing.  It may actually belong within the forcing object.
 
+    // TODO: it also needs to account for getting the right precip data point (i.e., t_index may not be "next")
+    double precip = this->forcing.get_next_hourly_precipitation_meters_per_second();
+    int response_result = run_formulation_for_timestep(precip, t_delta_s);
     // TODO: check t_index is the next expected time step to be calculated
 
     return fluxes.back()->Qout_m;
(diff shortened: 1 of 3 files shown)


ZacharyWills:
As a note we were seeing failures related to #201 so keep that in mind with regards to your results. Until that's fixed we might get a failure before your tests etc are run.

src/realizations/catchment/Tshirt_C_Realization.cpp
```diff
@@ -495,7 +495,13 @@ int Tshirt_C_Realization::run_formulation_for_timesteps(std::vector<double> inpu
...
     aorc_forcing_data empty_forcing[1];
 
-    double* input_as_array = &input_fluxes[0];
+    // Since the input_fluxes param values are in meters per second, this will need to do some conversions to what gets
+    // passed to Tshirt_C's run(), which expects meters per hour.
+    std::vector<double> input_meters_per_hour(input_fluxes.size());
+    for (int i = 0; i < input_fluxes.size(); ++i) {
+        input_meters_per_hour[i] = input_fluxes[i] * 3600;
+    }
+    double* input_as_array = &input_meters_per_hour[0];
```
hellkite500, line 504: I agree that this needs to be done, but not sure about doing it here.  When I fixed this locally I applied the conversion in `get_response` like this:

```diff
@@ -407,6 +408,8 @@ double Tshirt_C_Realization::get_response(time_step_t t_index, time_step_t t_del
 
     // TODO: this is problematic, because what happens if the wrong t_index is passed?
     double precip = this->forcing.get_next_hourly_precipitation_meters_per_second();
+    //convertp precip from m/s to m/time_step
+    precip = precip * t_delta_s;
     int response_result = run_formulation_for_timestep(precip);
 
     // TODO: check t_index is the next expected time step to be calculated

```

The conversion in `run_formulation_for_timesteps` ASSUMES a `dt` of 3600.  The rest of the formulation works in a `m/timestep` fashion (but is there ever an assumption that that timestep is 1 hr?), so this function is assuming inputs are in m/s, which is probably reasonable, but then hard codes the assumption that they are valid for 3600 seconds.  It works for the intended use right now, but we either need to move the conversion up or pass the delta time down.

robertbartel: @hellkite500, this relates to a higher-level issue I think needs to be address, in that `get_response` probably needs to be changed to not use any parameters.  I'll explain a little further down, but the above reviewed code was written with that eventuality in mind.  That matters here because the potential for inconsistent conversion factor for precipitation is only due to bad premise currently built into the code, and thus probably temporary. 

Regarding exactly where the unit conversion logic is, that is primarily for simplicity.  Any calls to `run_formulation_for_timestep` or `get_response` eventually result in a nested called to `run_formulation_for_timesteps`.   All expect to receive or read data in m/s, so converting ahead of this would require either duplicated code or a more significant rewriting.  

As for changing `get_response` ...

At least in some (and perhaps in all) formulations, including this one, it only ever makes sense to run the "next" time step, so `t_index` just causes problems.  Pretty sure this already qualifies as a bug (we get the "next" forcing precipitation value regardless of the `t_index`), and we just don't have other things written that exercise the problematic states.  I suppose we could consider adding an overloaded function that uses it and calculates multiple time steps sequentially, but we need to do something differently.

I also think the time step size should be taken out as a parameter.  I am fairly sure we want to support models requiring a static time step size, and I want to say the C-base Tshirt in particular does this.  Given that, it seems more appropriate not have a time step size in `get_response`, and instead expect formulations to handle whether/how time step size is mutable within their own implementations.  

I'm open to discussing those further, but I wasn't going to include them as part of this, since they seemed out of scope.  They may be a necessary prereq, though.

robertbartel: @hellkite500 from our conversation yesterday, I made some changes.  I have left the conversion logic where it was so it only has to be coded in one spot, but the class is now passing the `dt` value to the `run_formulation_*` functions to make sure the appropriate value is available, rather than just something hard coded.

hellkite500 (approved):

PR #206, opened 2020-11-16
Driver update

hellkite500:
Make sure `Tshirt_C_Realization::get_response()` returns `m/s` and write the computed volume (`m^3/s`) to the output file.

### DIFF
--- a/src/NGen.cpp
+++ b/src/NGen.cpp
@@ -233,6 +233,7 @@ int main(int argc, char *argv[]) {
       for (std::pair<std::string, std::shared_ptr<realization::Formulation>> formulation_pair : manager ) {
         formulation_pair.second->set_et_params(pdm_et_data);
         //get the catchment response
+        //TODO impose unit from formulation
         double response = formulation_pair.second->get_response(output_time_index, 3600.0);
         //dump the output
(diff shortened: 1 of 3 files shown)


robertbartel:
@hellkite500 , I'm looking at and thinking about this, and it highlights an interesting issue.  The _Formulation_ interface as documented (and _Catchment_Formulation_ as well) doesn't directly state the type of value for response output from the `get_response` function.  The documentation just refers to it as "output."   

Some subtypes specifically reference "discharge" in docstrings, even though those return values in `m/s` rather than `m^3/s`.  We'll need to address that inconsistency (though separately from this should be fine), but the real complication is the ambiguity.  

I think ambiguity in the design is necessary, as we want to avoid forcing restrictions on models.  There will need to be some way of getting to consistent units, though.  This probably deserves some discussion.

hellkite500:
I completely agree.  The driver should expect a result in a given unit, and each realization provides that.  Other than for making CFE work, I don't think I like this patch much.  It would be better to allow the realization to compute the area and adjust the return value inside `get_response` or somewhere else lower down the stack.

donaldwj:
A Nexus class can in theory be used to move data of arbitrary units. It is perhaps best if we add functions to the nexus classes that returns what units it dealing with. We will probably want to use the classes for values that are not m^3/s eventually.

Donald Johnson

PR #208, opened 2020-11-23
Simple Lumped fixes

hellkite500:
Working through a simple lumped calibration example, I identified several fixes and changes required for the realizaiton/formulation.
## Additions

- `smax` parameter that is used in the storage distribution function
- `gw_max_storage` parameter for building the groudwater/slow flow reservoir
- `nash_cascade_max_storage` parameter for building the quick flow reservoirs

## Removals

- `max_storage_parameter` essentially replaced by smax, in hindsight, maybe should have left this one and add the differentiation for gw/nash

## Changes

- Scaled storage distribution values by timestep before passing as input to reservoirs (which expect `m/s`)
- fix double accounting bug in mass balance
- compute mass balance using absolute value
- simplify create_formulation function adl overload

## Testing

1. Updated lumped realization config strings with new params, all tests pass

## Notes

- max outlet velocity may need some re-thinking...

## Todos

-

## Checklist

1.

- [ ] Linux


models/hymod/include/hymod_params.h
```diff
@@ -8,10 +8,12 @@
 struct hymod_params
 {
     double min_storage_meters; //!< minimum amount of water stored
-    double max_storage_meters; //!< maximum amount of water stored
+    double gw_max_storage_meters; //!< maximum amount of water stored in groundwater
+    double nash_max_storage_meters; //!< maximum amount of water stored in each nash cascade reservoir
     double activation_threshold_meters_nash_cascade_reservoir; //!< meters from the bottom of the reservoir to the bottom of the outlet
     double activation_threshold_meters_groundwater_reservoir; //!< meters from the bottom of the reservoir to the bottom of the outlet
     double reservoir_max_velocity_meters_per_second; //!<max outlet velocity in meters per second
```
robertbartel, line 15: Does `gw_storage` not need to be added here as well?

models/hymod/include/Hymod.h
```diff
@@ -113,14 +113,13 @@ class hymod_kernel
             inital_mass_meters += current_state.Sr[i];
             final_mass_meters += next_state.Sr[i];
         }
-
-        // increase the inital mass by input value
-        inital_mass_meters += input_flux_meters;
+        // increase the inital mass by input value, NJF run function above adds input_flux to state before it is passed here
+        //inital_mass_meters += input_flux_meters;
```
robertbartel, line 117: This feels problematic to me.  It might be worth discussing it a little more in depth, though I also don't want to tie this up in a can of worms if it's too hard to adjust the related upstream behavior in the `run` function.

Regardless, if this is going to be commented out as such, shouldn't we also remove the `input_flux_meters` function parameter?

models/hymod/include/hymod_params.h
hellkite500: `gw_storage` is a state variable managed in this case entirely by the `Simple_Lumped_Model_Realization`

models/hymod/include/Hymod.h
hellkite500: What is problematic about it?  line 56 clearly adds the `input_flux` to the state, so doing it here double counts that input and causes the mass balance to fail incorrectly.  The only other solution would be to copy the state at the beginning of the `run` function so the mass balance check has a "clean" state to check against.

I do agree if we don't need the parameter, it should be dropped from the calling statement.

robertbartel: Having the _input_ flux not be separated from the _initial_ state stands out in my mind for some reason, like the type of thing that could have unanticipated, subtle implications.  On the other hand, I may just be biased from having seen it done that way previously, and from seeing the fluxes separate from the _next_ state.  

This is really a question about something beyond the scope of this PR, though, so it can be tabled if it needs to be pursued.  For now, I think as long as the parameter is removed, this should be fine.

models/hymod/include/hymod_params.h
robertbartel: Ah, I see.  I was thinking about it like a model param, but it belongs with model state and is already in that struct.

models/hymod/include/Hymod.h
hellkite500: Feel free to open an issue outlining your other concern and we can circle back to it.

hellkite500: paramter removed in ac00d07

donaldwj (approved):
If I remember simple lumped has had a few problems when we added the reservoir objects to it. This hopefully resolves all those issues.

PR #217, opened 2021-03-10
Initial support for BMI models in C libraries via dynamic loading

robertbartel:
Initial version of dynamically loaded BMI C libraries to support multiple C BMI models without linker issues.  

## Additions

- The library file path is now part of the formulation config.
- Because the library file is now in the formulation config, the single example file for BMI usage has now been replaced with two OS-specific versions, each with the applicable format for the library file name (i.e., and file extension). 

## Changes

- The library, and in particular its required registration function, now are dynamically loaded as part of the formulation creation process (in particular when the _Bmi_C_Adapter_ object is instantiated).
- A handle to the loaded dynamic library is now maintained as a member variable within the _Bmi_C_Adapter_ object, which subsequently is used to close the library when the model is `Finalize`'d.  
  - This would likely cause problems with copying, and as such note the below-mentioned removal of the copy constructor.

## Removals

- The copy constructor of the _Bmi_C_Adapter_ class has been commented out.  
  - With how the handle to the dynamic library works, and how it must be closed, copying behavior needs to be more carefully thought out if it is necessary.  
  - This may have been needed anyway, as copying the model adapter object might have behaved unexpectedly without the ability to serialize/deserialize a backing model's state.
- Usages of and references to BMI-related CMake variables were removed.
- Forward declaration inside the framework for the required BMI struct registration function was remove, as now the function gets dynamically loaded.
- The secondary registration function within the BMI example library, with its name adjusted to be specific to the CFE library, was removed, with its logic just moved directly to the main registration function (which before just had a nested call to the former).

## Testing

- A few simple tests have been added, with macros defined to obtain the right OS-specific basename for the compiled CFE example library.  
- In particular, tests were added for testing multiple simultaneous model formulations, with these independently making the calls to dynamically load the backing library (though with each loading the same library).
- As mentioned above, the previous example JSON formulation config was duplicated, with there now being essentially two that just change the extension of the configured library file.

## Notes

Implementation relies on library-provided registration function for setup of the BMI struct function pointers, but this is consistent with the BMI-supplied example library (and, FWIW, the prior non-dynamic implementation for using a single configured BMI C library).  It is also discussed in the documentation for BMI models, with links to examples provided.  

Also, while tests were created that loaded the example CFE library multiple times and verified they could be operated in multiple simultaneous instances, it was not possible to fully test dynamically loading and using two different models.   

## Todos

- In the future, alternatives will need to be added to the aforementioned library-provided registration function.  Two good suggestions are:
  - supporting the standard BMI function names by default, accounting for a small number of formats (i.e., 1 or 2, like their default with underscore and camel case)
  - a configured mapping of function names to the BMI function pointer with which they should associate.
- As discussed about in removals, there might need to be some thought put into if and how the copy constructor for the _Bmi_C_Adapter_ class should be re-implemented.
- We should add testing for simultaneous use of two different C-based models when a second library becomes available.


robertbartel:
Current failures in checks seems to be caused by the [removal of Boost as a provided dependency](https://github.com/actions/virtual-environments/issues/2667) in the GitHub Actions images we use.  That should be resolved once #218 is merged in.

include/realizations/catchment/Bmi_C_Adapter.hpp
```diff
@@ -29,22 +24,30 @@ namespace models {
 
         public:
 
-            explicit Bmi_C_Adapter(std::string forcing_file_path, bool model_uses_forcing_file, bool allow_exceed_end,
+            explicit Bmi_C_Adapter(std::string library_file_path, std::string forcing_file_path,
```
hellkite500, line 27: Are we assuming that BMI formulations will be reading forcings directly from disk?  I  was  under the impression that NGEN would read a  standard forcing and proxy the data to the formulations.

include/realizations/catchment/Bmi_C_Adapter.hpp
```diff
@@ -29,22 +24,30 @@ namespace models {
...
-            Bmi_C_Adapter(const std::string& bmi_init_config, std::string forcing_file_path,
+            Bmi_C_Adapter(std::string library_file_path, const std::string &bmi_init_config,
+                          std::string forcing_file_path,
                           bool model_uses_forcing_file, bool allow_exceed_end, bool has_fixed_time_step,
-                          const geojson::JSONProperty& other_input_vars, utils::StreamHandler output);
+                          const geojson::JSONProperty &other_input_vars, utils::StreamHandler output);
 
             // Copy constructor
-            Bmi_C_Adapter(Bmi_C_Adapter &adapter);
+            // TODO: since the dynamically loaded lib and model struct can't easily be copied (without risking breaking
```
hellkite500, line 46: Making a copy of the handle does seem problematic, and even if it is POSSIBLE using serialize/deserialize, I don't know that copying is a semantic that would be useful.  I'm thinking that maybe just move semantics is really what is wanted, so that all BMI formulations are singleton by nature.

src/realizations/catchment/Bmi_C_Formulation.cpp
```diff
@@ -21,10 +21,17 @@ std::string Bmi_C_Formulation::get_formulation_type() {
  */
 std::shared_ptr<Bmi_C_Adapter> Bmi_C_Formulation::construct_model(const geojson::PropertyMap& properties) {
     // First examine properties to see if other input variable values are provided
+    auto library_file_iter = properties.find(BMI_REALIZATION_CFG_PARAM_OPT__LIB_FILE);
```
hellkite500, line 24: I wonder if it is reasonable to use standard library paths and library naming conventions to simplify this configuration.  Pointing to an exact library can be quite useful, so I wouldn't drop that, but say a BMI realization is defined in  the configuration that doesn't provide an explicit path, we could search `LD_LIBRARY_PATH` or explicit directories like `/lib` a  library file named `lib<bmi_name>.so`

hellkite500 (approved):

include/realizations/catchment/Bmi_C_Adapter.hpp
robertbartel: We are assuming that in general, both are possible depending on the model.  This is based on my interpretation of [the `initialize` documentation](https://bmi.readthedocs.io/en/latest/#initialize) and follow-up discussions with Scott.  

The initial CFE BMI implementation was built to read its own data, at least initially.  That may need to change in the future, though.  Support for the proxying-type situation is being added via #219.

include/realizations/catchment/Bmi_C_Adapter.hpp
robertbartel: Yeah, I don't think we could actually copy it.  We'd have to something more complicated; e.g., create a more complicated construct around holding it, and use a copyable shared pointer to that type.  Regardless, I agree it's something not worth worrying about unless/until something actually compels us to.

src/realizations/catchment/Bmi_C_Formulation.cpp
robertbartel: That's a good idea.  I think this should suffice for now, but I opened issue #221 to keep track of doing that in the future.

PR #218, opened 2021-03-10
Fix Actions config after Boost change.

robertbartel:
Fixing Github Actions config after Github [removal of Boost from Actions images](https://github.com/actions/virtual-environments/issues/2667).

## Changes

- Actions running on Ubuntu have steps added for acquiring Boost manually
- Actions running on Ubuntu have CMake build init step adjusted to set path to Boost via env variable


stcui007 (approved):
The pull request fixed no Boost library issue.

PR #219, opened 2021-03-11
Update BMI C Formulation to Properly Handle Input Variables

robertbartel:
Updating the BMI C formulation class to be able to handle setting the publicized _input_ variables of the backing BMI model as needed before model updates.

## Additions

- Macro constants and new functions in _Forcing_ class to make it easy to get the known available forcing data fields and get the current value for a field. 
- Members to keep track of and allow access to whether forcing data fields in _Forcing_ are an aggregate sum over a time step (e.g., rain fall), as opposed to an instantaneous reading or average amount (e.g., pressure).
- Stub pass-thru function in _Forcing_ class as a placeholder for eventual function that converts forcing value to other units.
- Function in _Forcing_ class to get the size of the current forcing data time step (needed when a model's time steps do not align).
- Configuration support in _Bmi_Formulation_ class for accepting a mapping of model input variable names to recognized forcing field names, along with updated documentation.
- Primary and helper functions for reading, interpreting, acquiring values for, and setting BMI model input variables, including the ability to get values correctly when needed input applies to only part of a forcing time step and/or multiple forcing time steps.

## Changes

- Modified BMI C formulation to call new function for setting BMI input variables with `get_response` function, just prior to executing BMI model `update`.

## Testing

- Several automated unit tests for new functionality.

## Todos

- A stub function was created for unit conversions for source forcing values, but right now no conversions are performed; eventually there will need to be something that can handle conversions based on the expected units the BMI model publicizes.

### DIFF
--- a/src/realizations/catchment/Bmi_C_Formulation.cpp
+++ b/src/realizations/catchment/Bmi_C_Formulation.cpp
@@ -251,3 +251,55 @@ double Bmi_C_Formulation::get_var_value_as_double(const int& index, const std::s
 bool Bmi_C_Formulation::is_model_initialized() {
     return get_bmi_model()->is_model_initialized();
 }
+
+/**
+ * Set BMI input variable values for the model appropriately prior to calling its `BMI `update()``.
+ *
+ * @param model_initial_time The model's time prior to the update, in its internal units and representation.
+ * @param t_delta The size of the time step over which the formulation is going to update the model, which might be
+ *                different than the model's internal time step.
+ */
+void Bmi_C_Formulation::set_model_inputs_prior_to_update(const double &model_initial_time, time_step_t t_delta) {
+    std::string std_name;
+
+    std::vector<std::string> in_var_names = get_bmi_model()->GetInputVarNames();
+    // Get the BMI variables' units and associated internal standard names, keeping them in these
+    std::vector<std::string> standard_names(in_var_names.size());
+    std::vector<std::string> bmi_var_units(in_var_names.size());
+    // We may need to sum contributions from multiple time steps before setting, so do that with this (indexes align)
+    std::vector<double> variable_values(in_var_names.size(), 0.0);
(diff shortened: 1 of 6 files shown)


src/realizations/catchment/Bmi_C_Formulation.cpp
```diff
@@ -181,3 +248,55 @@ double Bmi_C_Formulation::get_var_value_as_double(const int& index, const std::s
...
+            throw std::runtime_error(
+                    "BMI input variable '" + in_var_name + "' is an array, which is not currently supported");
+        }
+        // Assign associated standard name and model-side unit string to those collections at corresponding index
+        standard_names.push_back(std_name);
+        bmi_var_units.push_back(get_bmi_model()->GetVarUnits(in_var_name));
+    }
+    // Run separate (hopefully inline) function for getting input values for each variable, potentially by summing
+    // proportionaly contributions from multiple forcing time steps if model time step doesn't align
+    get_forcing_data_ts_contributions(t_delta, model_initial_time, standard_names, bmi_var_units, variable_values);
```
hellkite500, line 296: I'm  starting to wonder if we shouldn't do as much of this as possible at initialization.  Each catchment/formulation has its own forcing object unique to the catchment.  It might be nessicary to take a look at how much  we can do at initialization of that object since this function is called each time step.

hellkite500 (approved):

src/realizations/catchment/Bmi_C_Formulation.cpp
robertbartel: Among other things, that would require knowing in advance exactly what the schedule of model time steps is going to be.  But the `update_until` function in BMI suggests to me that there is a need for dynamically adjusting the next effective time step size during execution.   

Even assuming I'm right, it might still be beneficial to do this during init in certain situations, but I think there's a reasonable chance figuring out how to handle something like that could get complicated.

hellkite500 (approved):

PR #216, opened 2021-03-03
Fix bugs in bmi-cfe and validate the output consistency between bmi-c…

stcui007:
…fe and internal cfe.

Fixing bugs in bmi-cfe codes with validation test.

## Additions

data/bmi/c/cfe/cat_87_bmi_config.txt_old
data/example_realization_config.json_old
data/example_realization_config_w_bmi_c.json_old

## Removals

-

## Changes

data/bmi/c/cfe/cat_87_bmi_config.txt
data/example_realization_config.json
data/example_realization_config_w_bmi_c.json
extern/cfe/src/bmi_cfe.c
extern/cfe/src/cfe.c

## Testing

1. Validation testing: bmi-cfe code and internal ngen cfe code produce identical results.

## Notes

To run the testing,  the modified configuration files should be used.
Previous config files are renamed "old"

## Todos

-

## Checklist

- [ ] PR has an informative and human-readable title
- [ ] Changes are limited to a single goal (no scope creep)
- [ ] Code can be automatically merged (no conflicts)
- [ ] Code follows project standards (link if applicable)
- [ ] Passes all existing automated tests
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments
- [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [ ] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

## Testing checklist (automated report can be put here)

1.

### Target Environment support

- [ ] Linux

PR #220, opened 2021-03-11
Fix bugs in bmi-cfe and internal cfe codes

stcui007:
This PR replaces PR #216
It fixes  bugs in bmi-cfe and internal cfe codes found during the bmi-cfe testing and validation processes.

## Additions

-

## Removals

-

## Changes

extern/cfe/src/bmi_cfe.c
extern/cfe/src/cfe.c

## Testing

1. Thorough validation test performed

## Notes

- Some config files may possibly need modification

## Todos

-

## Checklist

1.

- [ ] Linux

### DIFF
--- a/extern/cfe/src/bmi_cfe.c
+++ b/extern/cfe/src/bmi_cfe.c
@@ -1032,7 +1032,7 @@ static int Get_value_ptr (Bmi *self, const char *name, void **dest)
 
     if (strcmp (name, "RAIN_RATE") == 0) {
         cfe_model* model = (cfe_model *)(self->data);
-        *dest = (void*)(model->forcing_data_precip_kg_per_m2 + model->current_time_step);
+        *dest = (void*)(model->forcing_data_precip_kg_per_m2 + model->current_time_step - 1);
         return BMI_SUCCESS;
     }
 
--- a/extern/cfe/src/bmi_cfe.c
+++ b/extern/cfe/src/bmi_cfe.c
@@ -188,6 +188,7 @@ int read_init_config(const char* config_file, cfe_model* model, double* alpha_fc
     int is_alpha_fc_set = FALSE;
     int is_soil_storage_set = FALSE;
     int is_K_nash_set = FALSE;
+    int is_K_lf_set = FALSE;
 
     // Keep track these in particular, because the "true" storage value may be a ratio and need both storage and max
     int is_gw_max_set = FALSE;
@@ -332,6 +333,11 @@ int read_init_config(const char* config_file, cfe_model* model, double* alpha_fc
             is_K_nash_set = TRUE;
             continue;
         }
+        if (strcmp(param_key, "K_lf") == 0) {
+            model->K_lf = strtod(param_value, NULL);
+            is_K_lf_set = TRUE;
+            continue;
+        }
         if (strcmp(param_key, "nash_storage") == 0) {
             nash_storage_string_val = strdup(param_value);
             is_nash_storage_string_val_set = TRUE;
@@ -426,6 +432,12 @@ int read_init_config(const char* config_file, cfe_model* model, double* alpha_fc
         return BMI_FAILURE;
     }
     if (is_K_nash_set == FALSE) {
+#if CFE_DEGUG >= 1
+        printf("Config param 'K_nash' not found in config file\n");
+#endif
+        return BMI_FAILURE;
+    }
+    if (is_K_lf_set == FALSE) {
 #if CFE_DEGUG >= 1
         printf("Config param 'K_nash' not found in config file\n");
 #endif
--- a/extern/cfe/src/cfe.c
+++ b/extern/cfe/src/cfe.c
@@ -77,9 +77,8 @@ extern void init_soil_reservoir(cfe_model* cfe, double alpha_fc, double max_stor
             (upper_lim - lower_lim);
     // lateral flow parameters --------------------------------------------------------
     // 0.0 to deactiv. else =lateral_flow_linear_reservoir_constant;   // m per ts
-    //cfe->soil_reservoir.coeff_secondary = get_K_lf_for_time_step(cfe);
-    //Klf is a calibration parameter in the ngen code
-    cfe->soil_reservoir.coeff_secondary = 0.01;
+    // TODO: look at whether K_lf needs to be a derived (i.e., via get_K_lf_for_time_step()) or explicit parameter
+    cfe->soil_reservoir.coeff_secondary = cfe->K_lf;
     // 1.0=linear
     cfe->soil_reservoir.exponent_secondary = 1.0;
     // making them the same, but they don't have 2B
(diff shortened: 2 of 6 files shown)


extern/cfe/src/bmi_cfe.c
```diff
@@ -1032,7 +1032,7 @@ static int Get_value_ptr (Bmi *self, const char *name, void **dest)
 
     if (strcmp (name, "RAIN_RATE") == 0) {
         cfe_model* model = (cfe_model *)(self->data);
-        *dest = (void*)(model->forcing_data_precip_kg_per_m2 + model->current_time_step);
+        *dest = (void*)(model->forcing_data_precip_kg_per_m2 + model->current_time_step - 1);
```
robertbartel, line 1035: I don't fully understand why this is correct.  In the primary `run` function, rainfall input is used as the start for calculations.  In particular:

`double timestep_rainfall_input_m = model->forcing_data_precip_kg_per_m2[model->current_time_step];`

I am having trouble wrapping my head around there being an offset when outputting but not when calculating.  Can this be further explained/clarified?  

Also, if this is changed as shown, wouldn't we need to do the same thing to `SURFACE_PRESSURE` a few lines below (again, I think it just isn't being used now, but that doesn't mean it won't be later)?

extern/cfe/src/cfe.c
```diff
@@ -77,7 +77,9 @@ extern void init_soil_reservoir(cfe_model* cfe, double alpha_fc, double max_stor
             (upper_lim - lower_lim);
     // lateral flow parameters --------------------------------------------------------
     // 0.0 to deactiv. else =lateral_flow_linear_reservoir_constant;   // m per ts
-    cfe->soil_reservoir.coeff_secondary = get_K_lf_for_time_step(cfe);
+    //cfe->soil_reservoir.coeff_secondary = get_K_lf_for_time_step(cfe);
+    //Klf is a calibration parameter in the ngen code
+    cfe->soil_reservoir.coeff_secondary = 0.01;
```
robertbartel, line 82: Even if `Klf` is not being calculated right currently with that function, hard-coded this value isn't correct.  Also, why that particular value?

robertbartel (changes requested):

stcui007:
For the first question: if you look at the aorc forcing data, the
precip_rate (last column) is one time step ahead of APCP_surface, which is
equivalent of rainfall. In the Ngen framework, the precip_rate is used. So
in the original CFE model, the alignment is such that the first non-zero
rain_rate appears at time 2015-12-02 08:00:00. The fix, though seems odd,
aligns the rain_rate at the right time.

For the second question, this is the first step in resolving the
disagreement between BMI-CFE and the original CFE. Although not elegant, it
produces consistent results with previously calibrated output thus
demonstrating correctness of BMI-CFE code. I agree that this can be
improved. There are possibly other places where parameters are hardcoded. I
haven't gotten around to check all of them.

Thanks for the careful checking.
.

stcui007:
I forgot to mention, the particular value for Klf was used in the original t-shirt model.

extern/cfe/src/bmi_cfe.c
robertbartel: > For the first question: if you look at the aorc forcing data, the
> precip_rate (last column) is one time step ahead of APCP_surface, which is
> equivalent of rainfall. In the Ngen framework, the precip_rate is used. So
> in the original CFE model, the alignment is such that the first non-zero
> rain_rate appears at time 2015-12-02 08:00:00. The fix, though seems odd,
> aligns the rain_rate at the right time.

I don't quite understand why it's correct for the input data and calculated outputs to not apply to the same time step, but that isn't necessarily something that needs to be resolved right now.  It leaves it unclear to me, though, whether the time step shift should apply to other forcing data values, in particular to `SURFACE_PRESSURE` since right now there's code in place specifically for that one.

extern/cfe/src/cfe.c
robertbartel: > For the second question, this is the first step in resolving the
> disagreement between BMI-CFE and the original CFE. Although not elegant, it
> produces consistent results with previously calibrated output thus
> demonstrating correctness of BMI-CFE code. I agree that this can be
> improved. There are possibly other places where parameters are hardcoded. I
> haven't gotten around to check all of them.

Wouldn't hard-coding the value like this cause the BMI CFE implementation to deviate from the original CFE if any other set of parameters is used?

robertbartel: Actually, I see now in the latest version of Fred's code (that I have) where this is indeed hard-coded, but with contradicting comments (on about line 407):

`soil_reservoir.coeff_secondary=0.01;  // 0.0 to deactiv. else =lateral_flow_linear_reservoir_constant;   // m per h`

This has the look of something done as a temporary setting to work around or test something specific, then later be removed.  Perhaps we should ask Fred to clarify how this should actually operate.

stcui007:
For isolated run, we have been using this value to compare NGEN (internal
CFE) to Fred's t-shirt code output. So we have established confidence for
output flows at this value for BMI-CFE, internal CFE, and the original
t-shirt code..

For parameter calibration, Klf is allowed to vary but that's a different
story.

stcui007:
This one is a bit tricky. For the BMI code, all the flow values are
correctly calculated with the code as it is. If you look at the output
without doing the shift, you can see the only column that has the
misalignment is the rain_rate, which is shifted by  1 hour. With the shift
index in rain_rate in the BMI code, the BMI-CFE output and the internal CFE
output match perfectly.
At the moment, my interpretation is as I said above, i,e,, the precip_rate
occurs one hour ahead in the AORC forcing. Whether the BMI-CFE
output-handling implicitly misaligns the  rain_rate with other flow
variables may need us to think a little deeper.

stcui007:
Looks like the two test failures were caused by hard coded ASSERT values in Bmi_C_Formulation_Test.cpp. Any time there is code change, or initial value change, these numbers would change. Maybe this can be modified so it won't demand the output values to be some specific numbers?

robertbartel:
@stcui007, it looks like for the internal version of CFE, we are using a static, but configurable value for `Klf`.  I think this makes the most sense, if we are not going to derive `Klf` from the other parameters via that function (though we should later consult with Fred on which is more appropriate).

I went ahead and made changes, based off your `bmi_cfe_new` branch, to add `Klf` as a read config param for the BMI CFE model, added to the config files, and updated the unit tests.  These are now in a `bmi_cfe_new` branch in my fork.   After you fetch my fork, you should be able to do a fast-forward merge of my branch to yours, then push the updates to your fork.   I think you already have my fork as a remote, so the commands should be something like:

```
git fetch robertbartel
git checkout bmi_cfe_new
git merge --ff-only robertbartel/bmi_cfe_new
git push origin bmi_cfe_new
```

I don't know exactly how your have your remotes named, so it's possible `robertbartel` and/or `origin` might need to be replaced with whatever you are using, but it should be those if you just used defaults.

That should get the changes from my branch, merge them to yours (but only if you haven't changed anything else from the point in your work I started from), then push the updated version to your fork (which will update the PR).

stcui007:
Thanks, Bobby.

Yes, we should ask Fred about it. Also notice that the Klf contains the
multiplier as a factor, which is a calibratable parameter in NWM. Here we
calibrate Klf instead if needed.

robertbartel (approved):

PR #223, opened 2021-03-23
Bmi dev

madMatchstick:

- Complete variable (grid and location) & grid (type, rank, size, etc) BMI function definitions to CFE (Tshirt) model
- Add these & other previously defined information based BMI functions to bmi-cfe unit tests

## Changes

- extern/cfe/src/bmi_cfe.c
- include/realizations/catchment/Bmi_C_Adapter.hpp 
- src/realizations/catchment/Bmi_C_Adapter.cpp
- test/realizations/catchments/Bmi_C_Adapter_Test.cpp 

## Testing

- 10 simple gtests added to Bmi_C_Adapter_Test suit under automated test target 'test_bmi_c'

## Notes

- run test_bmi_c executable from locally generated build directory's /test folder (e.g. ngen/cmake-build-debug/test/) rather than from project root directory.

## Checklist

[ RUN      ] Bmi_C_Adapter_Test.GetVarGrid_0_a
[       OK ] Bmi_C_Adapter_Test.GetVarGrid_0_a (1 ms)
[ RUN      ] Bmi_C_Adapter_Test.GetVarGrid_0_b
[       OK ] Bmi_C_Adapter_Test.GetVarGrid_0_b (2 ms)
[ RUN      ] Bmi_C_Adapter_Test.GetVarLocation_0_a
[       OK ] Bmi_C_Adapter_Test.GetVarLocation_0_a (2 ms)
[ RUN      ] Bmi_C_Adapter_Test.GetVarLocation_0_b
[       OK ] Bmi_C_Adapter_Test.GetVarLocation_0_b (1 ms)
[ RUN      ] Bmi_C_Adapter_Test.GetVarUnits_0_a
[       OK ] Bmi_C_Adapter_Test.GetVarUnits_0_a (2 ms)
[ RUN      ] Bmi_C_Adapter_Test.GetVarType_0_a
[       OK ] Bmi_C_Adapter_Test.GetVarType_0_a (10 ms)
[ RUN      ] Bmi_C_Adapter_Test.GetVarNbytes_0_a
[       OK ] Bmi_C_Adapter_Test.GetVarNbytes_0_a (2 ms)
[ RUN      ] Bmi_C_Adapter_Test.GetGridType_0_a
[       OK ] Bmi_C_Adapter_Test.GetGridType_0_a (1 ms)
[ RUN      ] Bmi_C_Adapter_Test.GetGridRank_0_a
[       OK ] Bmi_C_Adapter_Test.GetGridRank_0_a (2 ms)
[ RUN      ] Bmi_C_Adapter_Test.GetGridSize_0_a
[       OK ] Bmi_C_Adapter_Test.GetGridSize_0_a (2 ms)

- [x] Linux


extern/cfe/src/bmi_cfe.c
```diff
@@ -778,123 +804,67 @@ static int Get_adjusted_index_for_variable(const char *name)
...
 {
+
+    // Check to see if in output array first
+    for (int i = 0; i < OUTPUT_VAR_NAME_COUNT; i++) {
+        if (strcmp(name, output_var_names[i]) == 0) {
+            *grid = output_var_grids[i];
+            return BMI_SUCCESS;
+        }
+    }
+    // Then check to see if in input array
```
hellkite500, line 859: Is a variable ever possible an output AND and input variable?

extern/cfe/src/bmi_cfe.c
```diff
@@ -956,26 +926,42 @@ static int Get_var_itemsize (Bmi *self, const char *name, int * size)
...
 
 
-// TODO: complete implementation
-static int Get_var_location(Bmi *self, const char *name, char *location)
+static int Get_var_location (Bmi *self, const char *name, char * location)
 {
+    // Check to see if in output array first
+    for (int i = 0; i < OUTPUT_VAR_NAME_COUNT; i++) {
+        if (strcmp(name, output_var_names[i]) == 0) {
+            strncpy(location, output_var_locations[i], BMI_MAX_UNITS_NAME);
```
hellkite500, line 934: Is there a guarantee that that the largest location string is smaller than the largest unit name string?  If not,  using` BMI_MAX_UNITS_NAME` for the buffer size here may be problematic.  Do we need another macro for `BMI_MAX_LOCATION_SIZE` to be explicit?

extern/cfe/src/bmi_cfe.c
```diff
@@ -956,26 +926,42 @@ static int Get_var_itemsize (Bmi *self, const char *name, int * size)
...
+            return BMI_SUCCESS;
+        }
+    }
+    // Then check to see if in input array
```
hellkite500, line 938: Any particular reason to check output then input?  Could the variable possibly be in both?

extern/cfe/src/bmi_cfe.c
```diff
@@ -956,26 +926,42 @@ static int Get_var_itemsize (Bmi *self, const char *name, int * size)
...
 
 
+
 static int Get_var_units (Bmi *self, const char *name, char * units)
 {
     // Check to see if in output array first
     for (int i = 0; i < OUTPUT_VAR_NAME_COUNT; i++) {
         if (strcmp(name, output_var_names[i]) == 0) {
-            strncpy(units, output_var_units[i], BMI_MAX_TYPE_NAME);
+            strncpy(units, output_var_units[i], BMI_MAX_TYPE_NAME); //JG: should be BMI_MAX_UNITS_NAME
```
hellkite500, line 957: Use `BMI_MAX_UNITS_NAME`

src/realizations/catchment/Bmi_C_Adapter.cpp
```diff
@@ -402,6 +402,54 @@ std::string Bmi_C_Adapter::GetVarUnits(std::string name) {
     return units_str;
 }
 
+std::string Bmi_C_Adapter::GetVarLocation(std::string name) {
+    char location_c_str[BMI_MAX_UNITS_NAME];
+    int success = bmi_model->get_var_location(bmi_model.get(), name.c_str(), location_c_str);
+    if (success != BMI_SUCCESS) {
+        throw std::runtime_error(model_name + " failed to get variable location for " + name + ".");
+    }
+    std::string location_str(location_c_str);
```
hellkite500, line 411: IIRC you could probably simplify this to just `return std::string( location_c_str )`

src/realizations/catchment/Bmi_C_Adapter.cpp
```diff
@@ -402,6 +402,54 @@ std::string Bmi_C_Adapter::GetVarUnits(std::string name) {
     return units_str;
 }
 
+std::string Bmi_C_Adapter::GetVarLocation(std::string name) {
+    char location_c_str[BMI_MAX_UNITS_NAME];
```
hellkite500, line 406: Again, might want to consider a `BMI_MAX_LOCATION_SIZE`

test/realizations/catchments/Bmi_C_Adapter_Test.cpp
```diff
@@ -202,7 +247,7 @@ TEST_F(Bmi_C_Adapter_Test, Update_0_h) {
 
 /** Test that Schaake Runoff output variable values can be retrieved. */
 TEST_F(Bmi_C_Adapter_Test, GetValue_0_a_0) {
-    int out_var_index = 0;
+    int out_var_index = 0;  //JG: out_var_index = 1; 0->RAIN_RATE
```
hellkite500, line 250: reconcile this comment for test `GetValue_0_a_[ 0 - 4 ]`

hellkite500 (changes requested):
Like the test!!!  Need to make sure we are using the right size buffers for unit/type/location consistently, but otherwise looking good to me!

extern/cfe/src/bmi_cfe.c
madMatchstick: Yes, it is possible.  In fact many of the [BMI examples](https://github.com/csdms/bmi-example-c/blob/master/heat/bmi_heat.c) have a single variable defined as both input and output.  Don't see this separation being problematic for CFE as, 

- There are no true bmi input variables
- & even if there was an identical i/o var, they would share the same "information characteristics" (type, units, grid (id), grid_type, location, etc.)

extern/cfe/src/bmi_cfe.c
madMatchstick: Including a BMI_MAX_LOCATION_NAME may be best practice.  The bmi-examples simply use BMI_MAX_UNITS_NAME as size buffer for a handful of bmi functions defs (e.g. get_var_location).

extern/cfe/src/bmi_cfe.c
madMatchstick: Variables defined only as output.

extern/cfe/src/bmi_cfe.c
madMatchstick: ok

src/realizations/catchment/Bmi_C_Adapter.cpp
madMatchstick: ok

test/realizations/catchments/Bmi_C_Adapter_Test.cpp
madMatchstick: ok

src/realizations/catchment/Bmi_C_Adapter.cpp
madMatchstick: ok - this works

hellkite500 (dismissed):
Looks good to me. I want to give @robertbartel a chance to take look before we merge.

extern/cfe/src/bmi_cfe.c
robertbartel: This looks like it is templated from my original logic for `Get_var_units`.  It seemed reasonable for that to potentially  provide unit information for both input and output variables.  I assume the same is true for location information

There aren't currently input variables defined, but there were at one point and very easily could be again (I can think of at least one specific change worth considering that would require adding them).  I.e.,  pre-writing logic to account for there being a collection of input variables, and just having the collection be empty for now, seems sensible.

extern/cfe/src/bmi_cfe.c
```diff
@@ -1271,25 +1257,26 @@ Bmi* register_bmi(Bmi *model) {
...
-        model->get_grid_face_edges = Get_grid_face_edges;    // TODO: needs finished implementation
-        model->get_grid_face_nodes = Get_grid_face_nodes;    // TODO: needs finished implementation
-        model->get_grid_nodes_per_face = Get_grid_nodes_per_face;    // TODO: needs finished implementation
+        model->get_grid_size = Get_grid_size;    
+        model->get_grid_rank = Get_grid_rank;    
+        model->get_grid_type = Get_grid_type;    
+
+        model->get_grid_shape = NULL;
+        model->get_grid_spacing = NULL;
+        model->get_grid_origin = NULL;
```
robertbartel, line 1266: What motivated removing function pointer assignments to the "empty" function implementations - e.g., `model->get_grid_x = Get_grid_x` - and instead setting the pointer to `NULL`?  Off the top of my head, I don't recall what would happen if something tried to call a function using a null pointer like this, but I would (i.e. did) think it makes more sense to have the empty implementation that just immediately returns the error code.  Is there something in the spec suggesting this is a more correct approach?

madMatchstick: I was simply following [csdms BMI C example](https://github.com/csdms/bmi-example-c/blob/master/heat/bmi_heat.c).  Most of the model grid functions aren't applicable because we defined grid type to be of "scalar".  
Now looking at [BMI best practices](https://bmi.readthedocs.io/en/latest/bmi.best_practices.html#best-practices), which states that all BMI functions must be implemented, even if empty.
Best to "do as I say, not what I do"?  I'll go ahead revert back to the original setup for these, using the empty functions assignments. -thanks

hellkite500: Yeah, better than the `segfault` that would likey occur and drive someone crazy to figure out why...

hellkite500 (approved):

PR #225, opened 2021-03-24
Precision_Guidelines_for_Ngen_Code_Validation

stcui007:
A description on what to expect after a major Ngen code change or update.
Document in pdf format.

PR #226, opened 2021-03-25
Ngen code validation guidelines in markdown

stcui007:
A description on what to expect after a major Ngen code change or update

## Additions

- doc/Precision_Guidelines_for_Ngen_Code_Validation.md

## Removals

-

## Changes

-

## Testing

1.

## Notes

- Can replace PR #225 or keep both

## Todos

-

## Checklist

1.

- [ ] Linux


doc/Precision_Guidelines_for_Ngen_Code_Validation.md
```diff
@@ -0,0 +1,36 @@
...
...le (float, 4-byte) precision.  It’s possible that some entire models or components that are included in the future may also have a float precision, which is not uncommon in legacy model code (and is faster than double precision code).  Advanced models may even be able to switch between float and double precision implementations.  Double and float variables are typically now stored with 15- and 7-digit  precision (respectively), but certain operations or uses of the variables may result in an accumulation of error after multiple timesteps, and a loss of matching ability (even when code is correctly implemented) at slightly lower precision levels. _For this reason, 12 digits  is proposed as a standard for validating double values, and 5 digits is proposed as a standard for float variables_ (In practice, the precision for double values could be lower if float variables are involved at any stage of the calculations). While a match at these levels does not entirely rule out rule coding error in a new model version, it would give confidence that the chances of an error are low, and that if an error exists, its impact on simulation outcomes will likely be minimal even in long simulations.
```
hellkite500, line 9: >does not entirely rule out rule coding error

does not entirely rule out coding error?

doc/Precision_Guidelines_for_Ngen_Code_Validation.md
```diff
@@ -0,0 +1,36 @@
...
+
+
+**Possible reasons for precision discrepancies between CFE and BMI-CFE:**
+
+The double precision numbers have an accuracy of 15 decimal digits, the loss of accuracy from rounding off error is expected, but precisely how many digit loss probably varies and depends on the specific code. Since BMI-CFE code has additional interface and difference in coding details, we hypothesize the memory allocation, cache usage, data sets loading into the registers may be different. These likely affect the rounding off processes and cause the least significant digits to differ. According to some literature, computer hardware architecture, compiler, even optimization may all affect numerical precision.
```
hellkite500, line 23: >memory allocation, cache usage, datasets loading into the registers

These should have no effect on rounding error,  those are caused by operations on on the floats.  Copying bits, in the absence of bit flip or bit rot, should always give the same the bits, hence the same precision.  The only time this would happen under normal circumstances is casting between single/double precision.  If the interfaces don't do this, they shouldn't the be cause of precision loss.  But simply doing operations in different orders can definitely cause differences in rounding error and hence precsion.

doc/Precision_Guidelines_for_Ngen_Code_Validation.md
```diff
@@ -0,0 +1,36 @@
...
+
+The double precision numbers have an accuracy of 15 decimal digits, the loss of accuracy from rounding off error is expected, but precisely how many digit loss probably varies and depends on the specific code. Since BMI-CFE code has additional interface and difference in coding details, we hypothesize the memory allocation, cache usage, data sets loading into the registers may be different. These likely affect the rounding off processes and cause the least significant digits to differ. According to some literature, computer hardware architecture, compiler, even optimization may all affect numerical precision. 
+        
+In our CFE and BMI-CFE validation example, one particular case, deserves attention, which is: a key input variable, the rain rate, or specifically: precip_kg_per_m2, is a float point number, this could potentially reduces the precision in comparison between the two models even though all other variables are double, in consideration of rounding-off error in machine precision. There are many excellent discussions on rounding-off errors in literature due to machine precision, we refer interested readers to literature<sup>1-4</sup> for details.
```
hellkite500, line 25: One, `precip_kg_per_m2` is not _directly_ rain rate (or it would have a unit of time in the units?)

Two,  if the same float was provided to both implementations, are you  concluding that casting that number lost precision for the entire model which makes comparing outputs nessicary at the single precision level of accuracy (5 digits)?

doc/Precision_Guidelines_for_Ngen_Code_Validation.md
```diff
@@ -0,0 +1,36 @@
...
+
+On the other hand, the BMI-CFE and the original CFE demonstrate remarkable stability with 8-9 decimal point accuracy even after 720 time steps (hours). What are the possible reasons? We hypothesize two possible causes:
+1. The codes do not use space grids, all flow values are single valued for the catchment, i.e., there is no instability introduced by spatial matrices such as those in solving finite difference equations.
+2. Mass balance: In the evolution in time, the flows are constrained by mass balance. The mass balance, like energy conservation in classical molecular dynamics, should constrain the flow variation to a certain range, thus limiting the error propagation magnification in time (possibly through error cancellation).
```
hellkite500, line 29: Did the test actually account for/accumulate mass balance error?

doc/Precision_Guidelines_for_Ngen_Code_Validation.md
```diff
@@ -0,0 +1,36 @@
...
+
+
+References and Links  
+1. [What Every Computer Scientist Should Know About Floating-Point Arithmetic](https://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html).
```
hellkite500, line 33: I  like this reference, I go back to it often!

doc/Precision_Guidelines_for_Ngen_Code_Validation.md
stcui007: It should rules out major ones, perhaps not 100 percent about insignificant ones. Maybe coding error is too strong a word. It guarantees that if any difference exists, its impact on practical applications can be neglected without significant consequences (when certain precision is reached).

doc/Precision_Guidelines_for_Ngen_Code_Validation.md
stcui007: My thinking was, taking for example array operation, if in one case you load elements 1 to 99, do some calculations and save the result into memory. In another case, you load element 2 to 100, and do some calculations, save the result into memory, should the round off error be different?

Could you think of some good reasons these might have a role in round off error? If not we need to delete it.

doc/Precision_Guidelines_for_Ngen_Code_Validation.md
stcui007: One: you are right, kg/m^2 translate into mm for water, according to hydrologist. I can delete the word rain rate. Thanks for the correction.

Two: Not entirely sure. If casting is completely random, we are left with  7 reliable digits. Our results seem to indicate we can achieve slightly better precision, 8 to 9 digits. Not sure exactly why.

doc/Precision_Guidelines_for_Ngen_Code_Validation.md
stcui007: Based on my experience with molecular dynamics modeling, energy conservation law ensures a few most significant digits is unchanged during time evolution. I believe mass balance does the same thing for flows in time evolution. It keeps the flow values from drifting away in one direction. A correct code should have this property. We did not specifically test the mass balance error. In theory, if the computer is infinitely accurate, mass balance should be strictly followed. My guess is that the mass balance error is random and related to round off error. For a stable algorithm, a few most significant digits should be unchanged during time evolution.

doc/Precision_Guidelines_for_Ngen_Code_Validation.md
stcui007: The same for me. though I probably haven't read as thoroughly as you did.

doc/Precision_Guidelines_for_Ngen_Code_Validation.md
hellkite500: Sorry, was just trying to point out the typo "rule out rule coding"

doc/Precision_Guidelines_for_Ngen_Code_Validation.md
hellkite500: @donaldwj toughts?

doc/Precision_Guidelines_for_Ngen_Code_Validation.md
donaldwj: One thing to note in any code section where roundoff error is suspected is that the precision of floating point formats is much greater close to the value zero. This means that using normalized numbers can greatly reduce representation based error.

stcui007: precip_kg_per_m2, although declared as float in CFE code, it actually has many more than 7 significant digits and in scientific notation in the forcing file. But the exponent has 2 digits not 3 as in standard C++ double.

hellkite500 (approved):
Good place to start, we can get this added to the docs and iterate it in the future as needed.

PR #232, opened 2021-04-14
Update example config, fixes #231

### LINKED ISSUE #231: Build and Run Model workflow fails
Simple Lumped changes were merged in #208 cause the workflow (which was implemented after that PR IIRC) to fail.

## Current behavior
Workflow Fails

## Expected behavior
Workflow Succeeds

## Fix
Need to update  configuration file used by the workflow  to include simple lumped parameters for

- `smax parameter` that is used in the storage distribution function
- `gw_max_storage` parameter for building the groudwater/slow flow reservoir
- `nash_cascade_max_storage` parameter for building the quick flow reservoirs

and  remove

- `max_storage_parameter`

hellkite500:
SImple lumped updates changed required parameters, but the example realization wasn't properly updated.  This updates the example config so it should run without issue.

Closes #231

### DIFF
--- a/data/example_realization_config.json
+++ b/data/example_realization_config.json
@@ -100,7 +100,10 @@
                       1.0
                   ],
                   "storage": 1.0,
-                  "max_storage": 1000.0,
+                  "gw_storage": 1.0,
+                  "gw_max_storage": 10.0,
+                  "nash_max_storage": 2.0,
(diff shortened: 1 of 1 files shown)


hellkite500:
Note #233 explains why the test is still failing.

robertbartel (approved):

PR #250, opened 2021-05-17
Submodule and Helper Build Setup for BMI Models

robertbartel:
Submodule and helper build setup for OWP-maintained BMI models that are not contained directly within the NGen repo.  

This primarily sets up the `alt-modular` repo as a Git submodule, then creates a separated CMake config for building the shared libraries for CFE and TOPMODEL.  It also includes a bit of explanation in the helper README.  These are all under the `extern/alt-modular/` directory, with the submodule itself being at `extern/alt-modular/alt-modular/`.

Note that this requires pulling the submodule via Git's SSH functionality, which in turn requires a private key be set up for the user in Git for this.  Because of the current status and restrictions of `alt-modular`, it was either that or a Git Personal Access Token for HTTPS, and the SSH key seemed more standard.

PR #251, opened 2021-05-18
Integrate BMI formulations with framework ET kernel

robertbartel:
Integrates the BMI formulation types with the internal combination method ET kernel, along with a few other related changes.

### ~Initial Draft PR~
~These changes require the `alt-modular` CFE implementation to work, as they need a model that actually accepts as input the ET kernel's output.  As such, creating as a draft PR for now.  Once #250 is approved and merged into this, and some updates are made to the necessary Github Actions workflow(s), this can move out of _draft_ status.~

### Additions
- initial `calc_et` implementation in `Bmi_Formulation`, somewhat hardcoded for specific use-case relevant to upcoming milestone test (e.g., combination method, some fixed parameters, etc.)
- ET-specific namespaces for currently used combination kernel and essential components
- `*.cpp` files for currently used combination ET kernel and essential components
- ET-specific static library 
- helper functions in `Bmi_Formulation` for working with BMI variable names
- support for configuring name of function pointer registration function for BMI C models
- NGen "standard" mapped variable name for BMI model ET input variables
- variable name mappings parameter in example realization configs as required for `alt-modular` CFE implementation
- variable name mappings parameter in automated tests as required for `alt-modular` CFE implementation

### Changes
- `forcing_file` realization config param now optional rather than required
- `calc_et` function signature in `Et_Accountable` interface (and implementing subtypes) now has no parameters and expects class state to provide everything needed for calculations
- a few minor fixes to ET kernel components
- modifications to the functions used within `Bmi_C_Formulation.get_response`  to set BMI input variables prior to making the BMI `update` call
- declaration (though not definition) of the `get_var_value_as_double` convenience function for BMI output variable retrieval was moved up from `Bmi_C_Formulation` to `Bmi_Formulation`
- corrected reference in BMI usage documentation referencing where NGen "standard" variable names are defined
- switches to `alt-modular` CFE implementation instead of internally maintained CFE lib
- documentation as needed for other changes
- other required changes to unit tests as a result of code changes


robertbartel:
FWIW, I'm a little concerned about something in the ET kernel params and calculations.  This type of warning gets printed a large number of times during the BMI C formulation unit tests. 

> Invalid value of specific humidity with no supplied rel. humidity in ET calc. function:
> Relative Humidity: -99.900000 percent
> Specific Humidity: 0.009300 kg/kg

 There are not any test failures because the ET kernel code automatically makes an adjustment, but the comments in the kernel code specifically say that "this is bad" and should not happen.  See around line 417 in [EtCalcProperty.cpp](src/models/kernels/evapotranspiration/EtCalcProperty.cpp) for where this originates.  

As far as I can tell, things leading up to this block of code are operating correctly/as expected:  
- the `et_forcing.relative_humidity_percent` is deliberately hard-coded to `-99.9` to dictate use of specific humidity
  - this is set in `Bmi_Formulation.calc_et()`, which eventually calls the kernel
- the right forcing values for specific humidity, pressure, and temperature appear to be getting used for the `air_saturation_vapor_pressure_Pa` and `air_actual_vapor_pressure_Pa` calculations

stcui007:
My examination of the EtCalProperty.hpp (I don't see EtCalProperty.cpp)
code logic suggests that variable is not used because aorc_yes is set to
True.

Shengting

robertbartel:
Temporarily, due to permissions issues with the _alt-modular_ repo, the actions that run the automated BMI unit tests will still fail (though not the tests themselves, which don't actually get called).  These have have been successfully run locally on a Mac, and will be shortly on a Linux machine in preparation for the upcoming milestone test.

As such, moving this out of _draft_ status.

test/realizations/catchments/Bmi_C_Adapter_Test.cpp
```diff
@@ -226,9 +238,12 @@ TEST_F(Bmi_C_Adapter_Test, Update_0_g) {
     adapter.Finalize();
 }
 
-/** Test that the update function works fails after exceeding end time. */
+/** Test that the update function fails after exceeding end time. */
+/* This actually doesn't work currently, but this is due to a change in the way the example model is implemented.
+ * Re-add this test if/once we have a model that doesn't work beyond its time step.
```
hellkite500, line 243: Does this test actually need to be diabled?  This comment makes it seem so, but the test still is active.  We can diable the test by prepending `DISABLED_` to the test name, i.e.
`TEST_F(Bmi_C_Adapter_Test, DISABLED_Update_0_h) {`

hellkite500 (commented):
just a couple preliminary comments

src/models/kernels/evapotranspiration/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+cmake_minimum_required(VERSION 3.10)
```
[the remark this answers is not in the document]
robertbartel: Not necessarily.  There are features of CMake that aren't introduced until version 3.12, which we are using in some other CMake config files.  However, there are not any in this file.  

I personally think it's fine to have the requirement strictly reflect the file it's in, but I wouldn't argue too strongly against across-the-board consistency (though I think we need to introduce some automation if we decide to got that route).

test/realizations/catchments/Bmi_C_Adapter_Test.cpp
robertbartel: I'd forgotten about the `DISABLED_` capability.  I'm going to change this one and a few others that I'd commented out for similar reasons.

data/example_realization_config_w_bmi_c__linux.json
```diff
@@ -49,10 +49,22 @@
                     "name": "bmi_c",
                     "params": {
                         "model_type_name": "bmi_c_cfe",
-                        "library_file": "./extern/cfe/cmake_cfe_lib/libcfemodel.so",
+                        "library_file": "./extern/alt-modular/cmake_am_libs/libcfebmi.so",
```
hellkite500, line 52: I'm starting to wonder if we shouldn't take the `ld` approach to library names, that is not including the extension and let the build append the right extension.  Then we would only need a single input file, not one per OS.

include/realizations/catchment/Bmi_Formulation.hpp
```diff
@@ -57,13 +62,157 @@ namespace realization {
...
+            /*
+             * ********************************************************************************************************
+             * Logic based largely on et_wrapper_function() in EtWrapperFunction.hpp, and a little from et_setup() in
+             * EtSetParams.hpp.
+             * ********************************************************************************************************
+             */
+
+            // TODO: Putting made-up values here, but need to figure out where to actually get these from (these taken
+            //  from EtSetParams.hpp, with comments from relevant lines quoted)
+            double made_up_canopy_resistance_sec_per_m = 50.0; // "from plant growth model"
```
hellkite500, line 86: If these are constant, can we make them const static member variables?

test/realizations/catchments/Bmi_C_Adapter_Test.cpp
```diff
@@ -9,14 +9,16 @@
 
 #ifndef BMI_CFE_LOCAL_LIB_NAME
 #ifdef __APPLE__
-#define BMI_CFE_LOCAL_LIB_NAME "libcfemodel.dylib"
+#define BMI_CFE_LOCAL_LIB_NAME "libcfebmi.dylib"
```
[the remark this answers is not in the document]
robertbartel: In this cases it's because the configurations are part of the code.   The realization configurations used in these tests are not written in a separate file.  I can't remember entirely why I did those through the setup function and not from files.

There definitely does need to be some examination of how to improve some of the test design.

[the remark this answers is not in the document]
hellkite500: For some reason I thought this was in the Adapter, not in the Test.  Not as much of a worry, but something to think about.

PR #257, opened 2021-06-09
Add generic adapter specification 

donaldwj:

## Additions

New interface for generic adapter for multi-components connections in arbitrary directions
-

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

- [X ] PR has an informative and human-readable title
- [X ] Changes are limited to a single goal (no scope creep)
- [X ] Code can be automatically merged (no conflicts)
- [X ] Code follows project standards (link if applicable)
- [X ] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

1.

- [ ] Linux

### DIFF
--- a/include/core/adapter/Adapter2D.hpp
+++ b/include/core/adapter/Adapter2D.hpp
@@ -1,0 +1,69 @@
+#ifndef GENERIC_ADAPTER_H
+#define GENERIC_ADAPTER_H
+
+#include <memory>
+#include <string>
+#include <unordered_map>
+
+#include <boost/geometry.hpp>
+#include <boost/geometry/geometries/geometries.hpp>
+#include <HY_Catchment.hpp>
+#include <HY_HydroLocation.hpp>
+
+using namespace hy_features::hydrolocation;
+class GenericAdapter
+{
+    public:
+
+    typedef long time_step_t;
+
+
+    // create type aliases
+    using Models = std::vector<std::string>;
+    using long = time_point_t;
+    using long = time_delta;
+
+    GenericAdapter(std::string adapter_id, Models ls, Models rs);
+
+    virtual ~Adapter2D();
+
+    /** Contribute a flux with id flux_id from source contrib_id at time step t*/
+
+    virtual void add_flux(std::string flux_id, std::string contrib_id, time_point_t t, time_delta t1, double* vals)= 0;
+
+    /** Remove a contributed flux as aggreated between time_steps t1 and t2*/
+
+    virtual void remove_flux(std::string flux_id, std::string requesting_id, time_point_t t1, time_delta t2, double* vals) = 0;
+
+    virtual std::vector< std::pair<double*, std::string> > inspect_flux(std::string flux_id, time_point t)=0;
+
+    /** get the units that the flows are described in */
+    virtual std::string get_flux_units(std::string flux_id)=0;
+
+    protected:
+
+    //Identity is a string.  If implementations NEED a numeric id for indexing,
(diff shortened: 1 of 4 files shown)


include/core/adapter/Adapter2D.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+
+    // create type aliases
+    using Models = std::vector<std::string>;
+    using long = time_point_t;
+    using long = time_delta;
+
+    GenericAdapter(std::string adapter_id, Models ls, Models rs);
+
+    virtual ~Adapter2D();
```
robertbartel, line 28: The class/constructor and the file/destructor don't have the same name.  I am presuming this is not intended.

include/core/adapter/Adapter2D.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+    /** Contribute a flux with id flux_id from source contrib_id at time step t*/
+
+    virtual void add_flux(std::string flux_id, std::string contrib_id, time_point_t t, time_delta t1, double* vals)= 0;
```
robertbartel, line 32: This isn't necessarily something that needs to be changed, but I am curious about it.

My understanding was that this type is a more general abstraction for something like the (HY_Features-specific) _HY_HydroNexus_ (if that isn't correct, then this comment can probably be ignored).  Should we apply the relationship at this point?   Also, is the use of slightly different terminology (e.g., "flux" here versus "flow" in _HY_HydroNexus_) just due to some dialect differences when talking about similar concepts in the different contexts?

include/core/adapter/Adapter2D.hpp
```diff
@@ -0,0 +1,69 @@
...
+#include <memory>
+#include <string>
+#include <unordered_map>
+
+#include <boost/geometry.hpp>
+#include <boost/geometry/geometries/geometries.hpp>
+#include <HY_Catchment.hpp>
+#include <HY_HydroLocation.hpp>
+
+using namespace hy_features::hydrolocation;
```
robertbartel, line 13: Am I missing a reason why the _HY_Catchment.hpp_ and _HY_HydroLocation.hpp_ are being imported, and why the _hy_features::hydrolocation_ namespace is being used?  They don't seem to hurt anything, but I am curious (especially given the elsewhere mentioned name discrepancy) if any intended changes could have been accidentally left out.

robertbartel (changes requested):
There is at least one change related to the name of the new type that I think is needed, and a couple of other questions I had.  

Additionally, it looks like an extra commit - `76e5a60` - got added to the branch by mistake, as its changes seem completely unrelated.  The branch should probably be rebased anyway, so this commit can be dropped during that.

include/core/adapter/Adapter2D.hpp
donaldwj: That would be correct. I will update the branch

include/core/adapter/Adapter2D.hpp
donaldwj: They are imported because this class was made by modifing another and the includes where not removed, Another thing in the branch to correct.

donaldwj:
This request is now very out of date it may be difficult to reintegrate this.

PhilMiller:
This PR is now another 2.5 years out of date. Do you still want to do (something like) this? If not, it's probably best to just close it, to get it off the radar of "things to potentially be reviewed"

PR #264, opened 2021-07-15
Support for formulation that is an aggregate of multiple BMI module components

robertbartel:
A few highlights:

- adds new `ForcingProvider` interface for item that can provide one or more forcings
  - for now, leaves existing `Forcing` member variable of formulation types, but has `Forcing` class implement this new interface
- updates `Bmi_Formulation` to implement `ForcingProvider` to provide its output variables as forcings
- updates BMI formulations to have member variable for storage/lookup of correct forcing provider of individual BMI input variables (while falling back to default of previously-used forcing object)
- adds new `Bmi_Module_Formulation` and `Bmi_Multi_Formulation` for single-module and multi-module usages
- has `Bmi_Multi_Formulation` create sub-formulations as part of former's creation, including appropriately setting up the right data source for BMI input variables, be it either a "main" `Forcing` object or earlier sub-formulation
- creates wrapping type and structure for BMI formulations to self-reference to provide their own ET, while also supporting ET from a different forcing provider based on config


include/realizations/catchment/Bmi_Module_Formulation.hpp
```diff
@@ -0,0 +1,937 @@
...
+            double made_up_vegetation_height_m = 0.12; // "used for unit test of aerodynamic resistance used in Penman Monteith method."
+            double made_up_zero_plane_displacement_height_m = 0.0003; /* 0.1; */  // "0.03 cm for unit testing"
+            double made_up_surface_longwave_emissivity = 1.0; // "this is 1.0 for granular surfaces, maybe 0.97 for water"
+            double made_up_surface_shortwave_albedo = 0.22; /* 0.0; */ // "this is a function of solar elev. angle for most surfaces."
+            double made_up_surface_skin_temperature_C = 12.0; /* 20.0; */  // "from soil thermal model or vegetation model"
+            double made_up_momentum_transfer_roughness_length_m = 0.0; /* 1.0; */ // zero means that default values will be used in routine.
+            double made_up_heat_transfer_roughness_length_m = 0.0; /* 1.0; */ // zero means that default values will be used in routine.
+
+            // TODO: for now, hard-code these things
+            struct et::evapotranspiration_options et_options;
```
hellkite500, line 85: We should add some initializer functions to the et namespace and move this "model specific" code there and simply pass in an empty struct reference and let them be set there.

include/realizations/catchment/Bmi_Multi_Formulation.hpp
```diff
@@ -0,0 +1,300 @@
...
+         *
+         * An @ref std::out_of_range exception should be thrown if the time is not in any time step.
+         *
+         * @param epoch_time The point in time, as a seconds-based epoch time.
+         * @return The index of the forcing time step that contains the given point in time.
+         * @throws std::out_of_range If the given point is not in any time step.
+         */
+        size_t get_ts_index_for_time(const time_t &epoch_time) {
+            // TODO: come back and implement if actually necessary for this type; for now don't use
+            throw runtime_error("Bmi_Singular_Formulation does not yet implement get_ts_index_for_time");
```
hellkite500, line 204: This error message is wrong, should probably be "Bmi_Multi_Formulation"

src/realizations/catchment/Bmi_Multi_Formulation.cpp
```diff
@@ -0,0 +1,266 @@
...
+                throw std::runtime_error(
+                        "Multi BMI cannot be created with module " + module->get_model_type_name() +
+                        " with output variable " + framework_alias +
+                        (var_name == framework_alias ? "" : " (an alias of BMI variable " + var_name + ")") +
+                        " because a previous module is using this output variable name/alias.");
+            }
+            availableData[framework_alias] = module;
+        }
+        module_variable_maps[i] = var_aliases;
+    }
```
hellkite500, line 70: For these long, nested loops, it is often really helpful to put a comment after the closing bracket to indicate which loop has ended, i.e. `} // end for sub_formulations_list` or something to help clearly identify it.

src/realizations/catchment/Bmi_Multi_Formulation.cpp
```diff
@@ -0,0 +1,266 @@
...
+            out_headers[i] = out_headers_json_list[i].as_string();
+        }
+        set_output_header_fields(out_headers);
+    }
+    else {
+        set_output_header_fields(get_output_variable_names());
+    }
+}
+
+const bool &Bmi_Multi_Formulation::get_allow_model_exceed_end_time() const {
```
hellkite500, line 94: I would add a comment (even if redundant with the header) here describing the semantics of `allow_model_exceed_end_time` when it may not be the same across each module.

hellkite500 (dismissed):

hellkite500 (approved):

include/realizations/catchment/Bmi_Module_Formulation.hpp
robertbartel: This is now tracked in #276.

hellkite500 (approved):

Issue #262, opened 2021-07-14
Run Failure (Python Version)


ZacharyWills:
Action Testing fails to run nGen

## Current behavior
Failure to build and run nGen in Actions here:

https://github.com/NOAA-OWP/ngen/actions/runs/1027692345

Problem section:
```
-- Found Boost: /ngen/boost_1_72_0 (found suitable version "1.72.0", minimum required is "1.72.0")  
1561
INFO NGEN_ACTIVATE_PYTHON not defined
1562
INFO Did not find 'PYTHON_INCLUDE_DIR' in ENV
1563
INFO Did not find 'PYTHON_LIBRARIES' in ENV
1564
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:165 (message):
1565
  Could NOT find Python: Found unsuitable version "3.6.8", but required is at
1566
  least "3.6.12" (found , found components: Interpreter)
1567
Call Stack (most recent call first):
1568
  /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:456 (_FPHSA_FAILURE_MESSAGE)
1569
  /usr/share/cmake/Modules/FindPython/Support.cmake:2966 (find_package_handle_standard_args)
1570
  /usr/share/cmake/Modules/FindPython.cmake:393 (include)
1571
  CMakeLists.txt:130 (find_package)
1572

```

## Expected behavior
nGen should build and run in GitHub Actions

## Steps to replicate behavior (include URLs)

1. Run action

ZacharyWills:
I can try and boost the version of python in the image, or run an update prior since it looks like just a minor version update with fix the issue.

However @robertbartel is that ENV something that just needs to be set in the action?

jdmattern-noaa:
Not sure of best method with Docker, but I would recommend setting up a venv in the project root.

ZacharyWills:
This has been fixed! Thanks!

PR #261, opened 2021-07-13
GeoJSON Reader Exception

jdmattern-noaa:
Updated the example catchment_data.geojson property from 'area_sqkm' to 'areasqkm' to match the new hydro-fabric. Added exception to GeoJSON reader whenever a property is not found.

This should fix part of the automated tests that are failing on GitHub.

## Checklist

- [ ] PR has an informative and human-readable title
- [ ] Changes are limited to a single goal (no scope creep)
- [ ] Code can be automatically merged (no conflicts)
- [ ] Code follows project standards (link if applicable)
- [ ] Passes all existing automated tests
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments
- [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [ ] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

## Testing checklist (automated report can be put here)

1.

### Target Environment support

- [ ] Linux

PR #266, opened 2021-07-20
Fix Github Actions Run Failure from #262

robertbartel:
This corrects the issue that caused the failures in automated Github Actions runs as described in #262.  

It also addresses a secondary problem observed after the primary issue was resolved, where the format of the used catchment data file was not consistent with recent updates.

## Changes

- Updates custom Dockerfile for image used in example model run Actions workflow
  - Moves from Python 3.6 to 3.8 and ensures development package is installed
  - Rearranges some instructions to improve use of cache during image builds
  - Specifies a more explicitly indicative image tag (`8.3.2011` rather than `latest`) of parent _centos_ image
  - Installs Numpy package (via yum) and initializes _pybind11_ Git submodule
- Replaces instances of `area_sqkm` to `areasqkm` in _catchment_data.geojson_ to be consistent with expected format


robertbartel:
Curious ... it still fails, but with different, albeit very similar error, complaining about missing the NumPy include dir for all except invalid Python versions:

```
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:165 (message):
1585
  Could NOT find Python (missing: Python_NumPy_INCLUDE_DIRS NumPy) (found
1586
  suitable version "3.8.6", minimum required is "3.6.12")
```

I will experiment further.

data/catchment_data.geojson
```diff
@@ -3,8 +3,8 @@
...
..., [-80.72629344934451, 35.20243818376059], [-80.72433323444952, 35.202050235437646], [-80.72326622164341, 35.201980669569856], [-80.72218392325884, 35.20094136323369], [-80.72203343642354, 35.19869411060445], [-80.72240476606557, 35.197473803873024], [-80.72258571719085, 35.19565796653196], [-80.72488120361596, 35.19452190066924], [-80.72547498310766, 35.19379116746099], [-80.72677829031477, 35.192591067694586], [-80.72685423974872, 35.19182216938071], [-80.72758443862824, 35.18999337824186], [-80.72856218290306, 35.18978600853656], [-80.72889692584329, 35.18868190918505], [-80.72893964298866, 35.188273788489035], [-80.72975261128205, 35.18683119874651], [-80.73432689101276, 35.187490901526644], [-80.73553246927948, 35.18819166991151], [-80.7373756356953, 35.18571684336697], [-80.74541534533368, 35.18758221442001], [-80.74628367860348, 35.18660936366408], [-80.74944772637592, 35.185936382627936], [-80.74960698512865, 35.18514246888889], [-80.74900415290747, 35.18479215731499], [-80.75003165901474, 35.18302536158608], [-80.75298323754348, 35.18341083831957], [-80.75500403910205, 35.1834031407568], [-80.75648464051663, 35.18441115874704], [-80.75777700314079, 35.18389250592408]]]]}},
```
hellkite500, line 6: #261 fixes this as well.

hellkite500 (approved):

PR #285, opened 2021-08-18
Support for Optional Subdivided Hydrofabric Usage

robertbartel:
Adding support to optionally load only a subdivided portion of the hydrofabric in each MPI rank, where this portion corresponds to the rank's partition.

- Command line arg for indicating whether to run driver processes in "subdivided hydrofabric" mode, as opposed to default of just loading the entire hydrofabric in each process.
- Support for determining and loading different, rank specific subdivided hydrofabric files
- Support for generating subdivided hydrofabric files on the fly through use of external DMOD Python tool.
- Updates and additions to related documentation.

**CAVEAT:**

This change introduces an optional external dependency.  That dependency is in the sister DMOD project, but is still under review: https://github.com/NOAA-OWP/DMOD/pull/92.  I suggest this Nextgen change remain in review status until the DMOD change is approved and merged.


doc/DEPENDENCIES.md
```diff
@@ -163,4 +165,20 @@ If the above `update` command does not check out the expected version, this can
...
+
+### Setup
+
+This must be installed in the active Python environment at the time the `ngen` executable is called.  This is usually best done via a virtual environment, especially if one is already set up as described for the [Numpy](#numpy) dependency.  However, all that matters is that the Python environment available to the ngen driver has this package installed.
+
+If installing to a virtual env as suggested above, the installation steps are similar to what is described in the [DMOD project document for the subsetting CLI tool]( https://github.com/NOAA-OWP/DMOD/tree/master/doc/SUBSETTING_CLI_TOOL.md).  The difference is just to create and/or source a virtual environment locate here first, and then change to the DMOD project directory.  All the example commands after the step of sourcing the virtual environment can then be followed as described on that page.
+
+### Version Requirements
+
+Version `0.3.0` or greater must be installed, which is available after approval of DMOD PR #92.
```
hellkite500, line 184: This PR has been merged, I would remove the second part of this sentence alltogether.

doc/DISTRIBUTED_PROCESSING.md
```diff
@@ -0,0 +1,91 @@
...
+* all processes completely load the entire hydrofabric
+```
+mpirun -n 4 cmake-build/ngen catchment_data.geojson "" nexus_data.geojson "" realization_config.json partition_config.json
+```
+
+### Example 2 - Subdivided Hydrofabric
+* the CMake build directory is named `cmake-build/`
+* eight MPI processes are started
+* the catchment and nexus hydrofabric files, realization config, and partition config have intuitive names and are located in the current working directory
+* all processes completely load the full hydrofabric
```
hellkite500, line 76: This statement doesn't seem accurate given the `--subdivided-hydrofabric` flag is used in this example.

include/core/PyHydrofabricSubsetter.hpp
```diff
@@ -0,0 +1,129 @@
+#ifndef NGEN_PYHYDROFABRICSUBSETTER_HPP
+#define NGEN_PYHYDROFABRICSUBSETTER_HPP
+
+#ifdef ACTIVATE_PYTHON
+#include <pybind11/embed.h>
+namespace py = pybind11;
+#endif // ACTIVATE_PYTHON
```
hellkite500, line 7: Any reason not to wrap this entire header in the `#ifdef ACTIVATE_PYTHON` macro?  The way this class is now, you can create a valid `PyHydroFabricSubsetter` if `ACTIVATE_PYTHON` is False, but that object just has a few path strings in it and doesn't do anything...

src/NGen.cpp
```diff
@@ -53,12 +68,166 @@ pdm03_struct get_et_params() {
...
+ * ranks (except ``0``) send their local status to rank ``0``.  Rank ``0`` then applies a Boolean AND to the statuses
+ * (including its own) to produce a global status value.  The global status is then broadcast back to the other ranks.
+ * Finally, the value indicated by this global status is returned.
+ *
+ * @param status The initial individual state for the current MPI rank.
+ * @param taskDesc A description of the related task, used by rank 0 to print a message when included for any rank that
+ *                 is not in the success/ready state.
+ * @return Whether all ranks coordinating status had a success/ready status value.
+ */
+bool mpiSyncStatusAnd(bool status, const std::string &taskDesc) {
```
hellkite500, line 90: I would like to this function in a different compilation unit and included/used in the driver if needed.

src/NGen.cpp
```diff
@@ -53,12 +68,166 @@ pdm03_struct get_et_params() {
...
+/**
+ * Attempt to subdivide the passed hydrofabric files into a series of per-partition files.
+ *
+ * This function assumes that, when it is called, the intent is for it to produce a freshly subdivided hydrofabric and
+ * associated files.  As a result, if there are any other subdivided hydrofabric files present having the same names
+ * as the files the function will write, then those preexisting files are considered stale and overwritten.
+ *
+ * @return Whether subdividing was successful.
+ */
+bool subdivide_hydrofabric() {
```
hellkite500, line 182: Wondering if a `parallel_utils` would be a good place for a few of these function definitions.

hellkite500 (changes requested):
Lets find a better place for the driver functions that are MPI/parallel specific.

include/core/PyHydrofabricSubsetter.hpp
robertbartel: I think the answer to the first question is "yes." 

In a situation like this, where the type needs a file to function, I think throwing an exception in the constructor is the appropriate way to handle if the file is not available.  We could defer this kind of check until after the constructor I suppose, but I'm not sure that's really any better.

Regardless, sometimes the possibility of not being able to obtain a useable instance of a type is unavoidable, in particular when files are involved, but I expect in other situations also.

Given that, should using code be concerned with determining whether it _will_ get back a usable instance, when either way it is going to need to be concerned with whether it _did_ get back one?

A caveat and counter argument, however:  even assuming that approach is better in general, it is better in this case?  Right now, our using code is already necessarily pretty concerned with both whether Python is active and whether those files are available.

robertbartel: I went ahead and went with the simpler approach for now.  There are several things in the driver and driver utilities that need to be improved, so deciding on this probably should wait until then.

src/NGen.cpp
robertbartel: Moved to dedicated file.

src/NGen.cpp
robertbartel: Moved to dedicated file.

doc/DEPENDENCIES.md
```diff
@@ -8,8 +8,10 @@
 | [C/C++ Compiler](#c-and-c-compiler) | external | see below |  |
 | [CMake](#cmake) | external | \>= `3.12` | |
 | [Boost (Headers Only)](#boost-headers-only) | external | `1.72.0` | headers only library |
-| [Python 3 Libraries](#python-3-libraries) | external | \> `3.6.12` | Can be [excluded](#overriding-python-dependency). Requires ``numpy`` package |
+| [MPI](https://www.mpi-forum.org) | external | No current implementation or version requirements | Required for [multi-process distributed execution](DISTRIBUTED_PROCESSING.md) |
+| [Python 3 Libraries](#python-3-libraries) | external | \> `3.6.8` | Can be [excluded](#overriding-python-dependency). Requires ``numpy`` package |
```
hellkite500, line 12: I was recently reminded that python [3.6 EOL](https://endoflife.date/python) is in December.  We might consider making 3.7 our minimum.

hellkite500 (approved):

PR #291, opened 2021-09-09
Updating BMI module submodule configuration and usage

robertbartel:
Updating submodule config, using individual (public) repos for CFE and Topmodel, along with shared lib build helper files.  Also removing existing alt-modular submodule and updating path references to previous shared library file locations to reflect the changes.

Also, fixes an unrelated compilation bug in the main source tree that was somehow introduced (despite previous PRs passing checks).


donaldwj (approved):
I see no problems

PR #292, opened 2021-09-17
Noah-mp BMI shared library build file with netcdf dependency link

hellkite500:
This PR simply cleans up the extern noah-mp-modular BMI shared library build.  Specifically, it provides CMake cache variables for each key piece of netcdf info that is needed.  It tries to set them from automatically looking, but this doesn't always work, and even if it does, it may not be correct.  So you can run `cmake`, then edit the cache variables manually (using `ccmake` or `cmake gui` or directly editing `CMakeCahche.txt`.  The following variables control the liked netcdf libs

`netCDF_C_LIB` The path to the C shared object library file, `libnetcdf.so`
`netCDF_FORTRAN_LIB` The path to the Fortran shared object library file, `libnetcdff.so`
`netCDF_INCLUDE_DIR` NetCDF Include Directory.
`netCDF_MOD_PATH` Directory containing the netcdf.mod fortran module

### DIFF
--- a/cmake/FindNetCDF.cmake
+++ b/cmake/FindNetCDF.cmake
@@ -1,0 +1,131 @@
+#[==[
+Provides the following variables:
+
+  * `NetCDF_FOUND`: Whether NetCDF was found or not.
+  * `NetCDF_INCLUDE_DIRS`: Include directories necessary to use NetCDF.
+  * `NetCDF_LIBRARIES`: Libraries necessary to use NetCDF.
+  * `NetCDF_VERSION`: The version of NetCDF found.
+  * `NetCDF::NetCDF`: A target to use with `target_link_libraries`.
+  * `NetCDF_HAS_PARALLEL`: Whether or not NetCDF was found with parallel IO support.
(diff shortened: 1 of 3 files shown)


hellkite500:
Need to add notes to the `extern/noah-mp-modular/README.md` about these:
- `netCDF_C_LIB The path` to the C shared object library file, libnetcdf.so
- `netCDF_FORTRAN_LIB` The path to the Fortran shared object library file, libnetcdff.so
- `netCDF_INCLUDE_DIR` NetCDF Include Directory.
- `netCDF_MOD_PATH` Directory containing the netcdf.mod fortran module

As well as examples of setting these on the command line, i.e. `cmake -DnetCDF_MOD_PATH=/usr/lib64/gfortran/modules/ .` or using `ccmake`.

mattw-nws (approved):

mattw-nws (approved):

donaldwj (commented):
I see no problems

donaldwj (approved):

PR #296, opened 2021-09-23
fix cmake python build flag so it isn't always true

### LINKED ISSUE #294: Make sure fresh build system will work without Python installed
It should be possible to initialize a CMake build system just fine without Python and Numpy installed, as long as the appropriate CMake flags are set to deactivate Python integration.   However, at present, the initialization fails, regardless of the value of the Python build setting, unless Python and Numpy are indeed available.

hellkite500:
Fix bug which makes `NGEN_ACTIVATE_PYTHON` always `true` in the CMakeLists.txt, which inadvertently requires all the python dependencies to build.  Default should be false, or it should be documented that python is required unless explicitly disabled.  Closes #294.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,14 +92,14 @@ endif()
 
 #### Python stuff
 # First, define NGEN_ACTIVATE_PYTHON, if not already defined
-#   Use NGEN_ACTIVATE_PYTHON environment variable if available, or default to true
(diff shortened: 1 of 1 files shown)


hellkite500:
After some disscussion, it is clear that this needs a more sophisticated fix.  Python support should be optional, but default.  A cleaner implementation is required such that if python cannot be properly discovered and configured, then the feature is automatically disabled an the build continues with a warning.

PR #306, opened 2021-10-15
Fix hydrofabric subdividing

robertbartel:
Improvements to subdividing hydrofabric logic.  Most importantly, ensuring that, when it is necessary to perform subdividing, only one MPI process/rank loads the entire hydrofabric (otherwise, there may not be enough memory, which was the point of working with subdivided hydrofabrics).

Also improving/fixing build and execution issues with the subdividing logic when Python functionality is not set to be active.


CMakeLists.txt
```diff
@@ -106,31 +106,65 @@ endif()
 # If Python set to be active, run steps to make sure it is included properly, and set compiler definition to true
 if (${NGEN_ACTIVATE_PYTHON})
     add_compile_definitions(ACTIVATE_PYTHON=true)
-    # If there is a virtual environment directory at ./.venv, then look for numpy there
-    if(EXISTS ${PROJECT_SOURCE_DIR}/.venv)
-        file(GLOB_RECURSE NP_COMMON_HEADER ${PROJECT_SOURCE_DIR}/.venv/*npy_common.h)
+    # If there is a virtual environment directory then look for several things there
+    if(EXISTS ${PROJECT_SOURCE_DIR}/.venv OR EXISTS ${PROJECT_SOURCE_DIR}/venv)
```
hellkite500, line 110: It looks like this snuck in with #307 as well.

include/utilities/parallel_utils.h
```diff
@@ -127,6 +131,297 @@ namespace parallel {
...
+        }
+
+        // Get this rank's hostname
+        gethostname(hostnames[mpi_rank], hostnameBufferSize);
+        // Set the one for this rank
+        actualHostnameCStrLength[mpi_rank] = std::strlen(hostnames[mpi_rank]);
+
+        if (mpi_rank != 0) {
+            // First, send how long the string is
+            MPI_Send(&actualHostnameCStrLength[mpi_rank], 1, MPI_INT, 0, 0, MPI_COMM_WORLD);
```
hellkite500, line 167: This might be a good place for [MPI_Gather](https://www.rookiehpc.com/mpi/docs/mpi_gather.php)

include/utilities/parallel_utils.h
```diff
@@ -127,6 +131,297 @@ namespace parallel {
...
+            // Then get back a code, which will be -1 if bad and need to exit and otherwise good
+            MPI_Recv(&code, 1, MPI_INT, destRank, 0, MPI_COMM_WORLD, MPI_STATUS_IGNORE);
+            if (code == -1) {
+                // TODO: output error message
+                fclose(file);
+                return false;
+            }
+        }
+        // Once there is no more file to read and send, send an empty string
+        MPI_Send("", bufSize, MPI_CHAR, destRank, 0, MPI_COMM_WORLD);
```
hellkite500, line 250: What happens when the text file has a blank line in it?  Will the receiver continue as expected to receive the next line?  It doesn't look like it, the receiver will close the writing file and break the receiving loop.  But the sending is going continue to the next line (since it got a code of 0...the str length of the offending send), and it will hang waiting for the receiver to receive.   Or does `fgets` read newlines and consider that a line of length 1?  I guess the `while` will catch that case, since `fgets` doesn't read a single new line line, the pointer will be null, but then you might miss some of the file due to early stop.  So this function can only `send_text_file_up_to_first_blank_line` I think.

This might be a good place to use an MPI_TAG to indicate the difference between a data message and protocol message:

```c++
        while (fgets(buf, bufSize, file) != NULL) {
            // First send this batch
            MPI_Send(buf, bufSize, MPI_CHAR, destRank, DATA_TAG, MPI_COMM_WORLD);
            // Then get back a code, which will be -1 if bad and need to exit and otherwise good
            MPI_Recv(&code, 1, MPI_INT, destRank, DATA_TAG, MPI_COMM_WORLD, MPI_STATUS_IGNORE);
            if (code == -1) {
                // TODO: output error message
                fclose(file);
                return false;
            }
        }
        // Once there is no more file to read and send, send an empty string
        MPI_Send("", bufSize, MPI_CHAR, destRank, PROTOCOL_TAG, MPI_COMM_WORLD);
```

Then in receive, you can match the `PROTOCOL_TAG` in the appropriate protocol messages.

PR #307, opened 2021-10-15
Fix/Finalize Multi Module BMI

robertbartel:
Fixing a few issues and adding final required implementation details for the Bmi_Multi_Formulation.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -15,7 +15,7 @@
 	branch = master
 [submodule "extern/topmodel/topmodel"]
 	path = extern/topmodel/topmodel
-	url = git@github.com:NOAA-OWP/topmodel.git
+	url = https://github.com/NOAA-OWP/topmodel.git
 	branch = master
 [submodule "extern/noah-mp-modular/noah-mp-modular"]
 	path = extern/noah-mp-modular/noah-mp-modular
--- a/include/realizations/catchment/Bmi_Multi_Formulation.hpp
+++ b/include/realizations/catchment/Bmi_Multi_Formulation.hpp
@@ -5,7 +5,6 @@
 #include <vector>
 #include "Bmi_Formulation.hpp"
 #include "Bmi_Module_Formulation.hpp"
-#include "Bmi_C_Formulation.hpp"
 #include "bmi.hpp"
 #include "ForcingProvider.hpp"
 
@@ -17,12 +16,12 @@ namespace realization {
 
     /**
      * Abstraction of a formulation with multiple backing model object that implements the BMI.
-     *
-     * @tparam M The type for the backing BMI model object.
      */
     class Bmi_Multi_Formulation : public Bmi_Formulation {
 
     public:
+
+        typedef std::shared_ptr<Bmi_Module_Formulation<bmi::Bmi>> nested_module_ptr;
 
         /**
          * Minimal constructor for objects initialize using the Formulation_Manager and subsequent calls to
@@ -36,6 +35,18 @@ namespace realization {
                 : Bmi_Formulation(move(id), move(forcing_config), output_stream) { };
 
         virtual ~Bmi_Multi_Formulation() {};
+
+        /**
+         * Perform ET (or PET) calculations.
+         *
+         * This function simply defers to the analogous function of the first of this instance's nested modules.
+         *
+         * @return The ET calculation from the first of this instance's BMI nested modules.
+         */
+        double calc_et() override {
+            // TODO: add better way of determining which module to use for this
+            return modules[0]->calc_et();
+        }
 
         void create_formulation(boost::property_tree::ptree &config, geojson::PropertyMap *global = nullptr) override {
             geojson::PropertyMap options = this->interpret_parameters(config, global);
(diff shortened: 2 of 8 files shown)


src/realizations/catchment/Bmi_Multi_Formulation.cpp
```diff
@@ -23,11 +26,24 @@ void Bmi_Multi_Formulation::create_multi_formulation(geojson::PropertyMap proper
...
             module = std::dynamic_pointer_cast<Bmi_Module_Formulation<bmi::Bmi>>(std::make_shared<Bmi_C_Formulation>(identifier, forcing, output));
         }
-        else {
+        #endif // NGEN_BMI_C_LIB_ACTIVE
+        #ifdef NGEN_BMI_FORTRAN_ACTIVE
+        if (type_name == "bmi_fortran") {
+            module = std::dynamic_pointer_cast<Bmi_Module_Formulation<bmi::Bmi>>(std::make_shared<Bmi_Fortran_Formulation>(identifier, forcing, output));
+        }
+        #endif // NGEN_BMI_FORTRAN_ACTIVE
+        #ifdef ACTIVATE_PYTHON
```
donaldwj, line 40: Is python intended to be disabled currently?

donaldwj (changes requested):
There are a few issues, noted in the other comment I would like to be addressed.

.gitmodules
```diff
@@ -15,7 +15,7 @@
 	branch = master
 [submodule "extern/topmodel/topmodel"]
 	path = extern/topmodel/topmodel
-	url = git@github.com:NOAA-OWP/topmodel.git
+	url = https://github.com/NOAA-OWP/topmodel.git
```
hellkite500, line 18: Eventually we will get all these fixed, I'm pretty sure I have this in #304 as well.

src/realizations/catchment/Bmi_Multi_Formulation.cpp
```diff
@@ -109,17 +148,43 @@ void Bmi_Multi_Formulation::create_multi_formulation(geojson::PropertyMap proper
...
+ * alias was included in the configuration.
+ *
+ * This is part of the @ref ForcingProvider interface.  This interface must be implemented for items of this
+ * type to be usable as "forcing" providers for situations when some other object needs to receive as an input
+ * (i.e., one of its forcings) a data property output from this object.
+ *
+ * @return The collection of forcing output property names this instance can provide.
+ * @see ForcingProvider
+ */
+const vector<std::string> &Bmi_Multi_Formulation::get_available_forcing_outputs() {
```
hellkite500, line 172: Could they also provide forcing from the "ForcingProvider", e.g. AORC variables at the top of the provider "stack"

src/realizations/catchment/Bmi_Multi_Formulation.cpp
```diff
@@ -205,13 +269,69 @@ const string &Bmi_Multi_Formulation::get_config_mapped_variable_name(const strin
...
 
+// TODO: remove from this level - it belongs (perhaps) as part of the ForcingProvider interface, but is general to it
 const string &Bmi_Multi_Formulation::get_forcing_file_path() const {
-    // TODO: add something that ensures these are set to same path for all modules (allowing some to be unset also)
     return modules[0]->get_forcing_file_path();
 }
 
-const vector<std::string> &Bmi_Multi_Formulation::get_output_variable_names() const {
-    return modules.back()->get_output_variable_names();
+string Bmi_Multi_Formulation::get_output_line_for_timestep(int timestep, std::string delimiter) {
```
hellkite500, line 277: This is a bit of a complicated mess.  Definitely worth coming back to and refactoring this behavior.

include/realizations/catchment/Bmi_Multi_Formulation.hpp
```diff
@@ -270,6 +291,29 @@ namespace realization {
...
+         * This is part of the @ref ForcingProvider interface.  This interface must be implemented for items of this
+         * type to be usable as "forcing" providers for situations when some other object needs to receive as an input
+         * (i.e., one of its forcings) a data property output from this object.
+         *
+         * @param name The name of the forcing property for which the current value is desired.
+         * @return Whether the property's value is an aggregate sum.
+         */
+        bool is_property_sum_over_time_step(const string &name) override {
+            // TODO: verify with some kind of proof that "always true" is appropriate
+            return true;
```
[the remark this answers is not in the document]
robertbartel: Addressed in `73276d97aca4436f6a9fc5aaa0b23b62d06d31f5`.

hellkite500 (approved):

src/realizations/catchment/Bmi_Multi_Formulation.cpp
robertbartel: @hellkite500, @mattw-nws, I think I see where you both are going with this now ... the copying is excessive.  I still don't feel great about that kind of transitive providing, but I'm sure there is a better alternative than what it is doing now.

src/realizations/catchment/Bmi_Multi_Formulation.cpp
hellkite500: That is why the original `Forcing` concept existed all the way up the chain...I think we don't want as much of a unified interface for data access for any eventual representation that may do data transformations (models...).

hellkite500: No, I cannot.  My hasty fingers lied to the keyboard.  Edited the previous comment to **maybe** be more clear...

PR #308, opened 2021-10-15
Docker debugging

robertbartel:
Adding some debugging utilities, in particular for multi process debugging, that use Docker container and GDB server.


stcui007:
This could potentially be a very innovative approach if it can be seamlessly linked to MaaS. Especially if users can use the debugger in a Browser via GUI. On linux servers and supercomputers/clusters it is more likely people will prefer some native parallel debuggers, such as Allinea DDT or TotalView. Ultimately it depends on how easy this can be deployed and used.

PR #323, opened 2021-11-01
Basic automatic unit conversion

mattw-nws:
Basic unit conversion with UDUNITS2

## Additions

* UnitsHelper class
* Automatic conversion in CsvPerFeatureForcingProvider and Bmi_Multi_Formulation (may be needed elsewhere)
* UDUNITS-2 as a submodule and FindUDUNITS2.cmake to find installed packages

## Removals

-

## Changes

- 

## Testing

Unit tests added, coverage from existing tests passes.

## Screenshots

-

## Notes

-

## Todos

-

## Checklist

- [x] Linux
- [x] macOS


mattw-nws:
See also #325 with 1624ccbd3fcf4be57de290c00ed9a9630c3387ec

src/core/mediator/UnitsHelper.cpp
```diff
@@ -0,0 +1,38 @@
...
+ut_system* UnitsHelper::unit_system;
+std::once_flag UnitsHelper::unit_system_inited;
+
+double UnitsHelper::get_converted_value(const std::string &in_units, double value, const std::string &out_units)
+{
+    if(in_units == out_units){
+        return value; // Early-out optimization
+    }
+    if(in_units == "" || out_units == ""){
+        return value; //TODO: Warn? Make behavior configurable?
```
robertbartel, line 12: I don't think we want to do this.  If nothing else, this isn't going to throw a _runtime_error_ that will trigger the warning message.

Further, aside from being another place to possibly forget to go back to if we add configurability, what scenario is there that we need to support something not providing ~~valid~~ any units for itself?  Sure, it makes sense to just do value pass-thru for now if there is a hangup in _**the framework**_ recognizing how to convert between models' units.  But I'm pretty sure this case would only occur when a model (or a forcing data provider source) didn't provide units at all; i.e., a hangup in something external to the framework.  That seems different enough that I think we need to throw an exception that won't be caught and trigger the pass-thru behavior, even at this point.   Not sure whether that means using something other than _runtime_error_ here or changing the places upstream that call this and handle catching that type.

This one is complex, so let me know if you want to discuss further.

robertbartel (changes requested):
I have a few things I see issues with.  Also, it doesn't look like the latest changes in `master` are merged into this branch yet.  I'd suggest doing that also to make sure there aren't issues, as I ran into some runtime problems I had to work around trying to test this.

include/forcing/CsvPerFeatureForcingProvider.hpp
```diff
@@ -301,8 +311,9 @@ class CsvPerFeatureForcingProvider : public forcing::ForcingProvider
 
                 if(well_known_fields.count(n) > 0){
                     auto t = well_known_fields[n];
-                    n = std::get<0>(t);
                     u = u == "" ? std::get<1>(t) : u;
+                    available_forcings_units[n] = u; // Allow lookup of units by non-canonical name
```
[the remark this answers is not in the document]
mattw-nws: Yes, I'll improve the variable names, sorry.

Your intuition about the need for an `available_forcings.push_back(n)` is good, but yes this is handled a different way. Since `available_forcings_units` is dimensioned n x 1, I just add a second entry and there is little cost. However, the forcing _data_ is dimensioned n x t, and if I added that with both names, all the data points would have to have a second copy, which seemed memory inefficient. So, the alias is handled by lookup at retrieval time instead within `get_value_for_param_name`.

src/core/mediator/UnitsHelper.cpp
mattw-nws: Good question. I think I hit this when testing and one of the C++ models did this... or the legacy `Forcing` object... really I think this block existed primarily as an optimization: if it isn't there, then the unit strings just get passed to `ut_parse` and possibly `ut_get_converter` and somewhere along the line it will fail to convert... and I don't know what the execution costs are on those functions and they will be run a _lot_. Basically we probably need to do here whatever we would do in a unit conversion failure (in this case one or the other is _not_ blank, otherwise the previous block would have returned) but leave it here for performance reasons (unless we forcibly abort whenever a unit conversion fails).

I think I had another scenario in mind where a researcher/tinkerer is doing something simple but hasn't messed with implementing the units BMI functions but is using an existing model that does, but he knows that the units match... maybe we shouldn't cater to that.

mattw-nws: What may make sense is to have a method within `UnitsHelper` that is always called on unit conversion failure... that can be the choke point for a configurable behavior and elimintates the "forgot to do it here" problem.

mattw-nws: Huh... So, after changing this to `throw`, all the tests still pass. I think I thought the legacy `Forcing` object would cause this case, but I guess it's not an issue because the two paths never cross... strange to me that this is the case though after I did the testing enhancements in #328 ... I suppose we can leave it like this, unless the tinkerer scenario above seems valid.

include/forcing/CsvPerFeatureForcingProvider.hpp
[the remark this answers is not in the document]
robertbartel: I don't think I follow why you'd need to add a duplicate vector to `forcing_vectors`.   Can't we just  check in `get_value_for_param_name` whether the `name` is a key in `forcing_vectors`, and if not convert it to the "well known" equivalent?

Honestly, I bet we could also adjust `forcing_vectors` to use pointers to the vectors as the map values instead of the vectors directly, and use that to avoid the duplication.

Regardless, I am uneasy about `available_forcings` and `get_available_forcing_outputs` not actually returning the complete collection of valid names for which output can be requested.  I guess that's what I get for not documenting the function, but it seems invalid for something not returned by that to be usable in `get_value` call.  These also feel like assumptions that could easily be erroneously relied upon somewhere else, if they aren't already.

LICENSE
```diff
@@ -1,7 +0,0 @@
-“Software code created by U.S. Government employees is not subject to copyright
```
hellkite500, line 1: Revert this delete, and remove `COPYRIGHT`

docker/CENTOS_NGEN_RUN_incomplete.dockerfile
```diff
@@ -0,0 +1,38 @@
+FROM centos:8.4.2105 as builder
```
hellkite500, line 1: Remove this file if it is incomplete.

src/core/mediator/UnitsHelper.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+double UnitsHelper::get_converted_value(const std::string &in_units, double value, const std::string &out_units)
+{
+    if(in_units == out_units){
+        return value; // Early-out optimization
+    }
+    if(in_units == "" || out_units == ""){
+        throw std::runtime_error("Unable to process empty units value for pairing \"" + in_units + "\" \"" + out_units + "\"");
+    }
+    std::call_once(unit_system_inited, init_unit_system);
```
hellkite500, line 14: Why doesn't this init happen in the constructor?

mattw-nws: That's a static method... at present there is no constructor (is there such a thing as a static constructor in C++?)

CMakeLists.txt
```diff
@@ -39,16 +39,43 @@ endif()
...
+    if(GIT_SUBMODULE)
+        message(STATUS "Updating submodule ${PATH}")
+        execute_process(COMMAND ${GIT_EXECUTABLE} submodule update --init -- ${PATH}
+            RESULT_VARIABLE GIT_SUBMOD_RESULT
+            )
+        if(NOT GIT_SUBMOD_RESULT EQUAL "0")
+            message(FATAL_ERROR "git submodule update --init for ${PATH} failed with ${GIT_SUBMOD_RESULT}")
+        endif()
+    else()
+        message(FATAL_ERROR "Git not detected: cannot obtain Udunits submodule (but how is this even working?)")
```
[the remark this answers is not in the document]
mattw-nws: done

include/forcing/CsvPerFeatureForcingProvider.hpp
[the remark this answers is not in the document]
mattw-nws: >Honestly, I bet we could also adjust forcing_vectors to use pointers to the vectors as the map values instead of the vectors directly, and use that to avoid the duplication.
That was what I was thinking of when I wrote my "Edit" at the end of my above comment.
>Regardless, I am uneasy about available_forcings and get_available_forcing_outputs not actually returning the complete collection of valid names for which output can be requested.
That's tricky to me, in this case...you arguably don't have 14 variables available, you only have 7... at the same time, retrieving a name not in the list "works"... yes, it's somewhat ambiguous.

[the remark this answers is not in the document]
mattw-nws: I can take a crack at refactoring that to have `std::shared_ptr<std::vector>`s in the map, but I will definitely want some review, because my "modern C++"-pointer-fu is still weak.

src/core/mediator/UnitsHelper.cpp
hellkite500: Hmm...I guess the way it is used you may not need one...but then you may as well just use a namespace (and a nested anonymous namespace) for `private` components.  Unless there is a need for dynamic runtime state, the `class` seems to be the wrong construct here?

include/forcing/CsvPerFeatureForcingProvider.hpp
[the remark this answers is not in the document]
mattw-nws: >I don't think I follow why you'd need to add a duplicate vector to forcing_vectors. Can't we just check in get_value_for_param_name whether the name is a key in forcing_vectors, and if not convert it to the "well known" equivalent?

Doh... looking at this again I see why you were confused, and I was remembering wrong. `available_forcings` is only the list of names... the values are actually stored in `forcing_vectors`...so yes, we can totally add the name to `available_forcings` as well so that the output of `get_available_forcing_outputs` returns all available names.

src/core/mediator/UnitsHelper.cpp
mattw-nws: Dunno. Since this is using a C library in a C++ context it's hard to know how best to adapt to OO... in future we likely may want to cache converters (in addition to the `unit_system`), so there may be more "state" later....and putting them in global variables (even in a namespace) felt old-school (even if they are functionally indistinguishable). Not sure what common/best practice is here. But yes I think there may be more state later. And thread-safety was a concern, I'm not sure if a static class will make that easier or harder than namespaced globals.

src/core/mediator/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+
+# TODO: consider setting a minimum or required version
+find_package(Boost)
+
+target_link_libraries(core_mediator PUBLIC
+        Boost::boost                # Headers-only Boost
+        )
+
+if(MPI_ACTIVE)
+    add_compile_definitions(NGEN_MPI_ACTIVE)
```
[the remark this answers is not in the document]
mattw-nws: Yep this was a copypasta from https://github.com/NOAA-OWP/ngen/blame/master/src/core/hydrolocation/CMakeLists.txt#L23-L24 😏 should I remove that one, too?

src/core/mediator/UnitsHelper.cpp
mattw-nws: Fixed (with test) in 7020399.

src/core/mediator/CMakeLists.txt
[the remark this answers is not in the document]
mattw-nws: Fixed in 7020399

include/forcing/CsvPerFeatureForcingProvider.hpp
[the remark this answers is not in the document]
mattw-nws: Fixed in 7020399

include/forcing/CsvPerFeatureForcingProvider.hpp
```diff
@@ -143,7 +145,15 @@ class CsvPerFeatureForcingProvider : public forcing::ForcingProvider
...
                 value += involved_time_step_values[i] * ((double)involved_time_step_seconds[i] / (double)duration_s);
         }
-        return value;
+
+        // Convert units
+        try {
+            return UnitsHelper::get_converted_value(available_forcings_units[output_name], value, output_units);
+        }
+        catch (const std::runtime_error& e){
+            std::cerr<<"Unit conversion error: "<<std::endl<<e.what()<<std::endl<<"Returning unconverted value!"<<std::endl;
```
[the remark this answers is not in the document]
mattw-nws: Yes, a custom exception subclass might make sense here also. But concur, later...it will be interesting to see how often this occurs once we start running it.

robertbartel (approved):

PR #325, opened 2021-11-02
Add UDUNITS_ACTIVE to build runners

mattw-nws:

## Additions

-

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

1.

- [ ] Linux

### DIFF
--- a/.github/workflows/test_and_validate.yml
+++ b/.github/workflows/test_and_validate.yml
@@ -47,7 +47,7 @@ jobs:
       run: |
         export BOOST_ROOT="$(pwd)/boost_1_72_0"
         [ ! -d "$BOOST_ROOT" ] && echo "Error: no Boost root found at $BOOST_ROOT" && exit 1
-        cmake -B cmake_build -DBMI_C_LIB_ACTIVE:BOOL=OFF -DNGEN_ACTIVATE_PYTHON:BOOL=OFF -S .
+        cmake -B cmake_build -DBMI_C_LIB_ACTIVE:BOOL=OFF -DNGEN_ACTIVATE_PYTHON:BOOL=OFF -DUDUNITS_ACTIVE:BOOL=ON -S .
     
     - name: build_tests
       run: cmake --build cmake_build --target test_unit -- -j ${{ env.LINUX_NUM_PROC_CORES }}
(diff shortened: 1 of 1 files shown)


mattw-nws:
See also 1624ccbd3fcf4be57de290c00ed9a9630c3387ec

robertbartel (commented):
@mattw-nws I have a few things I'm curious about. 

First, this seems to be applicable to something that isn't yet committed to `master` (I see now that's probably #323).  Why separate this from the related set of changes?

Second, I think this adds `UDUNITS_ACTIVE` to the generated CMake build for every runner job, and I would have expected UDUnits to be a required dependency.   So why does there needs to be a CMake variable; is switching off `UDUNITS_ACTIVE`  a valid use case?  If so, is/will there be a separate suite of tests for those scenarios?

mattw-nws:
[First:] Yep, the builders won't run to pass #323 (draft) until this is in place, but once it is, they _should_ pass, which breaks the paradox.

[Second:] Yes, I thought it possibly worth making it a configurable dependency for at least two scenarios, A: we might have another option in the future or a more basic one built in without a dependency, or B: allow builld without unit conversion support which would just passthrough values, which might be valid and simpler for some research scenarios. We could alternatively make it configurable but enabled by default like Python is currently, but even Python has the switch specified in the runners... I thought it better to be explicit.

mattw-nws:
(addendum to [First] above: I'm pretty sure the workflows will only run the configuration in `master`, not what's in the PR branch...hence why to separate it.)

robertbartel:
> (addendum to [First] above: I'm pretty sure the workflows will only run the configuration in `master`, not what's in the PR branch...hence why to separate it.)

@mattw-nws Ah, I see, though [I don't think this is correct](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request).   See item number 3 toward the bottom of [this section](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#example-using-multiple-events-with-activity-types-or-configuration) also.  It runs on the branch triggering the run, and the workflow config is part of the branch.

An experiment testing the behavior for this case might not be a bad idea.

mattw-nws:
It appears as though it doesn't work... I just added 282aca5 to #323 and the builds still fail with a missing udunits2.h file.

mattw-nws:
Correction (late), it does work, the failure was for a different reason after that.

PR #326, opened 2021-11-02
Hydrofabric integration

hellkite500:
Latest updates to hydrofabric require some minor changes to work appropriately with NGen.

## Additions

- Error checking conditions in partition generator

## Changes

- Add support for *terminal nexus* using prefix `tnx`

## Todos

- Update network tests to test for `tnx` nexus types

## Checklist

- [x] Linux
- [x] MacOS

### DIFF
--- a/src/NGen.cpp
+++ b/src/NGen.cpp
@@ -265,7 +265,12 @@ int main(int argc, char *argv[]) {
                              r_c->get_output_line_for_timestep(output_time_index)+"\n";
         r_c->write_output(output);
         //TODO put this somewhere else.  For now, just trying to ensure we get m^3/s into nexus output
-        response *= (catchment_collection->get_feature(id)->get_property("areasqkm").as_real_number() * 1000000);
+        try{
+          response *= (catchment_collection->get_feature(id)->get_property("areasqkm").as_real_number() * 1000000);
+        }catch(std::invalid_argument &e)
+        {
+          response *= (catchment_collection->get_feature(id)->get_property("area_sqkm").as_real_number() * 1000000);
+        }
         //update the nexus with this flow
         for(auto& nexus : features.destination_nexuses(id)) {
           //TODO in a DENDRIDIC network, only one destination nexus per catchment
--- a/src/partitionGenerator.cpp
+++ b/src/partitionGenerator.cpp
@@ -166,6 +166,10 @@ void generate_partitions(network::Network& network, const int& num_partitions, c
                 nexus_set.emplace(downstream);
                 //nexus_list.push_back(downstream);
             }
+            if(nexus_set.size() == 0){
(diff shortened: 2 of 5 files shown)


src/NGen.cpp
```diff
@@ -265,7 +265,12 @@ int main(int argc, char *argv[]) {
                              r_c->get_output_line_for_timestep(output_time_index)+"\n";
         r_c->write_output(output);
         //TODO put this somewhere else.  For now, just trying to ensure we get m^3/s into nexus output
-        response *= (catchment_collection->get_feature(id)->get_property("areasqkm").as_real_number() * 1000000);
+        try{
+          response *= (catchment_collection->get_feature(id)->get_property("areasqkm").as_real_number() * 1000000);
+        }catch(std::invalid_argument &e)
```
robertbartel, line 270: While I'm ok with this, I am curious how much of an overhead the use of an exception here might introduce.  I've only done a quick investigation, but I did find [this](https://pspdfkit.com/blog/2020/performance-overhead-of-exceptions-in-cpp/), suggesting it could be fairly large (relatively speaking).  

Since this is in a loop of all catchments, within another loop of all time steps, it may be worth considering a runtime check of which key is present, though perhaps as a future improvement.

src/partitionGenerator.cpp
```diff
@@ -479,7 +491,7 @@ int main(int argc, char* argv[])
     global_nexus_collection->link_features_from_property(nullptr, &link_key);
     // make a global network
     Network global_network(global_nexus_collection);
-
+    std::cout<<"HERERERERE\n";
```
robertbartel, line 494: I suspect this `cout` statement was intended to be removed.

robertbartel (changes requested):
One small thing to remove, and one other to consider, though possibly as a future improvement.

src/NGen.cpp
hellkite500: This is really just to support our existing (legacy) example files.  I'm OK with reverting this to a single key once the example files get changed.  This has changed a few times in the past, and I figured for now this was the quickest way to support both.

And which key is present presents the challenge that you need to check the key for each catchment and use the right one in the right place.  It would probably be OK to assume it is used consistently within a file though...

src/partitionGenerator.cpp
hellkite500: Yes, I'll fix that quick.

src/NGen.cpp
robertbartel: Ah, I see.  I do like the idea of a having a variable and then have it be updated in the exception case, assuming that the key is consistent throughout a file.  But the (amortized) overhead I was worried about sounds like it is very low, given the additional context, so I'll leave it up to you.

robertbartel (approved):

PR #328, opened 2021-11-03
Fix forcing provider from config

mattw-nws:

## Additions

Fix flowing forcing `provider` value through the JSON config (global and formulation forcing config).

## Removals

-

## Changes

SimpleLumped formulation now works with both legacy and `CsvPerFeatureForcingProvider` (so that a test can be written without involving a BMI model).

## Testing

1. Added test that runs both global and formulation-specific forcing provider from JSON config

## Screenshots

-

## Notes

-

## Todos

-

## Checklist

1.

- [x] Linux


src/realizations/catchment/Simple_Lumped_Model_Realization.cpp
```diff
@@ -103,7 +103,14 @@ double Simple_Lumped_Model_Realization::calc_et()
 double Simple_Lumped_Model_Realization::get_response(time_step_t t, time_step_t dt)
 {
     //TODO input_et = this->forcing.get_et(t)
-    double precip = this->legacy_forcing.get_next_hourly_precipitation_meters_per_second();
+    double precip;
+    time_t t_unix = this->forcing->get_forcing_output_time_begin("") + (t * 3600);
```
robertbartel, line 107: Shouldn't the argument to `get_forcing_output_time_begin` be `"precip_rate"`?

robertbartel (commented):
I see one thing that looks out of place, though since it isn't breaking tests I won't formally "Request changes" on it.  Otherwise this looks fine.

src/realizations/catchment/Simple_Lumped_Model_Realization.cpp
mattw-nws: Ah. Possibly something other than `""` ... but on reflection I think `""` is the best choice under the circumstances. The idea of this change is to get it to work with three scenarios: legacy `Forcing`, `CsvPerFeatureForcingProvider` with standard columns, `CsvPerFeatureForcingProvider` with non-standard names... without manual mapping at least, `"precip_rate"` isn't a valid name in the third scenario, and the CSDMS standard name isn't valid in the first. ATM, the implementation of `get_forcing_output_time_begin` in both providers ignores the parameter anyway, and all variables are assumed to have the same range... I have been using `""` as a special value meaning "any", but maybe we should use `"*"` and document that.

robertbartel (approved):

PR #329, opened 2021-11-04
Support pattern match and ID replacement in BMI formulation init_config values

robertbartel:
Adds support for catchment ID substitution in the `init_config` value for BMI formulations within the realization config, similarly to how this previously was in place for the global forcing config.

Also includes a few other fixes for bugs found when testing the aforementioned change, which were necessary for the primary change to work.

## Additions

- Add static helper function for performing the pattern matching and substitution in the values of an inflated realization config _PropertyMap_.

## Changes

- Fixes a bug with creation of the nested formulations for a BMI multi formulation.
- Applies new config substitution logic within BMI multi formulation function that initializes nested formulations, replacing `{{id}}` with the feature's actual id value in `init_config_ values.
- Applies new config substitution logic as described above, but in _Formulation_Manager_ when creating "missing" formulations based on the `global` config.
- Adjusts the example configuration for recent forcing provider changes.

### DIFF
--- a/include/realizations/catchment/Catchment_Formulation.hpp
+++ b/include/realizations/catchment/Catchment_Formulation.hpp
@@ -18,6 +18,40 @@ namespace realization {
 
             Catchment_Formulation(std::string id) : Formulation(id){};
 
+        /**
+         * Perform in-place substitution on the given config property item, if the item and the pattern are present.
+         *
+         * Any and all instances of the substring ``pattern`` are replaced by ``replacement``, if ``key`` maps to a
+         * present string-type property value.
+         *
+         * @param properties A reference to the properties config object to be altered.
+         * @param key The key for the configuration property to potentially adjust.
+         * @param pattern The pattern substring to search for that, when present, should be replaced.
+         * @param replacement The replacement substring to potentially insert.
+         */
+        static void config_pattern_substitution(geojson::PropertyMap &properties, const std::string &key,
+                                                const std::string &pattern, const std::string &replacement) {
+            auto it = properties.find(key);
+            // Do nothing and return if either the key isn't found or the associated property isn't a string
+            if (it == properties.end() || it->second.get_type() != geojson::PropertyType::String) {
+                return;
+            }
+
+            std::string value = it->second.as_string();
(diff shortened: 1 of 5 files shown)


include/realizations/catchment/Catchment_Formulation.hpp
```diff
@@ -18,6 +18,40 @@ namespace realization {
...
+            size_t id_index = value.find(pattern);
+
+            if (id_index != std::string::npos) {
+                do {
+                    value = value.replace(id_index, sizeof(pattern.c_str()) - 2, replacement);
+                    id_index = value.find(pattern);
+                } while (id_index != std::string::npos);
+
+                properties.erase(key);
+                properties.template insert(
```
hellkite500, line 50: `properties.template insert(...)` Is this even valid syntax?

hellkite500: Hmm....this is interesting for sure.  The syntax is not very common, but is indeed valid, and (sometimes) required.  [Here](https://stackoverflow.com/questions/610245/where-and-why-do-i-have-to-put-the-template-and-typename-keywords) is a good explanation with references to this syntax in the C++ standard.

I'm curious why you don't simply use:
```c++
properties.erase(key);
properties.emplace(key, geojson::JSONProperty(key, value));
```
I tried this out locally and didn't get any compile issues at least...

hellkite500: I will note that `emplace` is __typically__ more efficient since it only has to essentially call the copy constructor of JSONProperty here, instead of having to build and copy the pair and then extract and copy the JSONProperty...It also helps reduce the template mess...

robertbartel: @hellkite500, I switch to your suggestion.  My impression was that the syntax of `insert` I was using would move the created pair, but I'm not totally certain.  Your way is at least as efficient, if not more, and a little easier to follow.

hellkite500 (approved):

hellkite500:
Will note that we should probably add some sort of unit test for this in the near future.

PR #335, opened 2021-11-10
Fixes bug introduced in #311

mattw-nws:

## Additions

-

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

1.

- [ ] Linux

### DIFF
--- a/.github/workflows/module_integration.yml
+++ b/.github/workflows/module_integration.yml
@@ -1,0 +1,65 @@
+name: Testing and Validation
+
+# Controls when the action will run.
+on:
+  push:
+    branches: [ master, dev, notreal ]
+  pull_request:
+    branches: [ master, dev, notreal ]
+
+env:
+  # Obtained from https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
+  # TODO: add something later that can check the runners for hyperthreading
+  LINUX_NUM_PROC_CORES: 2
+  MACOS_NUM_PROC_CORES: 3
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # Run general unit tests in linux environment
+  test_surfacebmi_plus_cfe:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+    # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+    - uses: actions/checkout@v2
+      
+    - name: Init Submodules
+      run: git submodule update --init --recursive
+
+    - name: Cache Boost Dependency
+      id: cache-boost-dep
+      uses: actions/cache@v1
+      with:
+        path: boost_1_72_0
+        key: unix-boost-dep
+
+    - name: Get Boost Dependency
+      if: steps.cache-boost-dep.outputs.cache-hit != 'true'
+      run: |
+        curl -L -O https://boostorg.jfrog.io/artifactory/main/release/1.72.0/source/boost_1_72_0.tar.bz2
+        tar xjf boost_1_72_0.tar.bz2
+
+    - name: Build surfacebmi module
+      run: |
+        cmake -B extern/noah-mp-modular/cmake_build cmake_build -S extern/noah-mp-modular
+        cmake --build extern/noah-mp-modular/cmake_build --target surfacebmi -- -j ${{ env.LINUX_NUM_PROC_CORES }}
+
+    - name: Build cfebmi module
+      run: |
+        cmake -B extern/cfe/cmake_build -S extern/cfe/
+        cmake --build extern/cfe/cmake_build/ --target cfebmi -- -j ${{ env.LINUX_NUM_PROC_CORES }}
+
+    - name: cmake_init_build
+      run: |
+        export BOOST_ROOT="$(pwd)/boost_1_72_0"
+        [ ! -d "$BOOST_ROOT" ] && echo "Error: no Boost root found at $BOOST_ROOT" && exit 1
+        cmake -B cmake_build -DBMI_C_LIB_ACTIVE:BOOL=ON -DBMI_FORTRAN_ACTIVE=ON -DNGEN_ACTIVATE_PYTHON:BOOL=OFF -S .
+
+    - name: Build ngen
+      run: cmake --build cmake_build --target ngen -- -j ${{ env.LINUX_NUM_PROC_CORES }}
+      timeout-minutes: 15
+
+    - name: Run surfacebmi plus cfebmi
+      run: ./build/ngen data/catchment_data.geojson "cat-27" data/nexus_data.geojson "" data/example_bmi_multi_realization_config.json
--- a/.github/workflows/module_integration.yml
+++ b/.github/workflows/module_integration.yml
@@ -43,7 +43,6 @@ jobs:
 
     - name: Build surfacebmi module
       run: |
-        gfortran -dumpversion
         cmake -B extern/noah-mp-modular/cmake_build cmake_build -S extern/noah-mp-modular
         cmake --build extern/noah-mp-modular/cmake_build --target surfacebmi -- -j ${{ env.LINUX_NUM_PROC_CORES }}
 
@@ -68,4 +67,4 @@ jobs:
       timeout-minutes: 15
 
     - name: Run surfacebmi plus cfebmi
-      run: ./cmake_build/ngen data/catchment_data.geojson "cat-27" data/nexus_data.geojson "" data/example_bmi_multi_realization_config.json
+      run: ./cmake_build/ngen data/catchment_data.geojson "cat-27" data/nexus_data.geojson "nex-26" data/example_bmi_multi_realization_config.json
--- a/.github/workflows/module_integration.yml
+++ b/.github/workflows/module_integration.yml
@@ -52,6 +52,11 @@ jobs:
         cmake -B extern/cfe/cmake_build -S extern/cfe/
         cmake --build extern/cfe/cmake_build/ --target cfebmi -- -j ${{ env.LINUX_NUM_PROC_CORES }}
 
+    - name: Build iso_c_fortran_bmi module
+      run: |
+        cmake -B extern/iso_c_fortran_bmi/cmake_build -S extern/iso_c_fortran_bmi
+        cmake --build extern/iso_c_fortran_bmi/cmake_build --target iso_c_bmi -- -j ${{ env.LINUX_NUM_PROC_CORES }}
+
     - name: cmake_init_build
       run: |
         export BOOST_ROOT="$(pwd)/boost_1_72_0"
--- a/.github/workflows/module_integration.yml
+++ b/.github/workflows/module_integration.yml
@@ -68,4 +68,4 @@ jobs:
       timeout-minutes: 15
 
     - name: Run surfacebmi plus cfebmi
-      run: ./build/ngen data/catchment_data.geojson "cat-27" data/nexus_data.geojson "" data/example_bmi_multi_realization_config.json
+      run: ./cmake_build/ngen data/catchment_data.geojson "cat-27" data/nexus_data.geojson "" data/example_bmi_multi_realization_config.json
--- a/.github/workflows/module_integration.yml
+++ b/.github/workflows/module_integration.yml
@@ -43,6 +43,7 @@ jobs:
 
     - name: Build surfacebmi module
       run: |
+        gfortran -dumpversion
         cmake -B extern/noah-mp-modular/cmake_build cmake_build -S extern/noah-mp-modular
         cmake --build extern/noah-mp-modular/cmake_build --target surfacebmi -- -j ${{ env.LINUX_NUM_PROC_CORES }}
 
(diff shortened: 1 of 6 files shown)


include/forcing/CsvPerFeatureForcingProvider.hpp
```diff
@@ -195,16 +195,16 @@ class CsvPerFeatureForcingProvider : public forcing::ForcingProvider
 
     // This map may be applicable to well-known netCDF formats as well?
     map<std::string, std::tuple<std::string, std::string>> well_known_fields = {
-        {"precip_rate", { CSDMS_STD_NAME_RAIN_RATE, "mm s^-1" } }, 
-        {"APCP_surface", { CSDMS_STD_NAME_LIQUID_EQ_PRECIP_RATE, "kg m^-2" } }, // Especially this one, is it correct? 
+        {"precip_rate", { CSDMS_STD_NAME_LIQUID_EQ_PRECIP_RATE, "mm s^-1" } }, 
+        {"APCP_surface", { CSDMS_STD_NAME_RAIN_RATE, "kg m^-2" } }, // Especially this one, is it correct?
```
hellkite500, line 199: I strongly recommend changing this macro in aorcforcing.hpp to be reflective of the variable it is mapped to:
```c++
#define CSDMS_STD_NAME_RAIN_RATE "atmosphere_water__rainfall_volume_flux"
```
Some quick dimensional analysis:
- volume flux implies dimensions of `mass^3 * time^-1 *  length^2`
- mass flux implies dimensions of `mass * time^-1`
- RAIN_RATE/precip_rate implies dimensions of `L * time^-1`

So, ACPC_surface, with units `kg * m^-2` is not even a flux.  To be a flux, there needs to be a time dimension.
I think RAIN_RATE and precip_rate both can map to CSDMS_STD_NAME_LIQUID_EQ_PRECIP_RATE (if its dimensions are `L * time^-1`.  ACPC_surface should probably map to a volume flux, but it cannot, without some unit of time, technically do so.  Now we do know that these are per hour, so perhaps we can put that assumption into this, but then I would correct the units of `APCP_surface` to be `kg * hr^-1 * m^-2` to make it a real volume flux.

mattw-nws: Just tell me what CSDMS name to make it... but it seems some model configs and/or realization configs will have to be adjusted to match.

mattw-nws: (we may make that a follow-on issue)

mattw-nws:
We may not want that test runner to become part of the standard suite of tests that run on pull requests? Or maybe we do... but it works on this one.

.github/workflows/module_integration.yml
```diff
@@ -0,0 +1,71 @@
...
+        export BOOST_ROOT="$(pwd)/boost_1_72_0"
+        [ ! -d "$BOOST_ROOT" ] && echo "Error: no Boost root found at $BOOST_ROOT" && exit 1
+        cmake -B cmake_build -DBMI_C_LIB_ACTIVE:BOOL=ON -DBMI_FORTRAN_ACTIVE=ON -DNGEN_ACTIVATE_PYTHON:BOOL=OFF -S .
+
+    - name: Build ngen
+      run: cmake --build cmake_build --target ngen -- -j ${{ env.LINUX_NUM_PROC_CORES }}
+      timeout-minutes: 15
+
+    - name: Run surfacebmi plus cfebmi
+      run: ./cmake_build/ngen data/catchment_data.geojson "cat-27" data/nexus_data.geojson "" data/example_bmi_multi_realization_config.json
```
hellkite500, line 71: If you are only running "cat-27", you should similarly pass the associated nexus "nex-26" to the nexus subset.

include/forcing/CsvPerFeatureForcingProvider.hpp
hellkite500:
```c++
#define CSDMS_STD_NAME_VOLUME_FLUX "atmosphere_water__rainfall_volume_flux"
#define CSDMS_STD_NAME_MASS_FLUX "atmosphere_water__precipitation_mass_flux"
```
and then, assuming we are ok with adjusting the units here to make it a flux???
```c++
"APCP_surface", { CSDMS_STD_NAME_VOLUME_FLUX, "kg h^-1 m^-2" } }
```
Can probably get away with just mapping to the flux with the same units for the time being and opening an issue to resolve later.  
```c++
"APCP_surface", { CSDMS_STD_NAME_VOLUME_FLUX, "kg m^-2" } }
```
Long story short, rain_rate should not be associated with APCP_surface in such a direct fashion.

hellkite500 (changes requested):
Fix the rate -> volume macro then we can merge.

include/forcing/CsvPerFeatureForcingProvider.hpp
mattw-nws: Oh. Didn't understand from the above that you wanted the macro _name_ changed. In either case, not really opposed, but those macro names are used elsewhere, scattered in the legacy `Forcing` class, in `Simple_Lumped_Model_Realization` and in the ET calculations in `Bmi_Module_Formulation`... so while it's a straightforward find and replace, it will bring in a lot more files. Is this in scope? Or should it be also a separate issue?

.github/workflows/module_integration.yml
mattw-nws: Fixed

include/forcing/CsvPerFeatureForcingProvider.hpp
mattw-nws: So I S&R'd the `RAIN_RATE` macro and added the `MASS_FLUX` one (which isn't presently used anywhere)... I think that fulfills the central objection.

As for the units, from the CsvReader's point of view, it's just capturing knowledge about what the units in the file are based on a common column name... so whatever units are there should be whatever the units in the column are... so the real question is "what are the units in the [CSV] column?" (Same question should apply to the CSDMS name!) If they don't match the way they're being used, that's a separate issue... what's here is supposed to be what the CSV file represents.

I have a "TODO" in there to parse units from the headers... and really even the CSDMS names are a convenience... both of those things should arguably come out in the future or at least be overridable (but expanding the `forcing_config` struct would have been a heavier lift). Point being, if the units should be changed, they should be changed to whatever the source format is believed to be, and will probably be drawn more directly from it in the future anyway.

mattw-nws:
Believe the requested changes have been addressed.

src/realizations/catchment/Simple_Lumped_Model_Realization.cpp
```diff
@@ -109,7 +109,7 @@ double Simple_Lumped_Model_Realization::get_response(time_step_t t, time_step_t
         precip = this->forcing->get_value("precip_rate", t_unix, dt, ""); // classic forcing object/format
     }
     catch (const std::exception& e){
-        precip = this->forcing->get_value(CSDMS_STD_NAME_RAIN_RATE, t_unix, dt, ""); // CsvPerFeatureForcingProvider
+        precip = this->forcing->get_value(CSDMS_STD_NAME_RAIN_VOLUME_FLUX, t_unix, dt, ""); // CsvPerFeatureForcingProvider
```
hellkite500, line 112: This should be `CSDMS_STD_NAME_LIQUID_EQ_PRECIP_RATE`.  The call to hymod does `precip*dt` to convert `m/s` to `m`.

include/realizations/catchment/Bmi_Multi_Formulation.hpp
```diff
@@ -275,11 +275,30 @@ namespace realization {
...
          */
         time_t get_forcing_output_time_begin(const std::string &forcing_name) override {
+            std::string var_name = forcing_name;
+            if(var_name == "*" || var_name == ""){
+                // when unspecified, assume all data is available for the same range.
+                // Find one that successfully returns...
+                for(std::map<std::string,std::shared_ptr<forcing::ForcingProvider>>::iterator iter = availableData.begin(); iter != availableData.end(); ++iter)
+                {
+                    var_name = iter->first;
+                    std::cerr<<"Trying var_name "<<var_name<<std::endl;
```
hellkite500, line 285: Also, do we want to remove these prints?

mattw-nws: Yes, thought I did, sorry 🤔

include/realizations/catchment/Bmi_Multi_Formulation.hpp
```diff
@@ -275,11 +275,30 @@ namespace realization {
          * @return The inclusive beginning of the period of time over which this instance can provide this data.
          */
         time_t get_forcing_output_time_begin(const std::string &forcing_name) override {
+            std::string var_name = forcing_name;
+            if(var_name == "*" || var_name == ""){
+                // when unspecified, assume all data is available for the same range.
```
robertbartel, line 280: > when unspecified, assume all data is available for the same range.

I see a potential issue with this.  We will (eventually) implement functionality where the last module feeds something back into the first.  We will need an extra data point, at a time before the "beginning," for an item that is (at least later) a forcing output.

How we implement things later will make a difference as to whether this directly leads to a conflict, but that's incidental to the real problem.  This usage is inconsistent with the design of _ForcingProvider_, the design itself is lacking.

We need to adjust the interface to add another function to do what we are trying to force this to do, which is get a "beginning" for _the provider_.  This function should (only) get the "beginning" for a particular forcing property.

include/realizations/catchment/Bmi_Multi_Formulation.hpp
```diff
@@ -293,10 +312,30 @@ namespace realization {
          */
         time_t get_forcing_output_time_end(const std::string &forcing_name) override {
             // If not found ...
-            if (availableData.empty() || availableData.find(forcing_name) == availableData.end()) {
-                throw runtime_error(get_formulation_type() + " cannot get output time for unknown " + forcing_name);
+            std::string var_name = forcing_name;
+            if(var_name == "*" || var_name == ""){
+                // when unspecified, assume all data is available for the same range.
```
robertbartel, line 316: Seem comments above for `get_forcing_output_time_begin`.

hellkite500 (dismissed):

include/realizations/catchment/Bmi_Multi_Formulation.hpp
mattw-nws: Concur. The original implementation/first iteration attempted to implement the feature without modifying the interface. A later iteration would be better suited to modify the interface where appropriate, and the assumption of all variables having the same time range holds in the present state of the system. This PR is a bugfix on the first iteration, not a second iteration.

robertbartel: I'm not super clear on the entire scope of the bugs from #311 that this is intended to address, but I consider using `get_forcing_output_time_begin` incorrectly to be a bug.   Regardless of whether it's separate from the bugs to fix in this PR, if we had sufficient test coverage in place for `Bmi_Multi_Formulation::get_forcing_output_time_begin`, I am pretty sure this change wouldn't pass.

That said, I'm alright going forward with this, as long as we are all on the same page.  I see this as not just deferring the fix of bug _A_, but actually extending bug _A_, as part of efforts to fix bug _B_.  That isn't the kind of thing we need to become desensitized to (or let ourselves do very often), so I think it's important we emphasize when it happens.

I'll plan to start working on the interface (and adding tests we are missing for `Bmi_Multi_Formulation`) once this PR is committed.

hellkite500:
@robertbartel @mattw-nws I have held off merging this due to some of the conversation around this interface and the unintended consequences that may come from "extending a bug".  Just wanted everyone to chime in on where we are with this PR.  If there is a clear path that isn't buggy to resolving the root cause, I would be OK taking that route as long as we can expedite the process so that we have a stable branch that is usable for the short term.

If that cannot be expedited, then using this PR as a workaround while we sort that out, while not ideal, may be necessary just to keep all teams moving forward progressively.  Let me know what your thoughts are.

mattw-nws:
Concur with @robertbartel that what we have here is not ideal, to say the least. The goal in #311 was to refactor without changing the interface, and that seemed possible given the nature of inputs at the time/at present (all `ForcingProvider`s had uniform time windows across all variables), so this was done. At the time of #311 it was acknowledged that the interface would need extending and renaming. Since a rapid fix was needed in this PR for multi-BMI formulations, overhauling the interface in this PR did not seem prudent.

I think we're generally in agreement that a separate signature needs to be added to the interface, something like `get_forcing_output_time_begin()` (and also `get_forcing_output_time_end()`) with no parameters to get the earliest (and latest) available data is probably necessary. [The legacy `Forcing` object had this method, which is what led to the problem.](https://github.com/mattw-nws/ngen/commit/b5ea393fa874fc7356094a4cc2830bf484981d2c#diff-ae1264775c477c87bced0f5ed3e382966fe721ec49ee36e7498e86fd01b9727dR463). But that and other recognized changes to the interface are larger issues that can't be done quickly with quality, and the scenarios in which they are needed (async coupling/feedback variables, or more complex forcing providers) are also far from trivial and will not be realized soon. Particularly, those changes are likely to drive changes to the interface much more comprehensively than this... so it might likely require double-work if we strike out to fix the interface anticipating what we might need rather than responding to needs with changes to the interface as they arise.

robertbartel:
@mattw-nws @hellkite500, my impression is that certain urgent tasks (in particular some experimentation and data collection) are being held up by this.  I think putting a workaround in place right away is acceptable given the circumstances, since we can't be sure how long it will take to get a more ideal solution ready.  And as long as we don't let this linger.

I've already started working on tests to help avoid getting into similar spots.   I can volunteer to work on the interface changes right after, though collective design discussions might be in order first.

hellkite500 (approved):

hellkite500 (approved):

PR #339, opened 2021-11-16
Execute git submodule commands from project root

hellkite500:
Quick patch of the recent CMakeLists build magic to help automate the usage of git submodules.  `git submodule` commands can only be run from the root git directory, so make sure to `cd` to the root before initiating calls to `git submodule`


mattw-nws (approved):
👍

PR #340, opened 2021-11-17
Documenting Udunits dependency and setup

robertbartel:
Adding documentation on the new Udunits dependency to the standard _Dependencies_ markdown document, include details on different setup options.

### DIFF
--- a/doc/DEPENDENCIES.md
+++ b/doc/DEPENDENCIES.md
@@ -116,7 +116,9 @@ Basically, the library and needed development files can be installed via typical
 
 The Udunits library and development files can be installed via standard OS package management tools.  The ngen project will automatically detect and use these.  Note that both the library itself and development files must be installed, which are typically two separate, similarly named packages.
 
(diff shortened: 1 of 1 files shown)


mattw-nws (commented):
These look good, apologies, I should have included something like this in my PR...I thought it worked too easily for docs I guess, which was false.

Only thought I had was, do we want to explicitly mention the libexpat dependency somewhere... and that probably starts to get out of scope fast. So probably good as is unless you think otherwise.

doc/DEPENDENCIES.md
```diff
@@ -101,6 +102,52 @@ The variable should be set to the value of the **boost root directory**, which i
...
+
+#### Option 1: Package Manager Install
+
+The Udunits library and development files can be installed via standard OS package management tools.  The ngen project will automatically detect and use these.  Note that both the library itself and development files must be installed, which are typically two separate, similarly named packages.
+
+The ngen build will automatically search for these, so no additional steps are required to configure the build or the environment once they are installed.  Installation can be done using the typical OS package manager, like `yum` or `apt`.  Different distributions like to call the packages different things (e.g., _udunits2-dev_, _udunits2-devel_, _libudunits2-dev_, etc.), so you may need to do some searching.  Also, for CentOS, you'll need [the EPEL repo enabled](https://docs.fedoraproject.org/en-US/epel/)
+
+Packages for both the _library_ and _development files_ are needed.  Package managers usually make this easy, but just make sure to pay attention to it if things do not work as expected.
```
hellkite500, line 121: Might be worth dropping this link in to help find the correct package names for different distros:
https://pkgs.org/search/?q=udunits

robertbartel: Added in `71cac887c60c3e2c3dafc64ca6ccebd150b18d30`.

hellkite500 (approved):

PR #341, opened 2021-11-18
Unit tests for multi-module BMI formulation

robertbartel:
Adding initial set of unit tests, under a separate `test_bmi_multi` target,  and automated test workflows for testing the _Bmi_Multi_Formulation_ class.


test/CMakeLists.txt
```diff
@@ -168,6 +168,20 @@ if (${NGEN_ACTIVATE_PYTHON})
     )
 endif()
 
+########################## BMI Multi Tests
+if (BMI_C_LIB_ACTIVE OR ${BMI_FORTRAN_ACTIVE} OR ${NGEN_ACTIVATE_PYTHON})
```
hellkite500, line 172: It appears the runners don't like this...
* test_unit_ubuntu_latest
* test_mpi_remote_nexus_ubuntu_latest
* test_bmi_c_ubuntu_latest
* test_bmi_python_ubuntu_latest
* test_unit_macos_latest
* test_bmi_c_macos_latest
* test_bmi_python_macos_latest

All fail with some form of
```cmake
CMake Error at test/CMakeLists.txt:172 (if):
  if given arguments:

    "BMI_C_LIB_ACTIVE" "OR" "OR" "ON"

  Unknown arguments specified


-- Configuring incomplete, errors occurred!
```

hellkite500: Also the `build_and_run_model` job is affected in the same way.

hellkite500: Don't know why 64b9a1fb7bbaa5855daf708a9652ab19f2d2f05e fixed the `build_and_run_model` problem, but none of the others...

hellkite500: c6f4012a0641731209dcb14b1e28d846b0d6906b the nuke from orbit solution? 😄

robertbartel: Yeah, so this is reasonable to do, because I had initially intended on being able to pivot in the test scenarios based on what was active, but I eventually abandoned that.  To run this, all three things have to be active.  So there's no need for the check.

I'm not at all sure why the check worked fine for me locally and not on the runners, but given it isn't needed anymore, I didn't feel like finding out.

robertbartel: > Don't know why [64b9a1f](https://github.com/NOAA-OWP/ngen/commit/64b9a1fb7bbaa5855daf708a9652ab19f2d2f05e) fixed the `build_and_run_model` problem, but none of the others...

I'm going to guess that it has to do with the version of CMake somehow ...

hellkite500 (approved):

PR #346, opened 2021-11-26
Support look-back for nest BMI module input values and bootstrapping of defaults.

robertbartel:
Add support for allowing a nested formulation in a BMI multi formulation to use as a provider a later nested formulation.  Previously, only earlier nested formulations could be used like this, where earlier/later is with respect to their listed order in the realization config.  Since a later nested formulation's outputs won't have been (re)calculated yet, this effectively is a look-back to the previous time step.

Also, adding support for configuring a default value for nested formulation outputs, including having it be used as the value during the first time step in the aforementioned look-back scenario.

### DIFF
--- a/include/forcing/OptionalWrappedProvider.hpp
+++ b/include/forcing/OptionalWrappedProvider.hpp
@@ -277,6 +277,8 @@ namespace forcing {
          */
         bool setWrappedProvider(ForcingProvider *provider) override {
             // Disallow re-setting the provider if already validly set
+            // TODO: consider if this is still appropriate, given backing providers can be valid without providing
+            //  everything (because of defaults), and so a new provider may supply more or less of what is needed
             if (isWrappedProviderSet()) {
                 setMessage = "Cannot change wrapped provider after a valid provide has already been set";
                 return false;
@@ -288,13 +290,25 @@ namespace forcing {
                 return false;
             }
 
-            // Confirm this will provide everything needed, though allow for defaults when available
+            // Check this provides everything needed, accounting for defaults (and also tallying the outputs provided)
+            unsigned short providedByProviderCount = 0;
             for (const string &requiredName : providedOutputs) {
-                // When not provided by this provider and there is not a default value set up ...
-                if (!isSuppliedByProvider(requiredName, provider) && !isSuppliedWithDefault(requiredName)) {
+                // If supplied by the provider, increment our count and continue to the next required output name
+                if (isSuppliedByProvider(requiredName, provider)) {
+                    ++providedByProviderCount;
+                    continue;
+                }
+                // If something isn't supplied by the provider (i.e., the "else"), there has to be a default, or we bail
+                else if (!isSuppliedWithDefault(requiredName)) {
                     setMessage = "Given provider does not provide the required " + requiredName;
                     return false;
                 }
+            }
+
+            // Also make sure it provides at least 1 relevant output (possible no outputs were "needed" due to defaults)
+            if (providedByProviderCount == 0) {
+                setMessage = "Given provider does not provide any required outputs (all satisfied via defaults)";
+                return false;
             }
 
             // If this is good, set things and return true
--- a/include/forcing/OptionalWrappedProvider.hpp
+++ b/include/forcing/OptionalWrappedProvider.hpp
@@ -243,38 +243,6 @@ namespace forcing {
         }
 
         /**
-         * Record that there was an instance of a default value being used and manage the default usage "waits."
-         *
-         * When constructed, this instance may have been provided with a number of default usage "waits" for one or more
-         * outputs; i.e., a number of times it should wait to proxy a particular output value from the backing provider
-         * and instead allow a default value to override the aforementioned backing value.  This function, which should
-         * be called by @ref get_value any time a default is used, performs any modifications to the object's state that
-         * are necessary when a default is used, in particular with respect to managing "waits."
-         *
-         * In this base implementation, a recorded default usage does not count against the required number of "waits"
-         * for an output unless/until there is a wrapped backing provider set that can provide the given output.
-         * Otherwise, positive wait counts for this output be reduced by 1.  This allows negative wait counts to be used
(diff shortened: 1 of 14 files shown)


include/forcing/OptionalWrappedProvider.hpp
```diff
@@ -0,0 +1,361 @@
...
+         * a value for it.
+         *
+         * @param output_name The name of the output for which a default was used.
+         */
+        virtual void recordUsingDefault(const string &output_name) {
+            // Don't bother doing anything if there aren't waits assigned for this output
+            // Also, in this implementation, don't count usages until there is a backing provider that can provide this
+            auto waits_it = defaultUsageWaits.find(output_name);
+            if (waits_it != defaultUsageWaits.end() && isSuppliedByWrappedProvider(output_name)) {
+                // A value of 0 should be inferred and cleared from the collection, so ...
```
hellkite500, line 324: So what exactly happens with a value of 0???

include/realizations/catchment/Bmi_Multi_Formulation.hpp
```diff
@@ -506,6 +508,50 @@ namespace realization {
             return module->get_var_value_as_double(var_name);
         }
 
+        /**
+         * Initialize the deferred associations with the providers in @ref deferredProviders.
+         *
+         * During nested formulation creation, when a nested formulation requires as input some output expected from
+         * soon-to-be-created (i.e., later in execution order) formulation (e.g., in a look-back scenario to an earlier
+         * time step), then a deferred provider gets registered with the nested module and has a referenced added to
```
hellkite500, line 516: "and has a reference added to"

hellkite500 (approved):

hellkite500:
A couple additional thoughts.  Should the `default`/`init` value not found in configuration attempt to use the `bmi` `get_value` to query the formulation for a value?

Would we ever want more than a single time step "look back"?

mattw-nws:
This didn't make AGU/v0.1.0 but I have a feeling it will be wanted for "MVP" to support [SoilFreezeThaw](/NOAA-OWP/SoilFreezeThaw) along with #343 .

include/forcing/OptionalWrappedProvider.hpp
robertbartel: @hellkite500, the intent was for there never be a value less than `1` in the `defaultUsageWaits` map.  If we assume this condition when an object is created, then it should hold throughout, as any values equal to 1 before the decrementing step are instead removed from the map entirely.

However, your comment prompted me to double-check and discover that it is possible for the `> 0` condition to be violated.  There is this constructor: 

```c++
OptionalWrappedProvider(vector<string> providedOuts, map<string, double> defaultVals, map<string, int> defaultWaits)
```

This validates the keys in `defaultWaits` (needed for other reasons) but not that the values are all greater than 0.

I could fix this, but first I'd like some opinions on which is the right approach.  Instead of what I was originally trying, does it make more sense to do as you had expected and check in the `recordUsingDefault` function for the `<= 0` case?  It's slightly less efficient but more intuitive.

robertbartel:
>Should the default/init value not found in configuration attempt to use the bmi get_value to query the formulation for a value?

@hellkite500, I'm not sure I understand exactly what you are asking, though it does seem as though perhaps this may not sufficiently handle the case when a default is needed but not provided.  I'll have to double-check that, but I'd like to make sure I understand what you were suggesting before I do.

>Would we ever want more than a single time step "look back"?

Possibly, but that has the potential to be a considerably more significant undertaking.  E.g., how far back do we save?  How do we control what is saved beyond one time step?  Where/how is data being saved?  Are we doing anything to ensure we don't exceed storage capacity?  

Unless there is something specific that is fairly simple/similar and immediately needed, we should handle expanding this functionality separately.  If there are any particular use cases that we have need for, we'll also need those details.

robertbartel:
>the intent was for there never be a value less than 1 in the defaultUsageWaits map. If we assume this condition when an object is created, then it should hold throughout, as any values equal to 1 before the decrementing step are instead removed from the map entirely.

@hellkite500, I realized after our discussion last week, as I was working on fixing the "problem" with this, that I been wrong it how I remembered this was supposed to work.   But I think we need to change things anyway.

In short, things were made to support perpetually providing a default value for some output.  The 0-value case you originally asked about was to allow this, as was the possibility of the constructor receiving wait values less than 1.

Or, stated otherwise, the intent was actually for there never be a value less than 1 in the `defaultUsageWaits` map, _except_ when we would not want a wait value to be altered or removed.  And the latter would be for values indicating that we wanted to perpetually use a default value.

After our talk, I don't think this should be handled by the framework, so I've modified things to take that out.  It would have changed the nature of the type from being a wrapper/proxy to being a data source.

Also,`setup_nested_deferred_provider` in _Bmi_Multi_Formulation_ is written so that these deferred providers will not have an initialized default if none is provided via the realization config.  And I've added some unit tests to make sure the wrapper falls back to the back to the value of the source provider if no default is given.

robertbartel:
Added some test case that were not covered after conversation with @mattw-nws.  Also refactored the test class a bit.  Plus, rebased against latest `master`.

mattw-nws:
This (unit conversion stderr text removed:

```
/Users/runner/work/ngen/ngen/test/realizations/catchments/Bmi_Multi_Formulation_Test.cpp:653: Failure
  expected
    Which is: 4.86646e-08
  response
    Which is: 2.78098e-08
[  FAILED  ] Bmi_Multi_Formulation_Test.GetResponse_3_b (220 ms)
```

Looks like you duplicated test `GetResponse_0_b` before f5b33bc, you'll want to propagate the changes to the values in that commit.

hellkite500 (approved):

mattw-nws:
Merging, but noting possible gotcha: I don't see a test that makes sure reading `default_output_values` through from the realization config works. Testing the use of default vals is done via providing defaults direct via C++ API, though. Just keep an eye on whether it works when doing this via the realization config (wouldn't worry about it except this bit me in #343!).

PR #359, opened 2021-12-13
temporarily apply time scaling in driver

hellkite500:
So far, all BMI models produce output `per timestep`.  So we need to adjust the formulation response based on the engine's clock, in this case 3600 seconds.  Note that this should probably be moved, and a TODO in the code exists to remind us to do so in the future.

## Additions

- Divide response by 3600 seconds

## Testing

1. AGU Demos run with this change and produce appropriately scaled outputs...

## Todos

- Eventually move this functionality into the formulation...

## Checklist

- [x] Linux
- [x] MacOs

### DIFF
--- a/src/NGen.cpp
+++ b/src/NGen.cpp
@@ -294,6 +294,11 @@ int main(int argc, char *argv[]) {
         {
           response *= (catchment_collection->get_feature(id)->get_property("area_sqkm").as_real_number() * 1000000);
         }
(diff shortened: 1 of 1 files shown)


mattw-nws:
Added #360 to follow on. Also, we should probably have some integration test that failed for this change, but there isn't one.

mattw-nws (approved):

PR #367, opened 2022-01-24
Add initial Data Provider template.

donaldwj:

## Additions

Provisional template for generic data provider classes
-

## Removals

None
-

## Changes
None
-

## Testing

1. 

## Notes

-

## Todos
When design is accepted create binary forcing per feature classes as derived class from this template.

## Checklist

1.

- [ ] Linux

### DIFF
--- a/include/forcing/DataProvider.hpp
+++ b/include/forcing/DataProvider.hpp
@@ -20,6 +20,12 @@ namespace data_access
 
         public:
 
+        enum ReSampleMethod
+        {
+            MEAN,
+            SUM
+        };
+
         virtual ~DataProvider() = default;
 
         /** Return the variables that are accessable by this data provider */
@@ -28,11 +34,14 @@ namespace data_access
 
         /** Return the first valid time for which data from the request variable  can be requested */
 
-        virtual int get_data_start_time(std::string var) = 0;
+        virtual long get_data_start_time() = 0;
 
         /** Return the last valid time for which data from the requested variable can be requested */
 
-        virtual int get_data_stop_time(std::string var) = 0;
+        virtual long get_data_stop_time() = 0;
+
+        /** Return the last valid time for which data from the requested variable can be requested */
+        virtual long record_duration() = 0;
 
         /**
          * Get the index of the data time step that contains the given point in time.
@@ -55,11 +64,12 @@ namespace data_access
          * @param init_time_epoch The epoch time (in seconds) of the start of the time period.
          * @param duration_seconds The length of the time period, in seconds.
          * @param output_units The expected units of the desired output value.
+         * @param m How data is to be resampled if there is a mismatch in data alignment or repeat rate
          * @return The value of the forcing property for the described time period, with units converted if needed.
          * @throws std::out_of_range If data for the time period is not available.
          */
         virtual data_type get_value(selection_type selector, const std::string &variable_name, const time_t &init_time, const long &duration_s,
-                                 const std::string &output_units) = 0;
+                                 const std::string &output_units, ReSampleMethod m=SUM) = 0;
 
         private:
     };
--- a/include/forcing/DataProvider.hpp
+++ b/include/forcing/DataProvider.hpp
@@ -1,0 +1,68 @@
+#ifndef NGEN_DATAPROVIDER_HPP
+#define NGEN_DATAPROVIDER_HPP
+
+
+namespace data_access
+{
+    /** An abstraction for classes that provider access to data.
+    *
+    * Data may be pre-provided from some external source, internally calculated by the implementing type, or some
+    * combination of both.
+    */
+
+    template <class data_type, class selection_type> class DataProvider
+    {
+        /** This class provides a generic interface to data services
+        *
+        */
+
+        public:
+
+        virtual ~DataProvider() = default;
+
+        /** Return the variables that are accessable by this data provider */
+
+        virtual const std::veector<std::string> get_avaliable_variable_names() = 0;
+
+        /** Return the first valid time for which data from the request variaable  can be requested */
+
+        virtual get_data_start_time(std::string var) = 0;
+
+        /** Return the last valid time for which data from the requested variablle can be requested */
+
+        virtual get_data_stop_time(std::string var) = 0;
+
+        /**
+         * Get the index of the data time step that contains the given point in time.
+         *
+         * An @ref std::out_of_range exception should be thrown if the time is not in any time step.
+         *
+         * @param epoch_time The point in time, as a seconds-based epoch time.
+         * @return The index of the forcing time step that contains the given point in time.
+         * @throws std::out_of_range If the given point is not in any time step.
+         */
+        virtual size_t get_ts_index_for_time(const time_t &epoch_time) = 0;
+
+        /**
+         * Get the value of a forcing property for an arbitrary time period, converting units if needed.
+         *
+         * An @ref std::out_of_range exception should be thrown if the data for the time period is not available.
+         *
+         * @param output_name The name of the data property of interest.
+         * @param init_time_epoch The epoch time (in seconds) of the start of the time period.
+         * @param duration_seconds The length of the time period, in seconds.
+         * @param output_units The expected units of the desired output value.
+         * @return The value of the forcing property for the described time period, with units converted if needed.
+         * @throws std::out_of_range If data for the time period is not available.
+         */
+        virtual data_type get_value(selection_type selector, const std::string &variable_name, const time_t &init_time, const long &duration_s,
+                                 const std::string &output_units) = 0;
+
+        private:
+    };
+
+}
+
+
+
+#endif // NGEN_DATAPROVIDER
--- a/include/forcing/DataProvider.hpp
+++ b/include/forcing/DataProvider.hpp
@@ -48,7 +48,8 @@ namespace data_access
          *
          * An @ref std::out_of_range exception should be thrown if the data for the time period is not available.
          *
-         * @param output_name The name of the data property of interest.
+         * @param selector Data required to establish what subset of the stored data should be accessed
+         * @param variable_name The name of the data property of interest.
          * @param init_time_epoch The epoch time (in seconds) of the start of the time period.
          * @param duration_seconds The length of the time period, in seconds.
          * @param output_units The expected units of the desired output value.
--- a/include/forcing/DataProvider.hpp
+++ b/include/forcing/DataProvider.hpp
@@ -23,7 +23,9 @@ namespace data_access
         enum ReSampleMethod
         {
             MEAN,
-            SUM
+            SUM,
+            FRONT_FILL,
+            BACK_FILL
         };
 
         virtual ~DataProvider() = default;
--- a/include/forcing/DataProvider.hpp
+++ b/include/forcing/DataProvider.hpp
@@ -30,7 +30,7 @@ namespace data_access
 
         /** Return the variables that are accessable by this data provider */
 
-        virtual const std::vector<std::string> get_avaliable_variable_names() = 0;
+        virtual const std::vector<std::string>& get_avaliable_variable_names() = 0;
 
         /** Return the first valid time for which data from the request variable  can be requested */
 
@@ -40,7 +40,7 @@ namespace data_access
 
         virtual long get_data_stop_time() = 0;
 
-        /** Return the last valid time for which data from the requested variable can be requested */
+        /** Return the stride in the time dimension */
         virtual long record_duration() = 0;
 
         /**
--- a/include/forcing/DataProvider.hpp
+++ b/include/forcing/DataProvider.hpp
@@ -6,7 +6,7 @@
 
 namespace data_access
 {
-    /** An abstraction for classes that provider access to data.
+    /** An abstraction for classes that provides access to data.
     *
     * Data may be pre-provided from some external source, internally calculated by the implementing type, or some
     * combination of both.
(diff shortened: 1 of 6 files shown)


include/forcing/DataProvider.hpp
```diff
@@ -0,0 +1,71 @@
...
+        *
+        */
+
+        public:
+
+        virtual ~DataProvider() = default;
+
+        /** Return the variables that are accessable by this data provider */
+
+        virtual const std::veector<std::string> get_avaliable_variable_names() = 0;
```
hellkite500, line 27: Looks like a typo here: `std::veector` -> `std::vector`

include/forcing/DataProvider.hpp
```diff
@@ -0,0 +1,71 @@
+#ifndef NGEN_DATAPROVIDER_HPP
+#define NGEN_DATAPROVIDER_HPP
+
+#include <string>
+
+
+namespace data_access
+{
+    /** An abstraction for classes that provider access to data.
```
hellkite500, line 9: "that provider" -> "that provide"

include/forcing/DataProvider.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+        public:
+
+        virtual ~DataProvider() = default;
+
+        /** Return the variables that are accessable by this data provider */
+
+        virtual const std::veector<std::string> get_avaliable_variable_names() = 0;
+
+        /** Return the first valid time for which data from the request variaable  can be requested */
```
hellkite500, line 29: "request variaable" -> "request variable"

include/forcing/DataProvider.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+        virtual get_data_start_time(std::string var) = 0;
+
+        /** Return the last valid time for which data from the requested variablle can be requested */
```
hellkite500, line 33: "request variablle" -> "request variable"

include/forcing/DataProvider.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+        virtual get_data_stop_time(std::string var) = 0;
+
+        /**
+         * Get the index of the data time step that contains the given point in time.
+         *
+         * An @ref std::out_of_range exception should be thrown if the time is not in any time step.
+         *
+         * @param epoch_time The point in time, as a seconds-based epoch time.
```
hellkite500, line 42: What is the reference point for the epoch time?  Should we formalize that?

include/forcing/DataProvider.hpp
```diff
@@ -0,0 +1,71 @@
...
+         *
+         * @param selector Data required to establish what subset of the stored data should be accessed
+         * @param variable_name The name of the data property of interest.
+         * @param init_time_epoch The epoch time (in seconds) of the start of the time period.
+         * @param duration_seconds The length of the time period, in seconds.
+         * @param output_units The expected units of the desired output value.
+         * @return The value of the forcing property for the described time period, with units converted if needed.
+         * @throws std::out_of_range If data for the time period is not available.
+         */
+        virtual data_type get_value(selection_type selector, const std::string &variable_name, const time_t &init_time, const long &duration_s,
```
hellkite500, line 61: How does this intend to deal with frequency?  Do we need to resample/interpolate/ect to align period and frequency of a data source to the requester?

include/forcing/DataProvider.hpp
donaldwj: In the case of NetCDF we can embed an epoch start time in the files. In other cases maybe set by ENV variables with a fallback to Unix standard epoch?

include/forcing/DataProvider.hpp
donaldwj: The function should resample as needed to get a record of **data_type**. For cases where a sequences of values is intended to be retrieved **data_type** can be a type the incapsulates and array with desired length or it can be multi-dimensional matrix for retrieval of data cubes.

donaldwj: In the example netcdf implementation **data_type** is double so the intent would be to retrieve one value. If the time range was sufficiently large for the file to contain multiple values they should be merged and if it does not span a full file time step a partial result should be returned.

robertbartel: Why separate `variable_name`, `init_time`, and `duration_s` from `selector`?

include/forcing/NetCDFPerFeatureDataProvider.hpp
```diff
@@ -0,0 +1,457 @@
...
+            if ( units_att.isNull() )
+            {
+                native_units = "unknown";      
+            }
+            else
+            {
+                units_att.getValues(native_units);
+            }
+
+            // case where less than one time step is requested
```
hellkite500, line 338: If there is a general solution, drop these cases.  From a cursory look, the general algorithm should work fine in both these cases.

include/forcing/DataProvider.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+    template <class data_type, class selection_type> class DataProvider
+    {
+        /** This class provides a generic interface to data services
+        *
+        */
+
+        public:
+
+        enum ReSampleMethod
```
hellkite500, line 23: I would add `front_fill` and `back_fill` as valid options here.  Not sure how we intend to offer this configurability to users, but these are two standard methods that are useful especially when working with rates over time.

include/forcing/DataProvider.hpp
donaldwj: That was a an attempt to make this API a close to the existing API as possible.

include/forcing/NetCDFPerFeatureDataProvider.hpp
```diff
@@ -0,0 +1,464 @@
...
+            ids.getVar(&string_buffers[0]);
+
+            // initalize the map of catchment-name to offset location and free the strings allocated by the C library
+            size_t loc = 0;
+            for_each( string_buffers.begin(), string_buffers.end(), [&](char* str)
+            {
+                loc_ids.push_back(str);
+                id_pos[str] = loc++;
+                
+                //free(str);
```
hellkite500, line 98: remove this comment if it isn't to be used

include/forcing/NetCDFPerFeatureDataProvider.hpp
```diff
@@ -0,0 +1,464 @@
...
+                break;
+            }
+
+            // read the meta data to get the epoc start
+            auto epoc_att = time_var.getAtt("epoc_start");
+            std::string epoc_start_str;
+
+            if ( epoc_att.isNull() )
+            {
+                epoc_start_str = "01/01/1970 00:00:00";
```
hellkite500, line 187: This should probably log a warning and indicate the "default epoc start time"

include/forcing/NetCDFPerFeatureDataProvider.hpp
```diff
@@ -0,0 +1,464 @@
...
+
+            // scale the time to account for time units and epoc_start
+            // TODO make sure this happens with a FMA instruction
+            time_vals.resize(raw_time.size());
+            std::transform(raw_time.begin(), raw_time.end(), time_vals.begin(), 
+                [&](const auto& n){return n * time_scale_factor + epoc_start_time; });
+                
+
+            // determine the stride of the time array
+            time_stride = time_vals[1] - time_vals[0];
```
hellkite500, line 207: This assumes the data has a known frequency.  This may not always be the case...

test/forcing/NetCDFPerFeatureDataProvider_Test.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+void NetCDFPerFeatureDataProviderTest::TearDown() 
+{
+
+}
+
+//Construct a forcing object
+void NetCDFPerFeatureDataProviderTest::setupForcing()
+{
+    nc_provider = std::make_shared<data_access::NetCDFPerFeatureDataProvider>("/local/ngen/data/huc01/huc_01/forcing/netcdf/huc01.nc");
```
hellkite500, line 52: We shouldn't merge this absolute path...

hellkite500:
This will also require an update to the GH runners to ensure they have the required netcdf dependencies for running the tests.

include/forcing/NetCDFPerFeatureDataProvider.hpp
```diff
@@ -0,0 +1,457 @@
...
+            else if ( time_unit_str ==  "ms" || time_unit_str == "miliseconds" )
+                    time_unit = TIME_SECONDS;
+            else if ( time_unit_str ==  "us" || time_unit_str == "microseconds" )
+                    time_unit = TIME_SECONDS;
+            else if ( time_unit_str ==  "ns" || time_unit_str == "nanoseconds" )
+                    time_unit = TIME_SECONDS;
+            
+            double time_scale_factor;
+
+            switch (time_unit)
```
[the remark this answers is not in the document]
donaldwj: In this case i wrote the switch first, no reason these can not be merged.

include/forcing/NetCDFPerFeatureDataProvider.hpp
```diff
@@ -0,0 +1,457 @@
...
+
+        NetCDFPerFeatureDataProvider(const char* input_path) : 
+            NetCDFPerFeatureDataProvider(std::string(input_path))
+        {
+
+        }
+
+        /** Return the variables that are accessable by this data provider */
+
+        const std::vector<std::string> get_avaliable_variable_names()
```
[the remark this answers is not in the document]
donaldwj: No reason not to.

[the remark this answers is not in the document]
donaldwj: I think I just forgot the &.

include/forcing/DataProvider.hpp
donaldwj: I can add the methods but then `get_data()` must handle them. What is the intended behavior?

include/forcing/NetCDFPerFeatureDataProvider.hpp
donaldwj: Done, it used to be used but I found there was a `freeString()` method that should be used instead.

include/forcing/NetCDFPerFeatureDataProvider.hpp
donaldwj: I agree, I suppose the data provider should be constructed with a logging interface?

include/forcing/NetCDFPerFeatureDataProvider.hpp
donaldwj: We need that measurement currently, do you see other options?

test/forcing/NetCDFPerFeatureDataProvider_Test.cpp
donaldwj: The file is to big to fit in the test data. Possibly try to make slice we can put in test data?

include/forcing/NetCDFPerFeatureDataProvider.hpp
hellkite500: My point here is that this implementation assumes the data occurs at a frequency that is consistent with the first two entries.  If there are cases where that assumption is used in other parts of the code, then the data should be validated for the assumption before applying it.

test/forcing/NetCDFPerFeatureDataProvider_Test.cpp
hellkite500: Yeah, a small test data set should be suitable to check into the repo....

include/forcing/NetCDFPerFeatureDataProvider.hpp
donaldwj: Validation might prove to very expensive if the time dimension is large enough. We certainly would not want to keep revalidating the same file.

include/forcing/NetCDFPerFeatureDataProvider.hpp
donaldwj: Will update

include/forcing/NetCDFPerFeatureDataProvider.hpp
donaldwj: Some options here:

1.  Add frequency to meta data and use that <preferred>
2.  Do some random sampling to see if there is consistent time step
3. Assume the time step is not consistent and use a binary_search to find the correct index positions (O(lg n) instead of O(1) )

It is worth notting that **3** still assumes that the time values are in order and increasing

include/forcing/NetCDFPerFeatureDataProvider.hpp
```diff
@@ -335,99 +325,52 @@ namespace data_access
...
-                    {
-                        rvalue += raw_values[i];
-                    }
+                case MEAN: 
+                { 
+                    // This is getting a length weighted mean
+                    // the data values where allready scaled for where there was only partial use of a data value
+                    // so we just need to do a final scale to account for the differnce between time_stride and duration_s
+                    
+                    double scale_factor = (duration_s > time_stride ) ? (time_stride / duration_s) : (1.0 / (a + b));
```
hellkite500, line 367: should definitely expand the testing to test these various edge cases

include/forcing/NetCDFPerFeatureDataProvider.hpp
mattw-nws: My thought: what is assumed here is going to be the most common use case. We should require some other signal that this assumption should not be used if we need to do something more expensive (metadata in NetCDF file? subclass `NonLinearTimeNetCdfPerFeatureDataProvider`? Flag in forcing configuration in realization config?). We could also do a minimal test to see if the assumption is likely incorrect, that is, see if delta from 0 to 1 times the number of timesteps matches the end time--if not, something is wrong (would not catch all cases where the assumption was false, but likely 99% of them).

include/forcing/NetCDFPerFeatureDataProvider.hpp
```diff
@@ -0,0 +1,425 @@
...
+
+    class NetCDFPerFeatureDataProvider : public DataProvider<double, Catchment_Id>
+    {
+        
+        public:
+
+        enum TimeUnit
+        {
+            TIME_HOURS,
+            TIME_MINUETS,
```
christophertubbs, line 43: I'm spying on your work for other stuff, but here's a typo

donaldwj:
This was added to master with the netcdf forcing code.

PR #387, opened 2022-03-07
Fix 384, prioritize use of BMI potential evapotranspiration 

### LINKED ISSUE #384: Automatic running of internal evapotranspiration routines does not function as expected
## Current behavior

When running a multi-BMI of the [PET module](https://github.com/NOAA-OWP/evapotranspiration) with [CFE](https://github.com/NOAA-OWP/cfe), the framework returns a unit conversion error:

```
Unit conversion error:
Unable to convert mm s^-1 to kg m^-2
Returning unconverted value!
```

This should not happen because the updated PET module does not have precipitation in kg/m2 as part of its BMI spec (open PR here: https://github.com/NOAA-OWP/evapotranspiration/pull/10/files). If you run PET as a single module through Ngen, this error does not occur. This suggests something different is happening in the single vs. multi BMI. The following is what I believe the issue to be.

The model engine defines two standard variable names for potential evapotranspiration (https://github.com/NOAA-OWP/ngen/blob/0d04d52f3b0420a762478eea3943b14d62db12fc/include/realizations/catchment/Bmi_Formulation.hpp#L35):

```
// Supported Standard Names for BMI variables
// This is needed to provide a calculated potential ET value back to a BMI model
#define NGEN_STD_NAME_POTENTIAL_ET_FOR_TIME_STEP "potential_evapotranspiration"

// Taken from the CSDMS Standard Names list
// TODO: need to add these in for anything BMI model input or output variables we need to know how to recognize
#define CSDMS_STD_NAME_POTENTIAL_ET "water_potential_evaporation_flux"
```

The engine also has logic to automatically run its own native evapotranspiration routines if a model in a formulation has one of those two standard names in its output variables (https://github.com/NOAA-OWP/ngen/blob/1b7988969a4a79ec9d205268191362a93995d35a/include/realizations/catchment/Bmi_Module_Formulation.hpp#L374):

```
            // Handle ET requests slightly differently
            //TODO: This should really only happen if neither of these output_names are provided by the module...
            // But this code should also probably be removed in the near future (see GH #297 second comment).
            if (output_name == NGEN_STD_NAME_POTENTIAL_ET_FOR_TIME_STEP || output_name == CSDMS_STD_NAME_POTENTIAL_ET) {
                return calc_et();
            }
```

This has the effect of only running the native model engine evapotranspiration routines when the PET module is loaded in a multi BMI because it is the only module that has the `CSDMS_STD_NAME_POTENTIAL_ET` as an output var.

## Expected behavior

The native evapotranspiration routines should not run when the PET module is used in a multi BMI. The issue is threefold:

1. The logic should run on the **input vars** if we keep it (i.e., it should search for modules that require PET, not those that produce it)
2. The logic should also check if there is a module in the stack producing the required PET (this is more complicated because plenty of modules may or will not use either of the standard names, meaning the number of options is undefined)
3. The PET module provides the same functionalities as the native evapotranspiration routines and for consistency we should prioritize the BMI-enabled module (there is nothing unique about PET as a flux that would suggest it should be handled differently than other modules)

1. Run a multi-BMI formulation with PET and CFE

Interestingly, the error does not manifest when running Noah-OWP-Modular and CFE because the former does not use either standard names as its output var for PET.

hellkite500:
The current implementation uses internal PET kernels to `calc_et()` in the framework when a module request an evapotranspiration variable as input.  This is done regardless of whether or not that variable is available from another BMI module source.  This PR is backwards compatible, and will still provide PET when no BMI module in the stack knows how to produce the PET standard name, but it allows for a BMI module to be used first if it is capable of producing the value.

Closes #384.

## Additions

- protected `check_internal_providers` function to look for forcing variables that aren't available from BMI providers.

## Removals

-

## Changes

- Implement TODO function for finding bmi_var_name as function `get_bmi_var_name`
- If a BMI var name isn't found in output variables or name map, check for internal providers of that data

## Testing

1. Passes all existing unit tests

## Checklist

- x] Changes are limited to a single goal (no scope creep)

- [x] Linux
- [x] MacOS


hellkite500:
Looking at the one test failure, here are the relevant 3 failures:
```console
[ RUN      ] Bmi_C_Formulation_Test.GetOutputLineForTimestep_0_a
/Users/runner/work/ngen/ngen/test/realizations/catchments/Bmi_C_Formulation_Test.cpp:308: Failure
  output
    Which is: "0.000000,-0.000000"
  "0.000000,0.000000"
Unit conversion error: 
[  FAILED  ] Bmi_C_Formulation_Test.GetOutputLineForTimestep_0_a (80 ms)
Unable to convert mm s^-1 to m
[ RUN      ] Bmi_C_Formulation_Test.GetOutputLineForTimestep_1_a
Returning unconverted value!
/Users/runner/work/ngen/ngen/test/realizations/catchments/Bmi_C_Formulation_Test.cpp:324: Failure
  output
    Which is: "-0.000000,0.000000"
  "0.000000,0.000000"
[  FAILED  ] Bmi_C_Formulation_Test.GetOutputLineForTimestep_1_a (114 ms)
Unit conversion error: 
[ RUN      ] Bmi_C_Formulation_Test.GetOutputLineForTimestep_1_b
Unable to convert mm s^-1 to m
Returning unconverted value!
Unit conversion error: 
Unable to convert mm s^-1 to m
Returning unconverted value!
Unit conversion error: 
Unable to convert mm s^-1 to m
/Users/runner/work/ngen/ngen/test/realizations/catchments/Bmi_C_Formulation_Test.cpp:339: Failure
Returning unconverted value!
Unit conversion error: 
  output
Unable to convert mm s^-1 to m
    Which is: "-0.000000,0.000002"
Returning unconverted value!
  "0.000000,0.000002"
Unit conversion error: 
[  FAILED  ] Bmi_C_Formulation_Test.GetOutputLineForTimestep_1_b (133 ms)
```

Not entirely sure why at the moment these values are coming through as `-0.000000`.

hellkite500:
4ccd04c1e19affd9b9182c90ad20bf460e101ec4 might be relevant here...

robertbartel (commented):
Heh, I remember this working differently.  

@hellkite500, I'll wait on final approval, since I expect more changes are going to be needed to address the problem with the checks, but otherwise this looks good.

hellkite500:
I think the failing test may be fixed by d230f8b5f165e5547086f82c40499cec1c163bf2

robertbartel:
@hellkite500, that check still fails.  Is d230f8b actually included in your branch?   I did a quick search of the log and didn't see it.

hellkite500:
>  I did a quick search of the log and didn't see it.~ Ah, never mind. I can't find the commit, but I do see the changes. And they work for me locally. Very curious ...

That commit was merged into master with #389...

Not sure why I cannot get the strings to match now...

mattw-nws:
Might suggest [EXPECT_THAT with MatchesRegex](https://github.com/google/googletest/blob/main/docs/reference/matchers.md#string-matchers), might give better output on the test failure.

hellkite500:
> Might suggest [EXPECT_THAT with MatchesRegex](https://github.com/google/googletest/blob/main/docs/reference/matchers.md#string-matchers), might give better output on the test failure.

Except that is a feature of googlemock, and I'm not finding it.  I thought googlemock and googletest had merged...but we may be on an older commit???

mattw-nws (approved):

hellkite500:
So a non-deterministic error seems to exist relating to this PR.  In the post-merge check:
```console
/Users/runner/work/ngen/ngen/test/realizations/catchments/Bmi_C_Formulation_Test.cpp:310: Failure
Value of: output
Expected: matches regular expression "-?0.000000,-?0.000000"
  Actual: "0.000000,-4.000000"
[  FAILED  ] Bmi_C_Formulation_Test.GetOutputLineForTimestep_0_a (65 ms)
[ RUN      ] Bmi_C_Formulation_Test.GetOutputLineForTimestep_1_a
/Users/runner/work/ngen/ngen/test/realizations/catchments/Bmi_C_Formulation_Test.cpp:326: Failure
Value of: output
Expected: matches regular expression "-?0.000000,-?0.000000"
  Actual: "-4.000000,0.000000"
 ```
 I'll open an issue to track this...

mattw-nws:
> Place your bets now on whether it fails again in the post-merge checks?

Called it, apparently :-\

PR #388, opened 2022-03-07
Added SOURCE_LOC macro

mattw-nws:
Adds an easy way to improve exception messages (or other logging output)

## Additions

Added a `SOURCE_LOC` macro in `all.h` that can be used to easily (and sustainably) report the location in code from which an exception was thrown or a log message was written.

Also adds some examples of use

## Removals

-

## Changes

-

## Testing

1. Added to a Forcing object test, the actual string returned is not tested because this could easily change with unrelated modifications to Forcing.h, so would be a bad test. A commented-out `cerr` write is present that can be used to check the result.

## Screenshots

```
[----------] 1 test from ForcingTest2
[ RUN      ] ForcingTest2.TestForcingDataLoad
Error: Forcing data ../../test/data/forcing/cat-27115-nwm-aorc-variant-derived-format.csv does not have the expected number of columns.  Did you mean to use csvPerFeature forcing provider instead? (/home/mattw-nws/ngen/include/forcing/Forcing.h:658)
[       OK ] ForcingTest2.TestForcingDataLoad (53618 ms)
```

## Notes

-

## Todos

-

## Checklist

1.

- [x] Linux

### DIFF
--- a/include/all.h
+++ b/include/all.h
@@ -1,3 +1,5 @@
+#pragma once
+
 #ifndef __APPLE__
 #define GCC_VERSION (__GNUC__ * 10000 \
                      + __GNUC_MINOR__ * 100 \
@@ -7,3 +9,8 @@
 #include <unique.hpp>
 #endif
 #endif
+
+#define STRINGIFY(x) #x
+#define TOSTRING(x) STRINGIFY(x)
+#define SOURCE_LOC " (" __FILE__ ":" TOSTRING(__LINE__) ")"
+// ^ Credit https://www.decompile.com/cpp/faq/file_and_line_error_string.htm
--- a/include/core/mediator/UnitsHelper.hpp
+++ b/include/core/mediator/UnitsHelper.hpp
@@ -1,5 +1,6 @@
 #include <udunits2.h>
 #include <mutex>
+#include "all.h"
 
 #ifndef NGEN_UNITSHELPER_H
 #define NGEN_UNITSHELPER_H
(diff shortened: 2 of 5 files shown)


include/all.h
```diff
@@ -7,3 +9,8 @@
 #include <unique.hpp>
 #endif
 #endif
+
+#define STRINGIFY(x) #x
+#define TOSTRING(x) STRINGIFY(x)
+#define SOURCE_LOC " (" __FILE__ ":" TOSTRING(__LINE__) ")"
```
hellkite500, line 15: It occurs to me that this solution is also useful with build/verbosity macros. Should use those here to conditionally define `SOURCE_LOC` only for debug builds. In production, this will produce paths that probably aren't meaningful, or at least would only be useful in a bug report with a version we could check out and find the right relative file from the absolute paths this reports.

hellkite500: I agree, just thought it might be odd if you have a build system creating binaries that then report paths that don't even exist when an error is logged.  I'm good with it as is for the time being.

mattw-nws: Yeah, that's not uncommon in my experience--potentially a little confusing but you quickly figure out what's going on. Something that just occurs to me... should we try [embedding the Git commit ID into the file](https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes#_keyword_expansion) and hence the message?

mattw-nws: Eh...that's messier than it at first looks. Maybe not, or maybe later.

hellkite500 (approved):

hellkite500:
After merging this I decided to do a little digging on `#pragma once` and opened a can of worms I wasn't quite expecting.  [Here](https://stackoverflow.com/questions/1143936/pragma-once-vs-include-guards) is a good thread on the subject.  As I read through this, it made me think that we should probably move to an include guard as used throughout the rest of the codebase, e.g. 
```C++
#ifndef _NGEN_ALL
#define _NGEN_ALL

#endif
```
Just to ensure consistence with the rest of the code and not rely on a "non-standard" feature.

PR #389, opened 2022-03-09
precip_rate now computed using the right unit of temperature

hellkite500:
Discovered a bug in the script used to generate the example forcing files.  The density of water is computed using the following function:

```python
 def rho(temp):
     """
         Calculate water density at temperature (degrees C)
     """
     return 999.99399 + 0.04216485*temp - 0.007097451*(temp**2) + 0.00003509571*(temp**3) - 9.    9037785E-8*(temp**4)
```

But the AORC temperature variable is in kelvin, and was being passed without converting to celsius. 

Fixed this in the script and regenerated the example forcing data files to get correct `precip_rate` (in m/s)

## Changes

- There example forcing data files with corrected `precip_rate` column

## Checklist


mattw-nws:
Looks like the Fortran tests--maybe others--have expected result values based on the old incorrect `precip_rate` data. I was afraid of that when we found the issue, since I was about to base an updated test on the same values.

hellkite500:
> Looks like the Fortran tests--maybe others--have expected result values based on the old incorrect `precip_rate` data. I was afraid of that when we found the issue, since I was about to base an updated test on the same values.

I'm updating those now...

test/realizations/catchments/Bmi_Cpp_Formulation_Test.cpp
```diff
@@ -335,7 +335,7 @@ TEST_F(Bmi_Cpp_Formulation_Test, GetOutputLineForTimestep_1_b) {
         formulation.get_response(i++, 3600);
     formulation.get_response(i, 3600);
     std::string output = formulation.get_output_line_for_timestep(i, ",");
-    ASSERT_EQ(output, "0.000000,0.000002");
+    ASSERT_EQ(output, "0.000000,0.000001");
```
mattw-nws, line 338: I'll be kind of surprised if the `-0` thing only happens on C and not C++ ... but we can keep an eye on it.

mattw-nws (approved):

PR #394, opened 2022-03-10
Fix 393

### LINKED ISSUE #392: Failing unit tests on macos
A non-deterministic failure of unit tests has been introduced with #387 and/or possible #389.

## Current behavior

The PR checks were successful on #387, but the post merge checks once again failed the macos bmi c tests.
```console
[  FAILED  ] Bmi_C_Formulation_Test.GetOutputLineForTimestep_0_a
Returning unconverted value!
[  FAILED  ] Bmi_C_Formulation_Test.GetOutputLineForTimestep_1_a
Unit conversion error: 
[  FAILED  ] Bmi_C_Formulation_Test.GetOutputLineForTimestep_1_b
```

With the following errors reported for these tests
```console
[ RUN      ] Bmi_C_Formulation_Test.GetOutputLineForTimestep_0_a
/Users/runner/work/ngen/ngen/test/realizations/catchments/Bmi_C_Formulation_Test.cpp:310: Failure
Value of: output
Expected: matches regular expression "-?0.000000,-?0.000000"
  Actual: "0.000000,-4.000000"
[  FAILED  ] Bmi_C_Formulation_Test.GetOutputLineForTimestep_0_a (65 ms)
```
```console
[ RUN      ] Bmi_C_Formulation_Test.GetOutputLineForTimestep_1_a
/Users/runner/work/ngen/ngen/test/realizations/catchments/Bmi_C_Formulation_Test.cpp:326: Failure
Value of: output
Expected: matches regular expression "-?0.000000,-?0.000000"
  Actual: "-4.000000,0.000000"
[  FAILED  ] Bmi_C_Formulation_Test.GetOutputLineForTimestep_1_a (57 ms)
```
```console
[ RUN      ] Bmi_C_Formulation_Test.GetOutputLineForTimestep_1_b
/Users/runner/work/ngen/ngen/test/realizations/catchments/Bmi_C_Formulation_Test.cpp:341: Failure
Value of: output
Expected: matches regular expression "-?0.000000,0.000001"
  Actual: "-4.000000,0.000001"
[  FAILED  ] Bmi_C_Formulation_Test.GetOutputLineForTimestep_1_b (133 ms)
```

## Expected behavior

These three tests should return `0.000000` (or possibly `-0.000000`, though I'm starting to suspect that the -0.0 is related to the same bug???)

1. Seems non-deterministic, and doesn't fail consistently (local testing on various macbooks haven't been able to reproduce yet...)

### LINKED ISSUE #393: Bug in checking BMI output for forcing provider interface
# TL;DR
BMI_Module_Formulation forcing provider can self reference a requested output variable to a module's own input variable.

# Background
#387 exposed a subtle bug in the implementation of the forcing provider interface function `get_value`.  The logic previously looked like this:

```c++
double get_value(const std::string &output_name, const time_t &init_time, const long &duration_s,
                         const std::string &output_units) override
        {
            // First make sure this is an available output
            const std::vector<std::string> forcing_outputs = get_available_forcing_outputs();
            if (std::find(forcing_outputs.begin(), forcing_outputs.end(), output_name) == forcing_outputs.end()) {
                throw runtime_error(get_formulation_type() + " received invalid output forcing name " + output_name);
            }
            // TODO: do this, or something better, later; right now, just assume anything using this as a provider is
            //  consistent with times
            /*
            if (last_model_response_delta == 0 && last_model_response_start_time == 0) {
                throw runtime_error(get_formulation_type() + " does not properly set output time validity ranges "
                                                             "needed to provide outputs as forcings");
            }
            */

            // Handle ET requests slightly differently
            //TODO: This should really only happen if neither of these output_names are provided by the module...
            // But this code should also probably be removed in the near future (see GH #297 second comment).
            if (output_name == NGEN_STD_NAME_POTENTIAL_ET_FOR_TIME_STEP || output_name == CSDMS_STD_NAME_POTENTIAL_ET) {
                return calc_et();
            }

            // If not ET, now get correct BMI variable name, which may be the output or something mapped to this output.
            // TODO: move this to a dedicated function
            std::string bmi_var_name;
            std::vector<std::string> output_names = get_bmi_model()->GetOutputVarNames();
            if (std::find(output_names.begin(), output_names.end(), output_name) != output_names.end()) {
                bmi_var_name = output_name;
            }
            else {
                for (auto & iter : bmi_var_names_map) {
                    if (iter.second == output_name) {
                        bmi_var_name = iter.first;
                        break;
                    }
                }
            }
            
            double value = get_var_value_as_double(bmi_var_name);

            // Convert units
            std::string native_units = get_bmi_model()->GetVarUnits(bmi_var_name);
            try {
                return UnitsHelper::get_converted_value(native_units, value, output_units);
            }
            catch (const std::runtime_error& e){
                std::cerr<<"Unit conversion error: "<<std::endl<<e.what()<<std::endl<<"Returning unconverted value!"<<std::endl;
                return value;
            }
        }
```

The problematic logic is the following:
```c++
            // If not ET, now get correct BMI variable name, which may be the output or something mapped to this output.
            // TODO: move this to a dedicated function
            std::string bmi_var_name;
            std::vector<std::string> output_names = get_bmi_model()->GetOutputVarNames();
            if (std::find(output_names.begin(), output_names.end(), output_name) != output_names.end()) {
                bmi_var_name = output_name;
            }
            else {
                for (auto & iter : bmi_var_names_map) {
                    if (iter.second == output_name) {
                        bmi_var_name = iter.first;
                        break;
                    }
                }
            }
```

In #387 this code was refactored into `get_bmi_var_name(...)` and the check for internal ET was done after checking if it was available from a BMI module.

# The bug in action
What is problematic with this is the following scenario:

A realization configuration specifies a BMI model which has the following input variable, such as in CFE
```C
static const char *input_var_names[INPUT_VAR_NAME_COUNT] = {
        "water_potential_evaporation_flux"
};
```
Then this BMI module is configured with the following map
```json
"variables_names_map" : {
       "water_potential_evaporation_flux" : "potential_evapotranspiration"
 }
```

Now consider the following operations
- BMI Module formulation is constructed, and `inner_create_formulation` adds a self reference data provider for these two macros
   - ```c++
      // Supported Standard Names for BMI variables
      // This is needed to provide a calculated potential ET value back to a BMI model
      #define NGEN_STD_NAME_POTENTIAL_ET_FOR_TIME_STEP "potential_evapotranspiration"

      // Taken from the CSDMS Standard Names list
      // TODO: need to add these in for anything BMI model input or output variables we need to know how to recognize
      #define CSDMS_STD_NAME_POTENTIAL_ET "water_potential_evaporation_flux"
      ```
- CFE requests an input of `water_potential_evaporation_flux`
- in `set_model_inputs_prior_to_update(...)` this variable is recognized as an input to the model
- the name map is checked for this variable, and we look through forcing providers for `potential_evapotranspiration` and find the provider (self)
- call `get_value("potential_evapotranspiration")`

Now prior to #387  this would always hit the `calc_et()` conditional and early return from the function.  But other variables could potentially have triggered this bug.  Now with the refactored code, we have the following sequence of events.

- Validate `potential_evapotranspiration` is an available forcing output (it is...)
- Check if `potential_evapotranspiration` is a BMI variable
   - look in `get_bmi_model()->GetOutputVarNames()`, don't find it
   - look in `bmi_var_names_map`
      - Find the mapped name (iter.second == "potential_evapotranspiration)
      - set `bmi_var_name = iter.first`, which in this case is `water_potential_evaporation_flux`
- So now we have a `bmi_var_name = water_potential_evaporation_flux`
- Proceed to `get_var_value_as_double(bmi_var_name)`

We have effectively self referenced a requested output variable to an input variable!

# Solution
Make sure `get_bmi_var_name` validates mapped entries as output variables, since for the forcing provider interface, this is the intended semantic.

```c++
        inline void get_bmi_var_name(const std::string &name, std::string &bmi_var_name)
        {
            //check standard output names first
            std::vector<std::string> output_names = get_bmi_model()->GetOutputVarNames();
            if (std::find(output_names.begin(), output_names.end(), name) != output_names.end()) {
                bmi_var_name = name;
            }
            else
            {
                //check mapped names
                std::string mapped_name;
                for (auto & iter : bmi_var_names_map) {
                    if (iter.second == name) {
                        mapped_name = iter.first;
                        break;
                    }
                }
                //ensure mapped name maps to an output variable
                if (std::find(output_names.begin(), output_names.end(), mapped_name) != output_names.end()){
                    bmi_var_name = mapped_name;
                }
                //else not an output variable
            }
        }
```

hellkite500:
Fixes #393
Fixes #392 

## Changes

- Rename `get_bmi_var_name` to `get_bmi_output_var_name` and make it a protected function
- validate name map lookup resolves to an output variable
- Removed the -0.0 allowance in C formulation unit test

## Testing

1. Tested with unit tests and local runs using example realizations, but with standalone CFE and CFE+PET multi-bmi

## Notes

- I had worked up a more generic `get_bmi_alias` function that validates names/alias by type (output, input, any).  But I'm not sure if it is needed here or not...

## Checklist

- [x] Linux
- [x] MacOS

### DIFF
--- a/include/realizations/catchment/Bmi_Module_Formulation.hpp
+++ b/include/realizations/catchment/Bmi_Module_Formulation.hpp
@@ -438,6 +438,8 @@ namespace realization {
             return get_bmi_model()->GetOutputVarNames();
         }
 
+    protected:
+
         /**
          * @brief Get correct BMI variable name, which may be the output or something mapped to this output.
          * 
@@ -468,8 +470,6 @@ namespace realization {
                 //else not an output variable
             }
         }
-
-    protected:
 
         /**
          * @brief Check for implementation of internal calculators/data for a given requsted output_name
--- a/include/realizations/catchment/Bmi_Module_Formulation.hpp
+++ b/include/realizations/catchment/Bmi_Module_Formulation.hpp
@@ -374,7 +374,7 @@ namespace realization {
 
             // check if output is available from BMI
             std::string bmi_var_name;
-            get_bmi_var_name(output_name, bmi_var_name);
+            get_bmi_output_var_name(output_name, bmi_var_name);
             
             if( !bmi_var_name.empty() )
             {
@@ -444,19 +444,28 @@ namespace realization {
          * @param name 
          * @param bmi_var_name 
          */
-        inline void get_bmi_var_name(const std::string &name, std::string &bmi_var_name)
+        inline void get_bmi_output_var_name(const std::string &name, std::string &bmi_var_name)
         {
+            //check standard output names first
             std::vector<std::string> output_names = get_bmi_model()->GetOutputVarNames();
             if (std::find(output_names.begin(), output_names.end(), name) != output_names.end()) {
                 bmi_var_name = name;
             }
-            else {
+            else
+            {
+                //check mapped names
+                std::string mapped_name;
                 for (auto & iter : bmi_var_names_map) {
                     if (iter.second == name) {
-                        bmi_var_name = iter.first;
+                        mapped_name = iter.first;
                         break;
                     }
                 }
+                //ensure mapped name maps to an output variable, see GH #393 =)
+                if (std::find(output_names.begin(), output_names.end(), mapped_name) != output_names.end()){
+                    bmi_var_name = mapped_name;
+                }
+                //else not an output variable
             }
         }
 
--- a/test/realizations/catchments/Bmi_C_Formulation_Test.cpp
+++ b/test/realizations/catchments/Bmi_C_Formulation_Test.cpp
@@ -307,7 +307,7 @@ TEST_F(Bmi_C_Formulation_Test, GetOutputLineForTimestep_0_a) {
 
     formulation.get_response(0, 3600);
     std::string output = formulation.get_output_line_for_timestep(0, ",");
-    EXPECT_THAT(output, MatchesRegex("-?0.000000,-?0.000000"));
+    EXPECT_THAT(output, MatchesRegex("0.000000,0.000000"));
 }
 
 /** Simple test of output with modified variables. */
@@ -323,7 +323,7 @@ TEST_F(Bmi_C_Formulation_Test, GetOutputLineForTimestep_1_a) {
     // OUTPUT_VAR_1 first.
     formulation.get_response(0, 3600);
     std::string output = formulation.get_output_line_for_timestep(0, ",");
-    EXPECT_THAT(output, MatchesRegex("-?0.000000,-?0.000000"));
+    EXPECT_THAT(output, MatchesRegex("0.000000,0.000000"));
 }
 
 /** Simple test of output with modified variables, picking time step when there was non-zero rain rate. */
@@ -338,7 +338,7 @@ TEST_F(Bmi_C_Formulation_Test, GetOutputLineForTimestep_1_b) {
         formulation.get_response(i++, 3600);
     formulation.get_response(i, 3600);
     std::string output = formulation.get_output_line_for_timestep(i, ",");
-    EXPECT_THAT(output, MatchesRegex("-?0.000000,0.000001"));
+    EXPECT_THAT(output, MatchesRegex("0.000000,0.000001"));
 }
 
 TEST_F(Bmi_C_Formulation_Test, determine_model_time_offset_0_a) {


hellkite500:
For reference, here is what I was thinking about for a generalized `get_bmi_alias` function for a Module Formulation

```c++
    protected:

        enum BMI_VAR_TYPE{ BMI_INPUT_VAR, BMI_OUTPUT_VAR, BMI_ANY_VAR};

        /**
         * @brief Get correct BMI variable name, which may be the output or something mapped to this output.
         * 
         * @param name 
         * @param bmi_var_name 
         */
        inline void get_bmi_alias(const std::string &name, std::string &bmi_var_name, BMI_VAR_TYPE type)
        {
            std::vector<std::string> names;
            switch(type){
                case BMI_OUTPUT_VAR:
                    //standard output names fi
                    names = get_bmi_model()->GetOutputVarNames();
                case BMI_INPUT_VAR:
                    names = get_bmi_model()->GetInputVarNames();
                case BMI_ANY_VAR:
                    names = get_bmi_model()->GetOutputVarNames();
                    std::vector<std::string> tmp = get_bmi_model()->GetInputVarNames();
                    //dump input names after output names
                    names.insert(names.end(), tmp.begin(), tmp.end());
            }
            if (std::find(names.begin(), names.end(), name) != names.end()) {
                bmi_var_name = name;
            }
            else
            {
                //check mapped/alias names
                std::string mapped_name;
                for (auto & iter : bmi_var_names_map) {
                    if (iter.second == name) {
                        mapped_name = iter.first;
                        break;
                    }
                }
                if (std::find(names.begin(), names.end(), mapped_name) != names.end()){
                    bmi_var_name = mapped_name;
                }
                //else not a BMI recognizable variable for given type
            }
        }
```

And then `get_value` would call it, knowing it is looking for valid output vars, like so
```c++
get_bmi_alias(output_name, bmi_var_name, BMI_OUTPUT_VAR);
```

mattw-nws (approved):

PR #402, opened 2022-03-23
Fix t-route test failure when cache is used

hellkite500:
fix action cache issue where t-route is installed with egg links but the .venv cache can't find the links when restored from cache.


mattw-nws:
Went ahead and changed the title since things are still installed with `-e` for now, other than that this is the best solution for the moment.

mattw-nws (approved):

PR #405, opened 2022-04-04
Bmi array support

hellkite500:
This PR allows a BMI module to provide an array of values as a single 1D output which can then be passed/consumed by another BMI module in a Multi BMI configuration.

## Additions

- `get_values` function added to ForcingProvider, with a default implementation
- BMI Module Formulation custom override of `ForcingProvider::get_values`
- `get_converted_values` function added to unit helper for unit conversion of an array of values
- BMI utilities module with function to cast and copy values based on string type tags 
- Test module for testing setting/getting array of values across two C++ BMI modules

## Removals

- Deprecated templated `GetValue( std::string & )` overload from BMI adapters.  Functionality moved to BMI utilities free function.

## Changes

- `check_internal_providers` now returns a vector of values instead of a single value
- add new array variable to c++ bmi test model
- updated implementation `set_model_inputs_prior_to_update` which checks for array like values and handles them accordingly.

## Testing

1. New functionality tested, all existing tests passing locally

## Checklist

- [x] Linux
- [x] MacOS

### DIFF
--- a/.github/workflows/test_and_validate.yml
+++ b/.github/workflows/test_and_validate.yml
@@ -287,6 +287,18 @@ jobs:
           mod-dir: "extern/test_bmi_c/"
           targets: "testbmicmodel"
 
+      #make sure cxx bmi is initialized/ready
+      - uses: ./.github/actions/ngen-submod-build
+        with: 
+          mod-dir: "extern/bmi-cxx/"
+
+      - name: Build Submodules
+        id: submod_build_4
+        uses: ./.github/actions/ngen-submod-build
+        with:
+          mod-dir: "extern/test_bmi_cpp/"
+          targets: "testbmicppmodel"
+
       - name: Build Unit Tests
         uses: ./.github/actions/ngen-build
         with:
@@ -314,6 +326,10 @@ jobs:
       - uses: ./.github/actions/clean-build
         with:
           build-dir: ${{ steps.submod_build_3.outputs.build-dir }}
+      - uses: ./.github/actions/clean-build
+        with:
+          build-dir: ${{ steps.submod_build_4.outputs.build-dir }}
+
 
   # TODO: fails due to compilation error, at least in large part due to use of POSIX functions not supported on Windows.
   # TODO: Need to determine whether Windows support (in particular, development environment support) is necessary.
--- a/include/core/mediator/UnitsHelper.hpp
+++ b/include/core/mediator/UnitsHelper.hpp
@@ -11,7 +11,7 @@ class UnitsHelper {
 
     static double get_converted_value(const std::string &in_units, const double &value, const std::string &out_units);
 
-    static double* get_converted_values(const std::string &in_units, double* values, const std::string &out_units, const size_t & count);
+    static double* convert_values(const std::string &in_units, double* values, const std::string &out_units, double* out_values, const size_t & count);
 
     private:
     static cv_converter* get_converter(const std::string &in_units, const std::string& out_units, ut_unit*& to, ut_unit*& from);
--- a/include/core/mediator/UnitsHelper.hpp
+++ b/include/core/mediator/UnitsHelper.hpp
@@ -11,6 +11,8 @@ class UnitsHelper {
 
     static double get_converted_value(const std::string &in_units, const double &value, const std::string &out_units);
 
+    static double* get_converted_values(const std::string &in_units, double* values, const std::string &out_units, const size_t & count);
+
     private:
     static cv_converter* get_converter(const std::string &in_units, const std::string& out_units, ut_unit*& to, ut_unit*& from);
     // Theoretically thread-safe. //TODO: Test?
--- a/include/core/mediator/UnitsHelper.hpp
+++ b/include/core/mediator/UnitsHelper.hpp
@@ -9,10 +9,10 @@ class UnitsHelper {
 
     public:
 
-    static double get_converted_value(const std::string &in_units, double value, const std::string &out_units);
+    static double get_converted_value(const std::string &in_units, const double &value, const std::string &out_units);
 
     private:
-
+    static cv_converter* get_converter(const std::string &in_units, const std::string& out_units, ut_unit*& to, ut_unit*& from);
     // Theoretically thread-safe. //TODO: Test?
     static ut_system* unit_system;
     static std::once_flag unit_system_inited;
--- a/include/realizations/catchment/Bmi_Multi_Formulation.hpp
+++ b/include/realizations/catchment/Bmi_Multi_Formulation.hpp
@@ -17,6 +17,7 @@
 using namespace std;
 
 class Bmi_Multi_Formulation_Test;
+class Bmi_Cpp_Multi_Array_Test;
 
 namespace realization {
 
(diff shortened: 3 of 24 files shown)


.github/workflows/test_and_validate.yml
```diff
@@ -314,6 +326,10 @@ jobs:
       - uses: ./.github/actions/clean-build
         with:
           build-dir: ${{ steps.submod_build_3.outputs.build-dir }}
+      - uses: ./.github/actions/clean-build
+        with:
+          build-dir: ${{ steps.submod_build_4.outputs.build-dir }}
+
 
   # TODO: fails due to compilation error, at least in large part due to use of POSIX functions not supported on Windows.
```
donaldwj, line 334: If we ever support windows it will probably need it own build system

include/core/mediator/UnitsHelper.hpp
```diff
@@ -9,10 +9,12 @@ class UnitsHelper {
 
     public:
 
-    static double get_converted_value(const std::string &in_units, double value, const std::string &out_units);
+    static double get_converted_value(const std::string &in_units, const double &value, const std::string &out_units);
 
-    private:
+    static double* get_converted_values(const std::string &in_units, double* values, const std::string &out_units, const size_t & count);
```
donaldwj, line 14: Do we want this to work on raw arrays. As defined this will change random memory if 'count' is not correctly set to the size of the array. Also returning a value is misleading as we are mutating the call argument.

include/realizations/catchment/Bmi_Multi_Formulation.hpp
```diff
@@ -483,11 +484,24 @@ namespace realization {
...
+        {
+            std::string output_name = selector.get_variable_name();
+            time_t init_time = selector.get_init_time();
+            long duration_s = selector.get_duration_secs();
+            std::string output_units = selector.get_output_units();
+
+            if (availableData.empty() || availableData.find(output_name) == availableData.end()) {
+                throw runtime_error(get_formulation_type() + " cannot get output values for unknown " + output_name + SOURCE_LOC);
+            }
+            return availableData[output_name]->get_values(CatchmentAggrDataSelector("",output_name, init_time, duration_s, output_units), m);
```
donaldwj, line 502: Why create a new selector, this appears identical to the input selector?

include/utilities/bmi_utilities.hpp
```diff
@@ -0,0 +1,110 @@
...
+        std::vector<T> GetValue(Bmi_Adapter<A>& model, const std::string& name) {
+            //TODO make model const ref
+            int total_mem = model.GetVarNbytes(name);
+            int item_size = model.GetVarItemsize(name);
+            int num_items = total_mem/item_size;
+            //Determine what type we need to cast from
+            std::string type = model.get_analogous_cxx_type(model.GetVarType(name), item_size);
+            
+            //C++ form of malloc
+            void* data = ::operator new(total_mem);
```
donaldwj, line 50: maybe attach smart pointer to this to make sure that memory is not leaked in the case of an exception. We have in general avoided new/delete malloc/free as much as possible.

include/utilities/bmi_utilities.hpp
```diff
@@ -0,0 +1,110 @@
...
+            }
+            else if (type == "unsigned long long" || type == "unsigned long long int"){
+                result = helper::make_vector<T>( (unsigned long long*) data, num_items);
+            }
+            else{
+                throw std::runtime_error("Unable to get value of variable " + name +
+                                " as " + boost::typeindex::type_id<T>().pretty_name() + ": no logic for converting variable type " + type);
+            }
+            //clean up the temporary data
+            ::operator delete(data);
```
donaldwj, line 104: At this position we will get a memory leak on the above exception.

donaldwj (changes requested):
The unit conversion code will leak memory on exceptions. As we don't always treat that as fatal error this needs to be addressed. The other comments are just places where I would like design choices explained.

include/realizations/catchment/Bmi_Multi_Formulation.hpp
hellkite500: @mattw-nws any thoughts, I believe this was one of your additions.

include/utilities/bmi_utilities.hpp
hellkite500: Pretty sure this is a fatal error at this point, but in case it is caught elsewhere, we could `delete` before we `throw` as well.

include/utilities/bmi_utilities.hpp
hellkite500: This is partly due to getting a smart pointer of the correct size which also properly deletes the correct amount of memory.  A `void*` of the correct size is required by BMI to populate the array in `GetValue`.  I recall trying to do this with a smart pointer and having issues with it, but I can't remember off the top of my head exactly what the complication was.  It very well may have been getting it typed as `void*` with the correct number of bytes allocated.

The use of manual allocation/cleanup here is at least isolated to this function, and we can fix the potential leak in the exception throwing.  If we can do this sufficiently with a smart pointer, we can change this.

include/core/mediator/UnitsHelper.hpp
hellkite500: This is just a thin wrapper around the UDUNITS function that works on arrays `cv_convert_doubles(conv, values, count, values);`
```c++
double* UnitsHelper::get_converted_values(const std::string &in_units, double* values, const std::string &out_units, const size_t& count)
{
    if(in_units == out_units){
        return values; // Early-out optimization
    }
    std::call_once(unit_system_inited, init_unit_system);
    ut_unit* to = NULL;
    ut_unit* from = NULL;
    cv_converter* conv = get_converter(in_units, out_units, to, from);

    cv_convert_doubles(conv, values, count, values);
    ut_free(from);
    ut_free(to);
    cv_free(conv);
    return values;
}
```
And returning the value (reference to the mutated input variable) is just syntactic convenience and follows the potentially expected pattern as the regular `get_converted_value` which returns the converted value.  It could be said that NOT returning it is misleading in that context.  I implemented it this way to  be flexible, and it is a somewhat common pattern when working with mutable args behind a pointer.

include/utilities/bmi_utilities.hpp
donaldwj: I would add the delete[] before the throw, no reason to leave a potential leak, even on something assumed to be fatal, if we dont need to.

include/utilities/bmi_utilities.hpp
mattw-nws: Interesting question. Maybe you ran into a problem using `unique_ptr<void>`? Maybe I'll play with this.

mattw-nws (approved):
I'm good with it at this point. @donaldwj see if your change requests have been addressed.

include/utilities/bmi_utilities.hpp
donaldwj: This fixes the memory link anytime there was an error, which was my only concern with the original request.

donaldwj (approved):

PR #432, opened 2022-08-11
Remove deprecated Forcing class dependency from legacy formulations

stcui007:
Removing the old legacy forcing

## Additions

-

## Removals

-

## Changes
include/realizations/catchment/Catchment_Formulation.hpp
include/realizations/catchment/Formulation_Constructors.hpp
include/realizations/catchment/Simple_Lumped_Model_Realization.hpp
include/realizations/catchment/Tshirt_C_Realization.hpp
include/realizations/catchment/Tshirt_Realization.hpp
src/realizations/catchment/Simple_Lumped_Model_Realization.cpp
src/realizations/catchment/Tshirt_C_Realization.cpp
src/realizations/catchment/Tshirt_Realization.cpp
test/realizations/catchments/Bmi_C_Cfe_IT.cpp
test/realizations/catchments/Bmi_C_Formulation_Test.cpp
test/realizations/catchments/Bmi_Cpp_Formulation_Test.cpp
test/realizations/catchments/Bmi_Fortran_Formulation_Test.cpp
test/realizations/catchments/Bmi_Py_Formulation_Test.cpp
test/realizations/catchments/LSTM_Realization_Test.cpp

## Testing

1.

## Notes

-

## Todos

-

## Checklist

1.

- [ ] Linux


src/realizations/catchment/Tshirt_C_Realization.cpp
```diff
@@ -37,13 +39,12 @@ Tshirt_C_Realization::Tshirt_C_Realization(forcing_params forcing_config,
                                            std::vector<double> giuh_ordinates,
                                            tshirt::tshirt_params params,
                                            const vector<double> &nash_storage)
-        : Catchment_Formulation(catchment_id, std::move(std::make_unique<Forcing>(forcing_config)), output_stream), catchment_id(std::move(catchment_id)),
+        //: Catchment_Formulation(catchment_id, std::move(std::make_unique<Forcing>(forcing_config)), output_stream), catchment_id(std::move(catchment_id)),
+        : Catchment_Formulation(catchment_id, std::move(std::make_unique<CsvPerFeatureForcingProvider>(forcing_config)), output_stream), catchment_id(std::move(catchment_id)),
```
mattw-nws, line 43: I'm betting this should not be specific to the CSV reader... where is this constructor used? Maybe this is only in test code and that's okay, but not sure.

src/realizations/catchment/Tshirt_C_Realization.cpp
```diff
@@ -99,15 +100,14 @@ Tshirt_C_Realization::Tshirt_C_Realization(forcing_params forcing_config,
...
 }
 
 Tshirt_C_Realization::Tshirt_C_Realization(
         std::string id,
         forcing_params forcing_config,
         utils::StreamHandler output_stream
-) : Catchment_Formulation(std::move(id), std::move(std::make_unique<Forcing>(forcing_config)), output_stream) {
-    _link_legacy_forcing();
+//) : Catchment_Formulation(std::move(id), std::move(std::make_unique<Forcing>(forcing_config)), output_stream) {
+) : Catchment_Formulation(std::move(id), std::move(std::make_unique<CsvPerFeatureForcingProvider>(forcing_config)), output_stream) {
```
mattw-nws, line 110: Same comment as above... where is this constructor used, and does this need to be non-specific to CsvPerFeature?

src/realizations/catchment/Tshirt_C_Realization.cpp
```diff
@@ -434,7 +432,25 @@ double Tshirt_C_Realization::get_response(time_step_t t_index, time_step_t t_del
...
+    //Checking the time step used is consistent with that provided in forcing data
+    time_t t_delta = this->forcing->record_duration();
+    if (t_delta != t_delta_s) {    //Checking the time step used is consistent with that provided in forcing data
+        throw std::invalid_argument("Getting response using insonsistent time step with provided forcing data");
+    }
+    //Negative t_index is not allowed
+    if (t_index < 0) {
+        throw std::invalid_argument("Getting response of negative time step in Tshirt C Realization is not allowed.");
+    }
+    time_t start_time = this->forcing->get_data_start_time();
```
mattw-nws, line 444: While `time_t` is probably reasonable here, I know in some other places we have used `long` because we may deal with dates < 1970. Is this based on code in another module?

src/realizations/catchment/Tshirt_Realization.cpp
```diff
@@ -3,6 +3,8 @@
 #include "TshirtErrorCodes.h"
 #include "Catchment_Formulation.hpp"
 using namespace realization;
+using data_access::MEAN;
```
mattw-nws, line 6: This one isn't actually used, right? And for `SUM`, I would skip the `using` and just use the namespaced name below in the one call it appears in.

src/realizations/catchment/Tshirt_C_Realization.cpp
```diff
@@ -9,6 +9,8 @@
 #include <boost/algorithm/string/join.hpp>
 
 using namespace realization;
+using data_access::MEAN;
```
mattw-nws, line 12: See comments in `Tshirt_Realization.cpp` on these

test/realizations/catchments/LSTM_Realization_Test.cpp
```diff
@@ -2,7 +2,7 @@
 
 #include "gtest/gtest.h"
 #include "realizations/catchment/LSTM_Realization.hpp"
-#include "forcing/Forcing.h"
+//#include "forcing/Forcing.h"
```
mattw-nws, line 5: Note that these tests are almost certainly not actually being run because of the guard at line 1...I am betting that if we can run these tests they won't succeed with only this change.

include/forcing/AorcForcing.hpp
```diff
@@ -14,6 +14,18 @@
 #define CSDMS_STD_NAME_WIND_V_Y "land_surface_wind__y_component_of_velocity"
 #define NGEN_STD_NAME_SPECIFIC_HUMIDITY "atmosphere_air_water~vapor__relative_saturation" // This is not present in standard names, use this for now... may change!
 
+// Recognized Forcing Value Names (in particular for use when configuring BMI input variables)
+// TODO: perhaps create way to configure a mapping of these to something different
+#define AORC_FIELD_NAME_PRECIP_RATE "precip_rate"
```
mattw-nws, line 19: Interestingly... these defines appear to only be used in test code... though the magic strings appear elsewhere (in this file even)... no action necessary here now, but this is probably something that should be refactored, some way, at some point.

mattw-nws (changes requested):
This is an intitial pass... I haven't checked into the logic for doing the epoch time differentials, for instance. But a few changes I already see.

Also, generally, go ahead and delete commented-out references to `Forcing.h` and `_link_legacy_forcing()`, we can see what it was previously via Git.

include/realizations/catchment/Catchment_Formulation.hpp
```diff
@@ -7,7 +7,8 @@
 #include "Et_Accountable.hpp"
 #include <HY_CatchmentArea.hpp>
 
-#include <Forcing.h> // Remove when _link_legacy_forcing() is removed!
+//#include <Forcing.h> // Remove when _link_legacy_forcing() is removed!
+#include "CsvPerFeatureForcingProvider.hpp"
```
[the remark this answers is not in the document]
stcui007: Some of the time functions, get_data_start_time() and several others, for example are only provided in the following header files, so we need to provide CsvPerFeatureForcingProvider some where. I just tried to use GenericDataProvider.hpp and DataProvider.hpp in its place, both gave build errors.
```
include/forcing/CsvPerFeatureForcingProvider.hpp:    long get_data_start_time() override {
include/forcing/DataProvider.hpp:        virtual long get_data_start_time() = 0;
include/forcing/WrappedForcingProvider.hpp:            return wrapped_provider->get_data_start_time();
include/forcing/WrappedForcingProvider.hpp:        long get_data_start_time() override {
include/forcing/WrappedForcingProvider.hpp:            return wrapped_provider->get_data_start_time();
include/forcing/Forcing.h:    long get_data_start_time() override
include/forcing/NetCDFPerFeatureDataProvider.hpp:        long get_data_start_time()
```

include/realizations/catchment/Formulation_Constructors.hpp
```diff
@@ -72,7 +72,8 @@ namespace realization {
         std::shared_ptr<data_access::GenericDataProvider> fp;
         if (formulation_type == "tshirt"  || formulation_type == "tshirt_c"  || formulation_type == "lstm" // These formulations are still using the legacy interface!
             ){
-            fp = std::make_shared<Forcing>(forcing_config);
+            //fp = std::make_shared<Forcing>(forcing_config);
```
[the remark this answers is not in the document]
stcui007: I'll try to reduce it. `fp` will still need to have a value though, as in the netCDF forcing provider a couple lines below.

src/realizations/catchment/Tshirt_C_Realization.cpp
stcui007: It looks like the class is use by a few testing codes. Not sure which form of the constructors.

src/realizations/catchment/Tshirt_C_Realization.cpp
stcui007: time_t is used all over the places in our codes. In CsvPerFeatureForcingProvider.hpp, time_t is used for epoch time. This is used here for consistency. If you think that might not be enough, we can change that.

src/realizations/catchment/Tshirt_Realization.cpp
stcui007: `Mean `is not used here. Will make the suggested change for SUM.

include/realizations/catchment/Catchment_Formulation.hpp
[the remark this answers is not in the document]
mattw-nws: Confused on this, since as you show above, `get_data_start_time()` is defined in both, and it's not even templated... will try this myself and have a look.

include/realizations/catchment/Formulation_Constructors.hpp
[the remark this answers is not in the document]
mattw-nws: ^ Exactly. What I mean is, remove `tshirt_c` and `tshirt` from the conditional above... they should fall through to the other blocks, depending on what the `forcing_config` says to use. They shouldn't be locked to CsvPerFeature, which is what the code does now.

test/realizations/catchments/LSTM_Realization_Test.cpp
stcui007: At the moment, Ngen framework does not seem to set up to handle LSTM module. So there is actually no test code ever built because of the guard lines.

include/forcing/AorcForcing.hpp
stcui007: Yeah, need to keep in mind.

include/realizations/catchment/Catchment_Formulation.hpp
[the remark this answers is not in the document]
stcui007: It seems to me the `get_data_start_time() `is only accessible through `Forcing, CsvPerFeatureForcingProvider, WrappedForcingProvider, or NetCDFPerFeatureDataProvider`.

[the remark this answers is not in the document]
stcui007: You maybe right. Probably need some code reorganization.

include/realizations/catchment/Formulation_Constructors.hpp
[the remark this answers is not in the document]
stcui007: Removing tshirt and tshirt_c appears to cause the `Formulation_Manager_Test` to fail. There are a bunch of `tshirt` and `tshirt_c` examples there using the old form of forcing explicitly. Do we still need the `Formulation_Manager_Test.cpp`?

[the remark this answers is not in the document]
mattw-nws: Fix the provider string in `Formulation_Manager_Test.cpp`:
```
          "\"forcing\": { "
              "\"provider\": \"legacy\","
              "\"path\": \"./data/forcing/cat-67_2015-12-01 00_00_00_2015-12-30 23_00_00.csv\" "
```

include/realizations/catchment/Formulation_Constructors.hpp
```diff
@@ -18,7 +18,7 @@
 #include "Bmi_Multi_Formulation.hpp"
 #include "Bmi_Py_Formulation.hpp"
 #include <GenericDataProvider.hpp>
-#include "CsvPerFeatureForcingProvider.hpp"
+//#include "CsvPerFeatureForcingProvider.hpp"
```
mattw-nws, line 21: This change is actually what is causing your problem in `Catchment_Formulation.hpp`. This is the one place that `CsvPerFeatureForcingProvider.hpp` is needed, because the instances of the providers are instantiated in this class.

include/realizations/catchment/Catchment_Formulation.hpp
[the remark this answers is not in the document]
mattw-nws: The issue here doesn't seem to have anything to do with `get_data_start_time()`. See the comment in `Formulation_Constructors.hpp`.

include/realizations/catchment/Formulation_Constructors.hpp
mattw-nws: You will need `#include "CsvPerFeatureForcingProvider.hpp"` in files where `CsvPerFeatureForcingProvider` is explicitly constructed... for the moment that means the two tshirt files also (though ideally those dependencies can be removed, per the discussion there).

mattw-nws (commented):
The only problem left has to do with `LSTM_Realization`... either it needs to be fixed, dropped, or this will have to keep `Forcing.h` for that one realization... hold until decided.

mattw-nws:
@stcui007 Can you repackage this as a PR to https://github.com/NOAA-OWP/ngen-forcing please--I'd like to move these efforts over there while there in a higher state of flux/earlier state of development. Just put it in a directory with an appropriate name for now... things in there will probably get moved around as things mature.

stcui007:
I have just pushed my codes to my fork.with the branch bane cui_esmpy

mattw-nws (approved):

src/realizations/catchment/LSTM_Realization.cpp
```diff
@@ -72,14 +39,33 @@ LSTM_Realization::LSTM_Realization(
...
-                           forcing_data.TMP_2maboveground_K, forcing_data.UGRD_10maboveground_meters_per_second, 
-                           forcing_data.VGRD_10maboveground_meters_per_second);
+    //Checking the time step used is consistent with that provided in forcing data
+    time_t t_delta = this->forcing->record_duration();
+    if (t_delta != t_delta_s) {    //Checking the time step used is consistent with that provided in forcing data
+        throw std::invalid_argument("Getting response using insonsistent time step with provided forcing data");
+    }
+    //Negative t_index is not allowed
+    if (t_index < 0) {
+        throw std::invalid_argument("Getting response of negative time step in Tshirt C Realization is not allowed.");
```
hellkite500, line 49: This is LSTM_Realization, not Tshirt C Realization

src/realizations/catchment/LSTM_Realization.cpp
```diff
@@ -72,14 +39,33 @@ LSTM_Realization::LSTM_Realization(
...
+    }
+    time_t start_time = this->forcing->get_data_start_time();
+    time_t stop_time = this->forcing->get_data_stop_time();
+    time_t t_current = start_time + t_index * t_delta_s;
```
hellkite500, line 53: is it guaranteed that data start time will align with simulation start time?  Also, there is no guarantee that `t_delta_s` is the SAME delta at each call to get response.

And I'm not sure I'm tracking time here correctly...

`t_index` should be the `current_time` relative to the simulation driver (or whatever is calling `get_response`.  `t_delta_s` is the amount of seconds the call wants to advance relative to `t_index`, or `current_time`.  Is this just trying to convert that into time into the forcing provider's time reference?

src/realizations/catchment/Tshirt_C_Realization.cpp
hellkite500: I would tend to agree...

src/realizations/catchment/Tshirt_C_Realization.cpp
```diff
@@ -434,7 +430,25 @@ double Tshirt_C_Realization::get_response(time_step_t t_index, time_step_t t_del
...
+    if (t_delta != t_delta_s) {    //Checking the time step used is consistent with that provided in forcing data
+        throw std::invalid_argument("Getting response using insonsistent time step with provided forcing data");
+    }
+    //Negative t_index is not allowed
+    if (t_index < 0) {
+        throw std::invalid_argument("Getting response of negative time step in Tshirt C Realization is not allowed.");
+    }
+    time_t start_time = this->forcing->get_data_start_time();
+    time_t stop_time = this->forcing->get_data_stop_time();
+    time_t t_current = start_time + t_index * t_delta_s;
```
hellkite500, line 444: again, the relationship between `start_time`, `t_index`, and `t_delta_s` is not super clear

src/realizations/catchment/LSTM_Realization.cpp
```diff
@@ -10,55 +10,22 @@ using namespace realization;
...
- * @param output_stream
- * @param catchment_id
- * @param pytorch_model_path
- * @param normalization_path
- * @param initial_state_path
- * @param latitude
- * @param longitude
- * @param area_square_km
- */
-LSTM_Realization::LSTM_Realization(
```
hellkite500, line 45: Also, any reason this constructor got removed?

mattw-nws: It was not used anywhere. It could be re-added with `forcing_params` changed to a `DataProvider`, but it would just be zombie code.

hellkite500: The intent of it was to ensure that if someone wanted to use the realization for something they could create it directly with the paths (it is a convenience  constructor).  It wasn't exactly meant to be used in our own code.

donaldwj: My guess it it was using a forcing config object, not sure why it was removed an not refactored though.

src/realizations/catchment/LSTM_Realization.cpp
donaldwj: as this code is for the generic provider the forcing time step could be anything, it does not necessarily match the model time step or is covenant multiple of the model time step. At least for all existing providers it is not a dynamic value.

src/realizations/catchment/LSTM_Realization.cpp
mattw-nws: Somewhat subjective, but it's unused in the application... this isn't a general purpose library (at least not yet)... so it's arguably goldplating. Just more code to maintain.

donaldwj (changes requested):
Nel's concerns in the LSTM code need to be addressed, otherwise no issues that I see

src/realizations/catchment/LSTM_Realization.cpp
mattw-nws: DP interface does support retrieving arbitrary time slices...however, this legacy code was already fixed to the forcing timestep by virtue of the fact that the old Forcing object kept the timestep and time state--that has shifted out of the DataProvider interface, but this didn't add any new restrictions.

mattw-nws: > Also, there is no guarantee that t_delta_s is the SAME delta at each call to get response.
Perhaps not via the interface, but the rest of the framework is currently locked at 3600. That's going to require way more complex changes than you'll find here to change.

src/realizations/catchment/LSTM_Realization.cpp
hellkite500: Nuke it if you want...we have moved away in large part from the development kit style of implementing realizations, relying instead on BMI for the moment.  I would, however, suggest that that deprecation/removal be done in its own isolated commit so that a history/blame on the file in the future would show a clean way to revive that constructor.

donaldwj (approved):

stcui007:
Will test it out.

PR #443, opened 2022-09-13
Partition optimization

hellkite500:
Significant performance concerns were raised when attempting to use the `partitionGenerator` on a CONUS hydrofabric.  After a cursory look at the code, some significant optimizations are possible by simply reducing the amount of copying that happens, especially when functions are called inside of large loops.  This PR addresses some low hanging fruit for optimizing the `partitionGenerator` by simply using references instead of copies/values.

Preliminary analysis shows these changes decrease runtime by nearly 50% on VPU 6, and now most all the run time in that run is in reading the initial catchment data. (20 seconds of the total 23.6 seconds).

@mattw-nws this should help a bit with your CONUS partitioning.

EDIT:
Was able to run the conus partitioning on my macbook in ~15 minutes (after doing a little more memory efficiency in the geojson reader keep from crashing when I ran out of swap and RAM...I'll push those edits in another PR).

I will also note that to get to that speed, I was using a RELEASE build, not a DEBUG build.  The boost ptree reading is highly sensitive to that!

`cmake -DCMAKE_BUILD_TYPE=Release` makes reading the geojson significantly faster!

## Additions

-

## Removals

-

## Changes

-  Pass as many variables as possible by `const` reference
- Refactor the validation algorithm from a `O(n^2)` to `O(n log n)` implementation.

## Testing

1.

## Notes

- The `Network` could be passed by reference, but not `const` reference due to many nested functions that were not const qualified.

## Todos

-

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows project standards (link if applicable)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

## Testing checklist (automated report can be put here)

1. Tested on VPU 06 

### Target Environment support

- [x] Linux
- [x] MacOs

PR #444, opened 2022-09-15
Geojson optimization

hellkite500:
After looking through #443, it became clear that a large part of the initialization overhead in `FeatureCollection` was due to an excessive amount of copies of strings (and some other data).  This PR utilized `std::move` to reduce the amount of copies of data that go through the system.  Also, I took the opportunity to finish implementing the `JSONProperty` using the `boost::variant` for all data types.

This work when used on the partition generator with the optimization in #443 was able to reduce the runtime of the conus run to about 8 minutes.

I noticed a point of diminishing returns working on this code, and investigated briefely.  Turns out, on my machine, I'll likely never get better than about 8 minutes on the conus run.  It literally takes about 5 minutes just to DESTRUCT the property tree as the stack unrolls.  This is because I was pushing ~56GB RAM out of my 16GB phsical memory, so the swapping was intense.

To test this idea a bit, I replaced all geometries in the `catchment_data.geosjon` with a simple `Point` geometry, instead of a `MultiPolygon`.  I then ran the partitioning using that input, and completed the entire thing in 1 minute and 20 seconds!

## Additions

-

## Removals

-

## Changes

- A lot of resource stealing using `std::move`
- Finish implementing `JSONProperty` using the `boost::variant` that was introduced to utilize `List` parameters in `ngen`

## Testing

1. Ran many partition runs

## Notes

- May want to consider implementing move constructors for `JSONProperty` and the various `Feature` types.

## Todos

- Actually was able to remove one from the code this time!

## Checklist

- [x] Linux
- [x] MacOS

### DIFF
--- a/include/geojson/FeatureBuilder.hpp
+++ b/include/geojson/FeatureBuilder.hpp
@@ -81,10 +81,9 @@ namespace geojson {
                 for (auto &coordinate : group.second) {
                     latLon.push_back(std::stod(coordinate.second.data()));
                 }
-                shape_coordinates.push_back(latLon);
+                shape_coordinates.push_back(std::move(latLon));
             }
-            total_coordinates.push_back(shape_coordinates);
-            
+            total_coordinates.push_back(std::move(shape_coordinates));
         }
 
         polygon_t polygon;
@@ -172,14 +171,12 @@ namespace geojson {
                     for(auto &coordinate : point.second) {
                         latLon.push_back(std::stod(coordinate.second.data()));
                     }
-
-                    coordinate_group.push_back(latLon);
+                    coordinate_group.push_back(std::move(latLon));
                 }
-
-                polygon_coordinates.push_back(coordinate_group);
+                polygon_coordinates.push_back(std::move(coordinate_group));
             }
 
-            coordinates_groups.push_back(polygon_coordinates);
+            coordinates_groups.push_back(std::move(polygon_coordinates));
         }
 
         bg::model::multi_polygon<polygon_t> multi_polygon;
@@ -190,7 +187,7 @@ namespace geojson {
                 for (auto &point : polygon_points) {
                     bg::append(polygon, coordinate_t(point[0], point[1]));
                 }
-                multi_polygon.push_back(polygon);
+                multi_polygon.push_back(std::move(polygon));
             }
         }
 
@@ -198,7 +195,7 @@ namespace geojson {
     }
 
     static geometry build_geometry(boost::property_tree::ptree &tree) {
-        std::string type = tree.get<std::string>("type");
+        std::string type = std::move(tree.get<std::string>("type"));
 
         if (type == "Point") {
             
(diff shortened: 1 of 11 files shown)


hellkite500:
I haven't built all ngen tests locally yet, so I didn't see these failing until the PR went up. I'll dig into these a little bit and see what is happening.

include/geojson/FeatureBuilder.hpp
```diff
@@ -277,7 +274,7 @@ namespace geojson {
                 }
             }
             else if (child.first == "id") {
-                id = child.second.data();
+                id = std::move(child.second.data());
```
christophertubbs, line 277: I _think_ that `std::move` might end up ripping stuff off off `child` here. It's hard to tell here without having example docs in front of me, but it looks like there's a chance that these `std::move` operations might modify the tree and bubble up the updates, harming reuse. There _might_ need to be some comments highlighting this so it's easier to identify while skimming.

hellkite500: Once processed and recorded does the state of that node in the ptree really matter? The reference in the tree should still be valid if I understand the semantics of move correctly, but even if it isn't does it matter in this context?

christophertubbs:
How computationally expensive would it be to wrap the `boost::get` calls in a function that calls `boost::get` in a try-catch block and prints out some of the identifiers that fail? That seems to be the problem with the unit tests and, if it's being run into here, I'd imagine it'd be run into elsewhere. It would be cool if we could get the full path, something like `failed on "root->foo->bar->5->foobar->0"`, but just seeing something like `failed on "foobar"` might provide enough context to find the issue in the parsing.

include/geojson/JSONProperty.hpp
```diff
@@ -25,27 +30,107 @@ namespace geojson {
...
 
     /**
-     * @brief Struct wrapping a map of nested \ref PropertyVariant representing a JSON object.
+     * @brief Struct wrapping a nested \ref PropertyMap representing a JSON object.
      * 
      */
     struct Object{
-        std::map<std::string, PropertyVariant> values;
+        private:
+        PropertyMap* values;
```
donaldwj, line 100: Should this be raw pointer?

donaldwj (changes requested):
My only concern is the addition of a raw pointer to JSONProperty, I think a shared pointer or unique pointer should be usable, depending on if ownership is being taken of the data.

hellkite500:
Finally, tests passing!  I also pushed a couple clean up commits and some additional `JSONProperty` test coverage.

hellkite500:
> My only concern is the addition of a raw pointer to JSONProperty, I think a shared pointer or unique pointer should be usable, depending on if ownership is being taken of the data.

This is really just a view to backing storage that is managed via stl containers in the `JSONProperty` class, but was required to indirectly access the data via the variant and subsequently any visitor (such as the vector visitor for lists) that may need to operate on the data.

This pointer is never allocated to anything, and the only way to set it is via the constructor of the struct by passing the reference it should point to when the struct is created.

donaldwj (approved):

PR #455, opened 2022-10-12
Update INSTALL.md

christophertubbs:
Updated the install documentation to fix some formatting issues.

Copying and pasting commands beforehand would have resulted in some command line issues. This change should fix some of that.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -34,41 +34,38 @@ mkdir ngen && cd ngen
 
 **Download the Boost Libraries:**
 
-`
+```shell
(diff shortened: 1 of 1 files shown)


ZacharyWills:
This could use a bit more explanation of what each step is used for, and how differing installations might work (installing for testing vs installing for running a realization etc.).

If we dont have time for that at the moment I can just push this through.

christophertubbs:
More documentation is generally always better. This isn't in the scope of anything really, though - it's just something I went to edit when some copy and pasting didn't work out. The `git submodule` statements, for example, were rendering on the same line. Turns out git doesn't know what the argument `git` means.

mattw-nws (approved):

PR #462, opened 2022-11-01
Adding snow17 module to nextgen

andywood:
Added the snow17 model as an external routine in nextgen
Created an initial example realization in the data/ directory, using cat-27 data. 

## Testing
Ran within nextgen and created output files with plausible data values & correct formats. 
Development and testing was done on Cheyenne with the gnu fortran compiler

## Notes
The new module can be built using:
* cmake -B extern/snow17/cmake_build -S extern/snow17
* cmake --build extern/snow17/cmake_build --target all

## Todos
* Improve the test case to use a snowy basin setting
* Further testing against standalone in the new snowy test case
* Update unit testing if needed

### DIFF
--- a/extern/iso_c_fortran_bmi/test/test_iso_c.c
+++ b/extern/iso_c_fortran_bmi/test/test_iso_c.c
@@ -121,7 +121,7 @@ int main(int argc, char** argv)
     for(i = 0; i < count; i++){
         names[i] = malloc(sizeof(char)*BMI_MAX_VAR_NAME);
         //names[i] = "Hello World\0";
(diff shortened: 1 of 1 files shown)


madMatchstick (commented):
Workflow outlined in README runs with minor adjustments to `snow17-init-cat-27.namelist.input` and `example_bmi_multi_realization_config_w_snow17.json`.  

Note: We will be including `ngen` build/run instructions in a future PR directly to [snow17 repo](https://github.com/NOAA-OWP/snow17), replacing (closing) this one.

PR #468, opened 2022-11-08
update to pybind 2.10.1

hellkite500:
To support python 3.11, as well as gather some important bug fixes, pybind11 needs to be updated.  At the time of this PR 2.10.1 is the latest stable release.

A note on the changes in `InterperterUtils.hpp`, with this version of pybind, the `module_` declarations had to be modified.  ngen will compile just fine with the old definitions, but at runtime, you end up with this type of error:
```sh
libc++abi: terminating with uncaught exception of type pybind11::type_error: Object of type 'type' is not an instance of 'module_'
```
But updating them to generic `py::object` types seemed to work, and I tested this change on the existing 2.6.1 pybind as well and there were no issues.

I was curious about the members in `Routing_Py_Adapter.hpp`, since it stores `py::module_` handles as well, but it seems like they weren't affected (I ran routing simulations with no runtime errors...).

Thanks to nmslib/nmslib#488 and microsoft/spacy-ann-linker#7 for pointing the way!

## Changes

- update pybind submodule to commit for version 2.10.1
- change `py::module_` types to `py::object` types

## Testing

1. Ran through ngen and ngen+routing locally

## Notes

- This PR alone does not address #370 but some related work is inbound that should resolve that issue.

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows project standards (link if applicable)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

### Target Environment support

- [x] Linux
- [x] MacOS

PR #469, opened 2022-11-08
explicitly import iostream for print message in routing adapter

hellkite500:
A quick fix to ensure that `std::cout` is available in this compilation unit by explicitly `#include <iostream>`.


mattw-nws:
This and #468 create a bit of a chicken-and-egg problem for the automated tests but this is the least complex of any of the related PRs... approving and bypassing branch protection!

mattw-nws (approved):

PR #470, opened 2022-11-08
Fix python segfaults

### LINKED ISSUE #370: Parallel runs with routing crash
When running parallel framework runs with routing enabled,  ngen crashes trying to load the routing module .  This crash occurs after the catchment formulations complete.

## Current behavior

A segmentation fault occurs trying to use pybind for the routing integration in parallel.

## Expected behavior

The parallel formulation execution should finish, and rank 0 should initialize and execute the t-route routing module.

hellkite500:
Under certain conditions (compilers, time of day, user mood...), ngen with python support active would hit a segmentation fault at the end of `main`, during the destruction/cleanup of the program.  This was first observed in #370 and has recently been seen even outside of the parallel runtime.  As noted in that issue, the static singleton used across multiple compilation units is likely to blame, as the the destruction order of resources is not guaranteed.

This PR addresses this problem while maintaining a static global interperter util by using a `weak_ptr` as the static resource, and then constructing `shared_ptr` objects as the singleton `instances`.  The `weak_ptr` ref counting prevents the destruction the of the underlying util class, which contains the `py::scoped_interperter` that defines the lifecycle of the python interperter.  

Note that if an instance is called but not referenced
```c++
utils::ngenPy::InterpreterUtil::getInstance();
```
then the interpreter is started and finalized, as there is no reference to count.  In this case, the next call to `getInstance()` creates and initializes a NEW interperter.

In contrast, if the reference is held
```c++
auto a = utils::ngenPy::InterpreterUtil::getInstance();
```
then the next call to `getInstance()` returns a shared pointer to the existing interpreter.

I also added a reference to the `Routing_Py_Adapter` class to ensure that it always had a valid interpreter available, instead of relying strictly on the assumption that someone created and was holding a reference at the right scope for it to operate.

## Changes

- Change the singleton implementation of `InterperterUtil` to use a ref-counting mechanism.

## Testing

1. Tested with and without routing enabled on a sample ngen run.

## Notes

- This was tested on multiple versions of pybind, namely 2.6.1 and 2.10.1.  This was branched off the #468 branch, and may want to be merged sequentially with that PR, but I don't expect any merge conflicts regardless.

## Todos

- Test this on the setup and example in #370

## Checklist

- [x] Linux
- [x] MacOS

Closes #370

### DIFF
--- a/include/routing/Routing_Py_Adapter.hpp
+++ b/include/routing/Routing_Py_Adapter.hpp
@@ -107,6 +107,14 @@ namespace routing_py_adapter {
 
     private:
 
+
+        /** Handle to the interperter util.
+         * 
+         * Order is important, must be constructed before anything depending on it
+         * and destructed after all dependent members.
+        */
+        std::shared_ptr<utils::ngenPy::InterpreterUtil> interperter;
+
         /** A binding to the Python numpy package/module. */
         py::object np;
 
--- a/include/utilities/python/InterpreterUtil.hpp
+++ b/include/utilities/python/InterpreterUtil.hpp
@@ -30,10 +30,54 @@ namespace utils {
         class InterpreterUtil {
 
         public:
-            static InterpreterUtil &getInstance() {
-                static InterpreterUtil instance;
+            static std::shared_ptr<InterpreterUtil> getInstance() {
+                /**
+                 * @brief Singleton instance of embdedded python interpreter.
+                 * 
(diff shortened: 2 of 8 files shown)


mattw-nws:
Approving based on discussions in standup this morning, seems everyone's in agreement that this makes sense.

mattw-nws (approved):

PR #475, opened 2022-11-15
Fix Partition reading empty list for remote connections

hellkite500:
Fixes bug with parsing empty remote-connection list when running in parallel.

## Additions

- A ptree constructor for `Partitions_Parser` (primarily for testing purposes)

## Removals

-

## Changes

- Check type when dealing with parsed list to ensure it is a list and not empty (which turns into a string when parsed via `boost::ptree`)

## Testing

1. Added unit test to test this condition in remote-connections
2. Disabled the old parser unit tests and added a simple test case to test
3. Enabled partition parser testing even if mpi isn't explicity active, since it has not direct MPI dependency

## Checklist

- [x] Linux
- [x] MacOS

### DIFF
--- a/include/core/Partition_Parser.hpp
+++ b/include/core/Partition_Parser.hpp
@@ -45,6 +45,10 @@ class Partitions_Parser {
             std::cout << "file_path: " << file_path << std::endl;
         };
 
+        Partitions_Parser(const boost::property_tree::ptree tree){
+            this->tree = tree;
+        }
+
         virtual ~Partitions_Parser(){};
 
         //The function that parses the json file and build a unordered set and vector of structs for each line in the json list
--- a/include/core/Partition_Parser.hpp
+++ b/include/core/Partition_Parser.hpp
@@ -93,16 +93,21 @@ class Partitions_Parser {
                 }
                 part_data.nexus_ids = nexus_ids;
                 nexus_ids.clear();
-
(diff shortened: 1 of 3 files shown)


stcui007:
I have applied the patch provided by this PR and built the ngen executable manually. Performing mpirun with this fix using 3 CPUs and a json file of 3 partitions completed the 720 steps without error. Interestingly mpirun with 4 CPUs appears to complete without error. However, increasing the CPU number to 5 and above produces a `Segmentation fault` error. Going above 3 CPUs is just for my own curiosity.

Building unit test (`test_unit`) generates a warning message:
```
ngen/test/utils/Partition_Test.cpp:13: warning: "NGEN_MPI_ACTIVE" redefined
 #define NGEN_MPI_ACTIVE
```

Nevertheless, build process completes successfully and running the test passes all checks. Building `test_all` as a target gives the same behavior. Removing line 13 in Partition_Test.cpp seems to get rid of the warning and unit tests also pass.

Not exactly sure what caused the warning. It may well have come from my build environment or other CMakelist.txt.

Combined with a restriction placed in the partition generator code limiting number of partitions, this should completely resolve the problem in 456. I approve this PR with the option either keep or remove line 13 inPartition_Test.cpp.

stcui007 (approved):

PR #482, opened 2022-12-07
I481 netcdf

stcui007:
Fixing netCDF file closing bug.

## Removals

-

## Changes

modified:   include/core/Partition_Parser.hpp
include/core/catchment/HY_CatchmentRealization.hpp
include/forcing/GenericDataProvider.hpp
include/forcing/NetCDFPerFeatureDataProvider.hpp
include/realizations/catchment/Formulation_Manager.hpp
src/NGen.cpp
test/CMakeLists.txt
test/utils/Partition_Test.cpp

## Testing

1. Run ngen in serial mode
2.  Run ngen in parallel mode

## Notes

-

## Todos

-

## Checklist

1.

- [x] Linux

### DIFF
--- a/data/test_bmi_multi_realization_config_w_netcdf.json
+++ b/data/test_bmi_multi_realization_config_w_netcdf.json
@@ -1,0 +1,154 @@
+{
+    "global": {
+        "formulations": [
+            {
+                "name": "bmi_multi",
+                "params": {
+                    "model_type_name": "bmi_multi_noahowp_cfe",
+                    "forcing_file": "",
+                    "init_config": "",
+                    "allow_exceed_end_time": true,
+                    "main_output_variable": "Q_OUT",
+                    "modules": [
+			{
+			    "name": "bmi_c++",
+			    "params": {
+                                "model_type_name": "bmi_c++_sloth",
+                                "library_file": "./extern/sloth/cmake_build/libslothmodel",
+                                "init_config": "/dev/null",
+                                "allow_exceed_end_time": true,
+                                "main_output_variable": "z",
+                                "uses_forcing_file": false,
+                                "model_params": {
+				    "sloth_ice_fraction_schaake(1,double,m,node)": 0.0,
+				    "sloth_ice_fraction_xinan(1,double,1,node)": 0.0,
+				    "sloth_smp(1,double,1,node)": 0.0
+                                }
+			    }
+                        },
+                        {
+                            "name": "bmi_fortran",
+                            "params": {
+                                "model_type_name": "bmi_fortran_noahowp",
+                                "library_file": "./extern/noah-owp-modular/cmake_build/libsurfacebmi",
+                                "forcing_file": "",
+                                "init_config": "./data/bmi/fortran/noah-owp-modular-init-{{id}}.namelist.input",
+                                "allow_exceed_end_time": true,
+                                "main_output_variable": "QINSUR",
+                                "variables_names_map": {
+                                    "PRCPNONC": "atmosphere_water__liquid_equivalent_precipitation_rate",
+                                    "Q2": "atmosphere_air_water~vapor__relative_saturation",
+                                    "SFCTMP": "land_surface_air__temperature",
+                                    "UU": "land_surface_wind__x_component_of_velocity",
+                                    "VV": "land_surface_wind__y_component_of_velocity",
+                                    "LWDN": "land_surface_radiation~incoming~longwave__energy_flux",
+                                    "SOLDN": "land_surface_radiation~incoming~shortwave__energy_flux",
+                                    "SFCPRS": "land_surface_air__pressure"
+                                },
+                                "uses_forcing_file": false
+                            }
+                        },
+                        {
+                            "name": "bmi_c",
+                            "params": {
+                                "model_type_name": "bmi_c_cfe",
+                                "library_file": "./extern/cfe/cmake_build/libcfebmi",
+                                "forcing_file": "",
+                                "init_config": "./data/bmi/c/cfe/{{id}}_bmi_config.ini",
+                                "allow_exceed_end_time": true,
+                                "main_output_variable": "Q_OUT",
+                                "registration_function": "register_bmi_cfe",
+                                "variables_names_map": {
+                                    "water_potential_evaporation_flux": "ETRAN",
+                                    "atmosphere_air_water~vapor__relative_saturation": "SPFH_2maboveground",
+                                    "land_surface_air__temperature": "TMP_2maboveground",
+                                    "land_surface_wind__x_component_of_velocity": "UGRD_10maboveground",
+                                    "land_surface_wind__y_component_of_velocity": "VGRD_10maboveground",
+                                    "land_surface_radiation~incoming~longwave__energy_flux": "DLWRF_surface",
+                                    "land_surface_radiation~incoming~shortwave__energy_flux": "DSWRF_surface",
+                                    "land_surface_air__pressure": "PRES_surface",
+				    "ice_fraction_schaake" : "sloth_ice_fraction_schaake",
+				    "ice_fraction_xinan" : "sloth_ice_fraction_xinan",
+				    "soil_moisture_profile" : "sloth_smp"
+                                },
+                                "uses_forcing_file": false
+                            }
+                        }
+                    ],
+                    "uses_forcing_file": false
+                }
+            }
+        ],
+        "forcing": {
+            "path": "./data/forcing/cats-27_52_67-2015_12_01-2015_12_30.nc",
+            "provider": "NetCDF"
+        }
+    },
+    "time": {
+        "start_time": "2015-12-01 00:00:00",
+        "end_time": "2015-12-30 23:00:00",
+        "output_interval": 3600
+    },
+    "catchments": {
+        "cat-27": {
+          "formulations": [
+                        {
+                            "name": "bmi_c",
+                            "params": {
+                                "model_type_name": "bmi_c_pet",
+                                "library_file": "./extern/evapotranspiration/evapotranspiration/cmake_build/libpetbmi",
+                                "forcing_file": "",
+                                "init_config": "./data/bmi/c/pet/cat-27_bmi_config.ini",
+                                "allow_exceed_end_time": true,
+                                "main_output_variable": "water_potential_evaporation_flux",
+                                "registration_function": "register_bmi_pet",
+                                "variables_names_map": {
+                                    "water_potential_evaporation_flux": "potential_evapotranspiration"
+                                },
+                                "uses_forcing_file": false
+                            }
+                        }
+            ],
+            "forcing": {
+                "path": "./data/forcing/cats-27_52_67-2015_12_01-2015_12_30.nc",
+                "provider": "NetCDF"
+            }
+        }
+    },
+    "catchments": {
+        "cat-52": {
+          "formulations": [
+                        {
+                            "name": "bmi_c",
+                            "params": {
+                                "model_type_name": "bmi_c_cfe",
+                                "library_file": "./extern/cfe/cmake_build/libcfebmi",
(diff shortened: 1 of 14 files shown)


include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -245,6 +245,18 @@ namespace realization {
                     return "";
             }
 
+   #ifdef NETCDF_ACTIVE
+            void close_nc_file()
+            {
+                for (std::vector<std::shared_ptr<Catchment_Formulation>>::iterator it = missing_formulation_vec.begin(); it != missing_formulation_vec.end(); ++it)
+                {
+                    if (it == missing_formulation_vec.begin()) {
+                        (*it)->get_forcing()->get_nc_file()->close();
```
mattw-nws, line 254: This will blow up if CSV forcing is used, won't it? And CSV forcing can be used when NetCDF support is compiled in, so the compile-time guard above doesn't prevent this.  In fact, a simulation can include some formulations that use NetCDF forcing and some that use CSV... and we're not done adding more.

include/forcing/GenericDataProvider.hpp
```diff
@@ -4,14 +4,26 @@
...
+
 namespace data_access
 {
     class GenericDataProvider : public DataProvider<double, CatchmentAggrDataSelector>
     {
         public:
 
+  #ifdef NETCDF_ACTIVE
+        /** Retrun the shared_ptr to the NcFile */
+        virtual std::shared_ptr<netCDF::NcFile> get_nc_file() {
```
mattw-nws, line 20: We should not have a NetCDF-specific function of any kind in GenericDataProvider.

include/forcing/GenericDataProvider.hpp
```diff
@@ -4,14 +4,26 @@
...
+            return 0;
```
mattw-nws, line 21: If I am understanding this correctly, this will return `0` as a `shared_ptr<netCDF::NcFile>`... At a minimum, this should return `nullptr` instead... and as mentioned below, if you call `->close()` on this later it will crash.

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -318,12 +330,14 @@ namespace realization {
                 forcing_params forcing_config = this->get_global_forcing_params(identifier, simulation_time_config);
 
                 std::shared_ptr<Catchment_Formulation> missing_formulation = construct_formulation(formulation_type_key, identifier, forcing_config, output_stream);
+                missing_formulation_vec.push_back(missing_formulation);
```
mattw-nws, line 333: I am pretty sure `missing_formulation_vec` is completely redundant with `Formulation_Manager::formulations`:
```
std::map<std::string, std::shared_ptr<Catchment_Formulation>> formulations
```
It's just a map and not a vector. You could use the contents of the map easily instead of keeping a second vector to contain them. Also this name is confusing... once they are in this vector, they are not missing anymore.

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -318,12 +330,14 @@ namespace realization {
...
                 // Need to work with a copy, since it is altered in-place
                 geojson::PropertyMap global_properties_copy = global_formulation_parameters;
                 Catchment_Formulation::config_pattern_substitution(global_properties_copy,
                                                                    BMI_REALIZATION_CFG_PARAM_REQ__INIT_CONFIG, "{{id}}",
                                                                    identifier);
                 missing_formulation->create_formulation(global_properties_copy);
+                //missing_formulation->forcing->get_nc_file()->close();
```
mattw-nws, line 340: remove

include/realizations/catchment/Formulation_Manager.hpp
mattw-nws: Also... have you tried this with a realization with more than one formulation? If you have a `global` and also another catchment-specific formulation that uses the same NetCDF forcing file--which is a valid configuration--I am pretty sure this will try to close the file _twice_, which may crash. (You could have any number of catchment formulations using the same NetCDF file, so you could try to close the same one a large number of times.)

stcui007: For the first question, I tested the code yesterday without the guards, that seemed to work ok locally with both netcdf and csv files. But it seemed to not passing some of the tests. So I tried to put some guards on.

For the second question, I have not. I assume we might design a test to see if all netCDF files are closed or somehow to make each pointer specific for an instance and then close each one separately.

include/forcing/GenericDataProvider.hpp
stcui007: Does the guard isolates the get_nc_file() function while still keeps the GenericDataProvider's generality? The reason I had it there is because a GenericDataProvider type of pointer is used in Formulation_Constructor.hpp to point to a NetCDFPerFeatureDataProvider. It seemed this not the case looking at the code. But both gdb debugger and output seem to say so.

include/forcing/GenericDataProvider.hpp
stcui007: Yes, it should return a nullptr.

include/realizations/catchment/Formulation_Manager.hpp
stcui007: As you can see from line 333, std::shared_ptr<Catchment_Formulation> missing_formulation is a pointer and so we are building a vector to keep all the missing_formulation in one place. Alternatively, we can use the formulations map but we have go know the catchment id when we use the map.

include/realizations/catchment/Formulation_Manager.hpp
stcui007: OK.

include/forcing/GenericDataProvider.hpp
mattw-nws: No it doesn't keep the generality, as I think you're finding out when the build fails to find the .h file here... since it's in there at all, now there's a dependency on NetCDF introduced in this class, which has to be resolved in the build system. More generally, this is an abstract base class and shouldn't know anything about the details of its various subclass implementations.

include/realizations/catchment/Formulation_Manager.hpp
mattw-nws: That surprises me... if a formulation uses a CSV forcing, then `GenericDataProvider::get_nc_file()` is going to return 0/nullptr and then `->close()` will be called on a nullptr... if that does not crash, then the NetCDF library authors were very clever in preventing it. It should crash or throw an exception.

stcui007:
Do you have some ideas how to access the "nc_file"
in NetCDFPerFeatureDataProvider.hpp without going through
the GenericDataProvider.hpp?

stcui007:
I think it does crash for CSV forcing input. Maybe this could be fixed.

stcui007:
You are right. We can iterate the map also.

include/realizations/catchment/Formulation_Manager.hpp
stcui007: I have put in a check now so the close operation only performed if the ptr is not a nullptr.

include/forcing/GenericDataProvider.hpp
stcui007: I don't have any good ideas at the moment how to by pass it. Will think about it.

include/realizations/catchment/Formulation_Manager.hpp
stcui007: With the new commit, the code works with multiple formulations.

include/realizations/catchment/Formulation_Manager.hpp
stcui007: The updated version in the new commit uses map. The vector is deleted.

data/test_bmi_multi_realization_config_w_netcdf.json
```diff
@@ -0,0 +1,154 @@
...
+                            }
+                        }
+            ],
+            "forcing": {
+                "path": "./data/forcing/cats-27_52_67-2015_12_01-2015_12_30.nc",
+                "provider": "NetCDF"
+            }
+        }
+    },
+    "catchments": {
```
mattw-nws, line 118: This makes two `"catchments"` keys in the same JSON dictionary, which is invalid... at the least, it seems that the formulation for `cat-52` was never executed--because when you fix this, that formulation fails because `sloth_ice_fraction_schaake` is not found in this formulation.

To fix the invalid JSON, remove lines 117 and 118, and add a comma to the end of line 116. After that the formulation below needs to be fixed... this is somewhat tangential to the actual PR, but this is an invalid test.

stcui007:
You are right. I have fixed the json file and am doing more testing.

stcui007:
My test does show the error that you mentioned. Is it because we don't have
this formulation completely set up? If I remove cat-52 part in the json
file, it does seem to work.

mattw-nws:
> My test does show the error that you mentioned. Is it because we don't have
this formulation completely set up? 

Yes, just like in #460 , CFE can't be in a `bmi_c` anymore, you have to turn it into a `bmi_multi` formulation and either have SLoTH or SMP/SFT provide the extra inputs it needs.

mattw-nws:
> If I remove cat-52 part in the json file, it does seem to work.

Yes, it runs, and finishes, but If you do that, do you not get an error similar to this one? I do...

```
HDF5-DIAG: Error detected in HDF5 (1.12.2) thread 0:
  #000: H5D.c line 326 in H5Dclose(): can't decrement count on dataset ID
    major: Dataset
    minor: Unable to decrement reference count
  #001: H5Iint.c line 1074 in H5I_dec_app_ref_always_close(): can't decrement ID ref count
    major: Object atom
    minor: Unable to decrement reference count
  #002: H5Iint.c line 916 in H5I_remove(): can't remove ID node
    major: Object atom
    minor: Can't delete message
  #003: H5Iint.c line 865 in H5I__remove_common(): can't remove ID node from hash table
    major: Object atom
    minor: Can't delete message
  #004: H5Iint.c line 1018 in H5I_dec_app_ref(): can't decrement ID ref count
    major: Object atom
    minor: Unable to decrement reference count
  #005: H5Iint.c line 949 in H5I_dec_ref(): can't locate ID
    major: Object atom
    minor: Unable to find atom information (already closed?)
NetCDF: HDF error
file: /tmp/netcdf-cxx-20221003-15940-9tllxe/netcdf-cxx4-4.3.1/cxx4/ncFile.cpp  line:33
```

stcui007:
Sorry for the delay in reply. Lynker was having a group lunch.
To your question, no, I don't get any error when I remove cat-52 from the
file.

stcui007:
Will you be able to test it on one of the Linux servers? Such as UCS3 or
others that have /local/lib?

include/forcing/GenericDataProvider.hpp
hellkite500: Ultimately, the issue here is injecting a netcdf specific function into the data provider interface.  So even if we have other provider types supported (e.g. CSV) the interface now suddenly includes a function that isn't relevant...also the macro guard makes the interface confusing in that general sense -- hence why this type of functionality is better suited to be encapsulated by the netcdf provider class.  I'll dig a little deeper into this PR and see if I can help with any ideas.

hellkite500: Seems like @mattw-nws has done the latter part already in #485 .  It isn't quite as explicit, but that solution does two things:
Isolates netcdf specific semantics to the netcdf provider class.
Leverages default destruction semantics to ensure destructors are called appropriately on the netcdf pointers (which in turn trigger destruction/cleanup of the underlying hdf5 resource...).

Ultimately, the problem to solve here is ensure resource management in the right context and the right point in the object life cycle.  C++ gives you a lot of good and bad ways to do that, haha.

mattw-nws:
Resolved via #485

PR #497, opened 2023-03-07
Multi level hydrofabric

donaldwj:

## Additions

- Support for multiple processing levels
- Support for separate models runtimes for each level
-

## Removals

- Old time loop has been replaced
-

## Changes

- Added Layer Objects to Ngen driver
- Layer objects now manage model realizations

## Testing

Manual Test of 5 catchment data set

## Notes

-

## Todos

-

## Checklist

1.

- [ ] Linux


include/core/HY_Features_MPI.hpp
```diff
@@ -29,6 +30,16 @@ namespace hy_features {
             return network.filter("cat");
         }
 
+        inline auto catchments(long lv) {
```
mattw-nws, line 33: Should this unused parameter be removed? It's not present in the non-MPI version.

src/NGen.cpp
```diff
@@ -327,46 +329,109 @@ int main(int argc, char *argv[]) {
...
+        auto& m_data = layer_meta_data.get_layer(keys[i]);
+        double c_value = UnitsHelper::get_converted_value(m_data.time_step_units,m_data.time_step,"s");
+        time_steps.push_back(c_value);
+
+        if ( i == 0 )
         {
-          response *= (catchment_collection->get_feature(id)->get_property("area_sqkm").as_real_number() * 1000000);
+            min_time_step = max_time_step = c_value;
         }
-        //TODO put this somewhere else as well, for now, an implicit assumption is that a modules get_response returns
```
mattw-nws, line 352: Concerning that the web editor is flagging this section as a merge conflict when it was added 2 years ago?? Perhaps the changes are just so large that it's nice and confused.

include/core/Partition_Parser.hpp
```diff
@@ -93,29 +93,35 @@ class Partitions_Parser {
...
+                    
                     for (auto &remote_conn : part.at("remote-connections").as_list())
                     {
-                        remote_mpi_rank = remote_conn.at("mpi-rank").as_natural_number();
-                        remote_nex_id = remote_conn.at("nex-id").as_string();
-                        remote_cat_id = remote_conn.at("cat-id").as_string();
-                        direction = remote_conn.at("cat-direction").as_string();
-                        tmp_tuple = std::make_tuple(remote_mpi_rank, remote_nex_id, remote_cat_id, direction);
-                        remote_conn_vec.push_back(tmp_tuple);
+                        if ( remote_conn.get_type() != geojson::PropertyType::String )
```
mattw-nws, line 107: What problem was this solving? Was this a string in some case?

include/core/Partition_Parser.hpp
```diff
@@ -93,29 +93,35 @@ class Partitions_Parser {
...
-                }
-                part_data.remote_connections = remote_conn_vec;
-                remote_conn_vec.clear();
+                    part_data.remote_connections = remote_conn_vec;
+                    remote_conn_vec.clear();
 
-                //Push part_data struct the vector
-                partition_ranks.push_back(part_data);
+                    //Push part_data struct the vector
+                    partition_ranks.push_back(part_data);
```
mattw-nws, line 121: Now this and the increment below only happens if the partition has a `remote-connections` key--which however may be empty. Was it previously valid to omit the `remote-connections` key entirely? (Just checking)

include/core/HY_Features_MPI.hpp
donaldwj: the non MPI version needs to be updated to have this signature also, Will try to fix that today.

include/core/Partition_Parser.hpp
donaldwj: If I remember it is a string when there are no remote members, but I was not the one to fix this.

include/core/Partition_Parser.hpp
donaldwj: The key has to be present, at least how current parsing works, but it can be empty. We should probably make it fully optional.

src/NGen.cpp
```diff
@@ -327,46 +329,109 @@ int main(int argc, char *argv[]) {
...
+    std::vector<int> output_time_index;
+    output_time_index.resize(keys.size());
+    std::fill(output_time_index.begin(), output_time_index.end(),0);
+
+    //Now loop some time, iterate catchments, do stuff for total number of output times
+    for( int count = 0; count < manager->Simulation_Time_Object->get_total_output_times(); count++) {
+      for ( long lv : features.levels() ) {
+        //std::cout<<"Output Time Index: "<<output_time_index<<std::endl;
+        if(output_time_index[lv]%100 == 0) std::cout<<"Running timestep " << output_time_index[0] <<std::endl;
+          std::string current_timestamp = manager->Simulation_Time_Object->get_timestamp(output_time_index[0]);
```
mattw-nws, line 392: Inadvertent indentation here confused me for a bit... possibly indent 390-391 to match to avoid changing the diff too much for now.

donaldwj: We probably want to update 'Simulation_Time_Object' to expect multiple time layers

mattw-nws: Yes I was just trying to figure out how that fit in... and whether the current assumptions cause any problems.

mattw-nws: I think what probably makes sense is to have a `Simulation_Time` instance per layer... and possibly rename it (prob later refactor). The instance might even be part of the `Layer_Description` class... and it could know the "global" simulation time and figure out the timestep index information--encapsulating that capability there.

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -128,6 +129,53 @@ namespace realization {
...
+                    {
+                        // layer description struct
+                        ngen::LayerDescription layer_desc;
+
+                        // extract and store layer data from the json
+                        layer_desc.name = layer_config.second.get<std::string>("name");
+                        layer_desc.id = layer_config.second.get<int>("id");
+                        layer_desc.time_step = layer_config.second.get<double>("time_step");
+                        boost::optional<std::string> layer_units = layer_config.second.get_optional<std::string>("time_step_units");
+                        if (*layer_units == "") layer_units = "s";
```
hellkite500, line 156: Need to check if the `optional` is valid to dereference first, and also set the default if it is `boost::none`
```c++
if ( (layer_units && *layer_units == "") || !layer_units ) layer_units = "s";
```

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -128,6 +129,53 @@ namespace realization {
...
+                        layer_desc.time_step_units = *layer_units;
+
+                        // check to see if this layer was allready defined
+                        if (layer_storage.exists(layer_desc.id) )
+                        {
+                            std::string message = "A layer with id = ";
+                            message += std::to_string(layer_desc.id);
+                            message += " was defined more than once";
+
+                            std::runtime_error r_error(message);
```
hellkite500, line 166: just `throw std::runtime_error(message)`

hellkite500 (changes requested):
An initial pass through this code shows a few things that need to be addressed, and some potentially larger design an integration discussions/decisions.  I think functionally it will serve the purpose we need, and can likely be lightly refactored to help with the code verbosity and visibility of certain components.  As I write this, one thing that I'm thinking about heavily is the complete abstraction of the feature updates behind the layer interface `update_models` and whether moving that logic that far "out of sight" of the driver/entrypoint is going to be good or if it will make understanding the flow of information and computation difficult to understand.  Just some musings...

Most of the comments here are addressable by themselves, and we can think about the bigger picture things as this progresses.  I definitely want to expand on some testing and test cases as well.

hellkite500:
I also suggest a local update of your branch to fix conflicts:
```sh
git fetch upstream
git rebase upstream/master
#fix conflicts
git add <conflicted file>
git rebase --continue
git push -f origin multi_level_hydrofabric
```

include/core/HY_Features.hpp
```diff
@@ -65,18 +66,7 @@ namespace hy_features {
          * @param network 
          * @param formulations 
          */
-        HY_Features( network::Network network, std::shared_ptr<Formulation_Manager> formulations);
```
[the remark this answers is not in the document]
donaldwj: So far as I could tell it was unused and out of date.

src/core/HY_Features.cpp
```diff
@@ -3,59 +3,61 @@
...
+      formulation->set_output_stream(feat_id+".csv");
+      // TODO: add command line or config option to have this be omitted
+      //FIXME why isn't default param working here??? get_output_header_line() fails.
+      formulation->write_output("Time Step,""Time,"+formulation->get_output_header_line(",")+"\n");
+      //Find upstream nexus ids
+      origins = network.get_origination_ids(feat_id);
+
+      // get the catchment level from the hydro fabric
+      const auto& cat_json_node = fabric->get_feature(feat_id);
+      long lv = cat_json_node->has_key("level") ? cat_json_node->get_property("level").as_natural_number() : 0;
```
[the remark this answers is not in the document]
donaldwj: I think that was left over from a pervious version

include/core/Layer.hpp
```diff
@@ -0,0 +1,147 @@
...
+                //since we are operating on a 1 hour (3600s) dt, we need to scale the output appropriately
+                //so no response is m^2/hr...m^2/hr * 1hr/3600s = m^3/hr
+                response /= 3600.0;
+                //update the nexus with this flow
+                for(auto& nexus : features.destination_nexuses(id)) {
+                    //TODO in a DENDRIDIC network, only one destination nexus per catchment
+                    //If there is more than one, some form of catchment partitioning will be required.
+                    //for now, only contribute to the first one in the list
+                    nexus->add_upstream_flow(response, id, output_time_index);
+                    std::cerr << "Add water to nexus ID = " << nexus->get_id() << " from catchment ID = " << id << " value = "
```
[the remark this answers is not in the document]
donaldwj: The remote connection type issue was discovered in testing with a very small dataset, I think it has been implemented others including you

Not sure if adding logging would be in scope for this

include/core/HY_Features.hpp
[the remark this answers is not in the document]
donaldwj: those changes are old enough now that I can't be sure, I know it was causing problems with getting layer meta data to be stored but that might have been before I updated the network class. It might work again now.

include/core/Layer.hpp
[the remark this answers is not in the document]
hellkite500: Doesn't need to be implemented in this work explicitly, but these debugging prints should be either commented out or wrapped up in one of the QUITE build macros (or a new macro if appropriate).  We are soon going to have to scrub all these in a meaningful way for porting to operations anyways.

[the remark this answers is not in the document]
mattw-nws: Yes, I am planning to make a "milestone" in the new schedule/project for "production logging"... I'd like to get a C++ static API that basically clones Python's logging, and maybe even shunt everything to that if Python is enabled.

include/core/HY_Features_MPI.hpp
mattw-nws: Resolved in b0d7395

mattw-nws:
I rebased and fixed at least one missing addition to `HY_Features.cpp`--just be aware you'll need to rebase or pull your local copy on this now, since I pushed to your branch @donaldwj .

.gitignore
```diff
@@ -97,7 +97,6 @@ CMakeCache.txt
 /cmake-build-debug
 /cmake_build
 /build
-/cmake_build*
```
PhilMiller, line 100: Why is this PR messing with `.gitignore`?

PhilMiller:
I intend to review this, in pursuit of my own code familiarization, and since it's pretty relevant to pieces I expect to be working on.

Would it be possible to elaborate the PR description with something of a "reading order" for how to understand the (relatively large) overall change, or rebase to restructure the commits made to date into a more structured narrative that can be understood commit-by-commit?

test/core/multilayer/MultiLayerParserTest.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+    ASSERT_TRUE(true);
+}
+
+TEST_F(MultiLayerParserTest, TestRead0)
+{
+    manager = std::make_shared<realization::Formulation_Manager>(realization_config_path.c_str());
+    manager->read(nexus_collection, utils::getStdOut());
+
+    ASSERT_TRUE(true);
```
PhilMiller, line 60: So far, this is just a 'smoke test', since it's not asserting anything about the results of its calls to the code being tested.

.gitignore
mattw-nws: Common problem I find myself caught in. Pro tip: https://stackoverflow.com/a/1753078 ... I've started adding my "personal" Git ignore patterns to `.git/info/exclude` instead.

src/NGen.cpp
mattw-nws: The merge conflict seems to have gone away--maybe my rebase fixed this.

src/core/HY_Features_MPI.cpp
```diff
@@ -34,17 +34,28 @@ HY_Features_MPI::HY_Features_MPI( PartitionData partition_data, geojson::GeoJSON
...
           formulation->write_output("Time Step,""Time,"+formulation->get_output_header_line(",")+"\n");
+          
+          // get the catchment level from the hydro fabric
+          const auto& cat_json_node = linked_hydro_fabric->get_feature(feat_id);
+          long lv = cat_json_node->has_key("level") ? cat_json_node->get_property("level").as_natural_number() : 0;
+
+          // add this level to the set of levels if needed
+          if (hf_levels.find(lv) == hf_levels.end() )
+          {
+              hf_levels.insert(lv);
```
mattw-nws, line 53: There's a mix of "layer" and "level" in identifiers here--we need to standardize on "layer" throughout.

include/core/Layer.hpp
```diff
@@ -0,0 +1,147 @@
...
+            if(output_time_index%100 == 0) std::cout<<"Running timestep " << output_time_index <<std::endl;
+            std::string current_timestamp = simulation_time.get_timestamp(output_time_index);
+            for(const auto& id : processing_units) 
+            {
+                int sub_time = output_time_index;
+                //std::cout<<"Running cat "<<id<<std::endl;
+                auto r = features.catchment_at(id);
+                //TODO redesign to avoid this cast
+                auto r_c = std::dynamic_pointer_cast<realization::Catchment_Formulation>(r);
+                r_c->set_et_params(pdm_et_data);
```
mattw-nws, line 99: Here's the root of the dependency on Pdm03 ... I don't believe it makes sense to propagate this into an architecture piece like this. What happens if `set_et_params` is never set? I would like to make the system throw in such a case, per #562, and avoid propagating what is now a legacy capability, if it can be done at all cleanly.

mattw-nws (changes requested):

src/NGen.cpp
```diff
@@ -327,74 +331,99 @@ int main(int argc, char *argv[]) {
...
+    for(int i = 0; i < keys.size(); ++i)
+    {
+        auto& m_data = layer_meta_data.get_layer(keys[i]);
+        double c_value = UnitsHelper::get_converted_value(m_data.time_step_units,m_data.time_step,"s");
+        time_steps.push_back(c_value);
+    }
+
+    std::vector<int> output_time_index;
+    output_time_index.resize(keys.size());
+    std::fill(output_time_index.begin(), output_time_index.end(),0);
```
PhilMiller, line 353:
```suggestion
    std::vector<int> output_time_index(keys.size(), 0);
```

src/NGen.cpp
```diff
@@ -327,74 +331,99 @@ int main(int argc, char *argv[]) {
...
+
+    // now create the layer objects
+
+    // first make sure that the layer are listede in decreasing order
```
PhilMiller, line 357:
```suggestion
    // first make sure that the layer are listed in decreasing order
```

src/NGen.cpp
```diff
@@ -327,74 +331,99 @@ int main(int argc, char *argv[]) {
...
+
+    std::vector<std::shared_ptr<ngen::Layer> > layers;
+    layers.resize(keys.size());
+
+    for(long i = 0; i < keys.size(); ++i)
+    {
+      auto& desc = layer_meta_data.get_layer(keys[i]);
+      std::vector<std::string> cat_ids;
+
+      // make a new simulation time object with a differnt output interval
```
PhilMiller, line 368:
```suggestion
      // make a new simulation time object with a different output interval
```

src/NGen.cpp
```diff
@@ -327,74 +331,99 @@ int main(int argc, char *argv[]) {
...
-                             r_c->get_output_line_for_timestep(output_time_index)+"\n";
-        r_c->write_output(output);
-        //TODO put this somewhere else.  For now, just trying to ensure we get m^3/s into nexus output
-        try{
-          response *= (catchment_collection->get_feature(id)->get_property("areasqkm").as_real_number() * 1000000);
-        }catch(std::invalid_argument &e)
+    for( int count = 0; count < manager->Simulation_Time_Object->get_total_output_times(); count++) 
+    {
+      // The Inner loop will advance all layers unless doing so will break one of two constraints
+      // 1) A layer may not preceed ahead of the master simulation objects current time
```
PhilMiller, line 387:
```suggestion
      // 1) A layer may not proceed ahead of the master simulation object's current time
```

src/NGen.cpp
```diff
@@ -327,74 +331,99 @@ int main(int argc, char *argv[]) {
...
+      // 2) A layer may not preceed ahead of any layer that is computed before it
+      // The do while loop insures that all layers are tested at least once while allowing 
+      // layers small time steps to be updated more than once
```
PhilMiller, line 390:
```suggestion
      // layers with small time steps to be updated more than once
```
or
```suggestion
      // layers' small time steps to be updated more than once
```
?

PhilMiller:
Sorry for all the spelling/punctuation fixes I'm marking, but they're all in spots where it took me longer to understand because of them. Hence, I think it will make a difference to future development velocity to make sure they're fixed, so no one else pays that comprehension cost.

src/NGen.cpp
```diff
@@ -327,74 +331,99 @@ int main(int argc, char *argv[]) {
...
+    // check the time loops for the existing layers
+    ngen::LayerDataStorage& layer_meta_data = manager->get_layer_metadata();
+
+    // get the keys for the existing layers
+    std::vector<int>& keys = layer_meta_data.get_keys();
+
+    // get the converted time steps for layers
+    double min_time_step;
+    double max_time_step;
+    std::vector<double> time_steps;
```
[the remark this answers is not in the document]
PhilMiller:
```suggestion
    std::vector<double> time_steps;
    time_steps.reserve(keys.size());
```
This will allocate the memory in one go, and avoid initializing it just to then overwrite it below

include/core/Partition_Parser.hpp
```diff
@@ -93,29 +93,37 @@ class Partitions_Parser {
                 }
                 part_data.nexus_ids = nexus_ids;
                 nexus_ids.clear();
-                if( part.at("remote-connections").get_type() == geojson::PropertyType::List ) {
+
+                if( part.at("remote-connections").get_type() == geojson::PropertyType::List ) 
+                {
                     //It is valid to have no remote connections, but the backend property tree parser
                     //can't represent empty lists/objects, so it turns into an ampty string (which is iterable)
```
PhilMiller, line 100:
```suggestion
                    //can't represent empty lists/objects, so it turns into an empty string (which is iterable)
```

include/core/SurfaceLayer.hpp
```diff
@@ -0,0 +1,87 @@
...
+                if( cat_ids.size() > 0 ) {
+                //Assumes dendridic, e.g. only a single downstream...it will consume 100%  of the available flow
+                cat_id = cat_ids[0];
+                }
+                else {
+                //This is a terminal node, SHOULDN'T be remote, so ID shouldn't matter too much
+                cat_id = "terminal";
+                }
+
+                std::cerr << "Requesting water from nexus, id = " << id << " at time = " <<current_time_index << ",  percent = 100, destination = " << cat_id << std::endl;
```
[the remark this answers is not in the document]
mattw-nws: Is this WRT future #574? Or are you suggesting using `cout` instead? Or wrap in `#ifndef NGEN_QUIET`? Or are you making a hydrology joke?

include/core/SurfaceLayer.hpp
```diff
@@ -0,0 +1,87 @@
...
+                    nexus_ids(n_u), 
+                    nexus_outfiles(output_files)
+        {
+
+        }
+
+        /***
+         * @breif Run one simulation timestep for each model in this layer
+        */
+        virtual void update_models()
```
PhilMiller, line 32: The function implementations belong in a separate .cpp file. We shouldn't blow up ngen.cpp compilation time by having them here. It won't even help optimization, because virtual dispatch will impede inlining or other inter-procedural analysis/transformation anyway.

src/NGen.cpp
```diff
@@ -327,74 +331,99 @@ int main(int argc, char *argv[]) {
...
-        //TODO put this somewhere else.  For now, just trying to ensure we get m^3/s into nexus output
-        try{
-          response *= (catchment_collection->get_feature(id)->get_property("areasqkm").as_real_number() * 1000000);
-        }catch(std::invalid_argument &e)
+    for( int count = 0; count < manager->Simulation_Time_Object->get_total_output_times(); count++) 
+    {
+      // The Inner loop will advance all layers unless doing so will break one of two constraints
+      // 1) A layer may not preceed ahead of the master simulation objects current time
+      // 2) A layer may not preceed ahead of any layer that is computed before it
+      // The do while loop insures that all layers are tested at least once while allowing
```
mattw-nws, line 389:
```suggestion
      // The do while loop ensures that all layers are tested at least once while allowing 
```
@PhilMiller #bandwagon

include/core/Layer.hpp
```diff
@@ -0,0 +1,147 @@
...
+{
+
+    class Layer
+    {    
+        public:
+
+        #ifdef NGEN_MPI_ACTIVE
+            typedef hy_features::HY_Features_MPI feature_type;
+        #else
+            typedef hy_features::HY_Features feature_type;
```
PhilMiller, line 25:
```suggestion
            using feature_type = hy_features::HY_Features_MPI;
        #else
            using feature_type = hy_features::HY_Features;
```
There's no good reason to keep using the mixed-order C-style typedefs in modern C++ code.

include/core/LayerData.hpp
```diff
@@ -0,0 +1,59 @@
...
+namespace ngen
+{
+    /**
+     * @brief A simple structure to hold meta data realated to a computational layer
+    */
+    
+    struct LayerDescription
+    {
+        std::string name;             //> The name of the computational layer
+        std::string time_step_units;  //> The time units that time_step is expressed in
```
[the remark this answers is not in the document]
PhilMiller: And to spare costs of copying and comparing these things.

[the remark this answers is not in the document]
PhilMiller: Oh, actually, perhaps `time_step` and `time_step_units` should be combined in a single `std::chrono::duration`. That's available from C++11 onward
https://en.cppreference.com/w/cpp/chrono/duration

include/core/LayerData.hpp
```diff
@@ -0,0 +1,59 @@
...
+        int id;                       //> The layer id which determins the order that this layer will be processed in
```
PhilMiller, line 18:
```suggestion
        int id;                       //> The layer id which determines the order that this layer will be processed in
```

include/core/LayerData.hpp
```diff
@@ -0,0 +1,59 @@
...
+        double time_step;             //> The time step that will be used for models in this layer
+    };
+    
+    /**
+     * @brief A class to hold a manage layer metadata for the system
```
PhilMiller, line 23:
```suggestion
     * @brief A class to hold and manage layer metadata for the system
```
Maybe what was intended?

include/core/LayerData.hpp
```diff
@@ -0,0 +1,59 @@
...
+        std::vector<int>& get_keys() { return keys; }
+
+        private:
+
+        std::unordered_map<int,LayerDescription> stored_layers;
+        std::vector<int> keys;
+    };
+}
+
+#endif
```
PhilMiller, line 59: Missing EOF newline

include/core/Layer.hpp
```diff
@@ -0,0 +1,147 @@
...
+        /***
+         * @brief Return the next timestep that will be processed by this layer in epoc time units
+        */
+        time_t next_timestep_epoch_time() { return simulation_time.next_timestep_epoch_time(); }
+
+
+        /***
+         * @brief Return the last timesteo that was processed by this layer in epoc time units
+        */
+        time_t current_timestep_epoch_time() { return simulation_time.get_current_epoc_time(); }
```
PhilMiller, line 58:
```suggestion
        time_t current_timestep_epoch_time() { return simulation_time.get_current_epoch_time(); }
```
Or is there really a typo in the function name? If so, please fix that too!

include/core/Layer.hpp
```diff
@@ -0,0 +1,147 @@
...
+        */
+        const double get_time_step() const { return this->description.time_step; }
+
+        /***
+         * @brief Return the units for the time_step value of this layer
+        */
+        const std::string& get_time_step_units() const { return this->description.time_step_units; }
+
+        /***
+         * @breif Run one simulation timestep for each model in this layer
```
PhilMiller, line 82:
```suggestion
         * @brief Run one simulation timestep for each model in this layer
```
Doxygen will notice that difference

include/core/Layer.hpp
```diff
@@ -0,0 +1,147 @@
...
+                std::string output = std::to_string(output_time_index)+","+current_timestamp+","+
+                                    r_c->get_output_line_for_timestep(output_time_index)+"\n";
+                r_c->write_output(output);
+                //TODO put this somewhere else.  For now, just trying to ensure we get m^3/s into nexus output
+                try{
+                    response *= (catchment_data->get_feature(id)->get_property("areasqkm").as_real_number() * 1000000);
+                }catch(std::invalid_argument &e)
+                {
+                    response *= (catchment_data->get_feature(id)->get_property("area_sqkm").as_real_number() * 1000000);
+                }
```
PhilMiller, line 110: It would be nice if the actual variation between the two arms (the key name) were isolated from the repeated logic.

Perhaps this whole thing should actually be encapsulated inside a member `catchment_data::get_area()` (cf [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter#In_object-oriented_programming), and then it can be made to cache it so we don't incur repeated lookup costs

PhilMiller: The performance concern holds doubly so because `catchment_data` is a (should-be-`const`) member of this class, and so the area can be looked up once outside any timestep logic.

include/core/Layer.hpp
```diff
@@ -0,0 +1,147 @@
...
+
+        protected:
+
+        LayerDescription description;
+        std::vector<std::string> processing_units;
+        Simulation_Time simulation_time;
+        feature_type& features;
+        geojson::GeoJSON catchment_data;
+        std::shared_ptr<pdm03_struct> pdm_et_data;
+        long output_time_index;
```
PhilMiller, line 142: Whichever of these members can be `const` ought to be.

include/core/Layer.hpp
```diff
@@ -0,0 +1,147 @@
...
+                                    r_c->get_output_line_for_timestep(output_time_index)+"\n";
+                r_c->write_output(output);
+                //TODO put this somewhere else.  For now, just trying to ensure we get m^3/s into nexus output
+                try{
+                    response *= (catchment_data->get_feature(id)->get_property("areasqkm").as_real_number() * 1000000);
+                }catch(std::invalid_argument &e)
+                {
+                    response *= (catchment_data->get_feature(id)->get_property("area_sqkm").as_real_number() * 1000000);
+                }
+                //TODO put this somewhere else as well, for now, an implicit assumption is that a modules get_response returns
```
PhilMiller, line 111:
```suggestion
                //TODO put this somewhere else as well, for now, an implicit assumption is that a module's get_response returns
```

include/core/Layer.hpp
```diff
@@ -0,0 +1,147 @@
...
+                //m/timestep
```
PhilMiller, line 112: Please use different temporary variables to reflect each step of dimension conversion and scaling, so that any one variable can be understood has having fixed dimension and units. If they're named accordingly, then comments like this will be much less load-bearing for someone like me trying to understand what the code is doing.

include/core/SurfaceLayer.hpp
```diff
@@ -0,0 +1,87 @@
...
+                std::unordered_map<std::string, std::ofstream>& output_files) : 
+                    Layer(desc,p_u,s_t,f,cd,pdmd,idx), 
+                    nexus_ids(n_u), 
+                    nexus_outfiles(output_files)
+        {
+
+        }
+
+        /***
+         * @breif Run one simulation timestep for each model in this layer
```
PhilMiller, line 30:
```suggestion
         * @brief Run one simulation timestep for each model in this layer
```

include/core/SurfaceLayer.hpp
```diff
@@ -0,0 +1,87 @@
...
+            long current_time_index = output_time_index;
+            
+            Layer::update_models();
+
+            //At this point, could make an internal routing pass, extracting flows from nexuses and routing
+            //across the flowpath to the next nexus.
+            //Once everything is updated for this timestep, dump the nexus output
+            for(const auto& id : features.nexuses()) 
+            {
+                std::string current_timestamp = simulation_time.get_timestamp(current_time_index);
```
PhilMiller, line 43:
```suggestion
                if(!nexus_outfiles[id].is_open()) continue;

                std::string current_timestamp = simulation_time.get_timestamp(current_time_index);
```

Is `SurfaceLayer::update_models()` ever expected to do actual work, or is it purely here to handle the necessary output? In the latter case, we can omit all the other data juggling up front.

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -129,6 +130,62 @@ namespace realization {
...
+                        // extract and store layer data from the json
+                        layer_desc.name = layer_config.second.get<std::string>("name");
+                        layer_desc.id = layer_config.second.get<int>("id");
+                        layer_desc.time_step = layer_config.second.get<double>("time_step");
+                        boost::optional<std::string> layer_units = layer_config.second.get_optional<std::string>("time_step_units");
+                        if (*layer_units == "") layer_units = "s";
+                        layer_desc.time_step_units = *layer_units;
+
+                        // check to see if this layer was allready defined
+                        if (layer_storage.exists(layer_desc.id) )
```
PhilMiller, line 170: Why isn't this check inside `put_layer`?

include/simulation_time/Simulation_Time.hpp
```diff
@@ -86,12 +93,23 @@ class Simulation_Time
         return output_interval_seconds;
     }
 
+    /**
+     * @brief Accessor to the the current simulation time
+     * @return current_date_time_epoch
+    */
+
+    time_t get_current_epoc_time()
```
PhilMiller, line 101:
```suggestion
    time_t get_current_epoch_time()
```

include/simulation_time/Simulation_Time.hpp
```diff
@@ -68,6 +69,12 @@ class Simulation_Time
         total_output_times = simulation_total_time_seconds / output_interval_seconds + 1;
     }
 
+    Simulation_Time(const Simulation_Time& t, int interval) : Simulation_Time(t)
```
PhilMiller, line 72: This needs documentation of what it's doing. It looks like it's a copy constructor with an over-ridden output interval?

I would prefer if the logic of calculating `total_output_times` weren't duplicated from the other constructor.

More broadly, it seems like output interval shouldn't actually be a member of this at all, but I don't know the code well enough to be confident asserting that

include/simulation_time/Simulation_Time.hpp
```diff
@@ -109,6 +127,42 @@ class Simulation_Time
...
+    {
+        return int(epoch_time_seconds - start_date_time_epoch) / output_interval_seconds;
+    }
+
+    inline int next_timestep_index()
+    {
+        return next_timestep_index(current_date_time_epoch);
+    }
+
+    inline time_t next_timestep_epoch_time(int epoch_time_seconds){
```
PhilMiller, line 140: `time_t` wasn't previously part of the interface of `Simulation_Time` other than what's passed in through the `simulation_time_params` constructor. Could we use the better specified `std::chrono` types and routines for this instead?

PhilMiller: Arguments that name a duration with a unit should just be `std::chrono::duration`, too

src/core/HY_Features_MPI.cpp
```diff
@@ -34,17 +34,28 @@ HY_Features_MPI::HY_Features_MPI( PartitionData partition_data, geojson::GeoJSON
...
           //Find and prepare formulation
           auto formulation = formulations->get_formulation(feat_id);
           formulation->set_output_stream(feat_id+".csv");
           // TODO: add command line or config option to have this be omitted
           //FIXME why isn't default param working here??? get_output_header_line() fails.
           formulation->write_output("Time Step,""Time,"+formulation->get_output_header_line(",")+"\n");
+          
+          // get the catchment level from the hydro fabric
+          const auto& cat_json_node = linked_hydro_fabric->get_feature(feat_id);
+          long lv = cat_json_node->has_key("level") ? cat_json_node->get_property("level").as_natural_number() : 0;
```
PhilMiller, line 48: The degree of duplication between `HY_Features` and `HY_Features_MPI` is worrisome

PhilMiller: This might be ripe for refactoring.

Or, I can take the excuse to write a replacement for `HY_Features_MPI`

include/simulation_time/Simulation_Time.hpp
```diff
@@ -86,12 +93,23 @@ class Simulation_Time
...
 
+    /**
+     * @brief Accessor to the the current simulation time
+     * @return current_date_time_epoch
+    */
+
+    time_t get_current_epoc_time()
+    {
+        return current_date_time_epoch;
+    }
```
PhilMiller, line 104: Please mark all methods that ought to be `const` as such, to ease understanding and guard against mistakes.

src/NGen.cpp
```diff
@@ -327,74 +331,99 @@ int main(int argc, char *argv[]) {
...
+      // 1) A layer may not proceed ahead of the master simulation object's current time
+      // 2) A layer may not proceed ahead of any layer that is computed before it
+      // The do while loop ensures that all layers are tested at least once while allowing 
+      // layers with small time steps to be updated more than once
+      // If a layer with a large time step is after a layer with a small time step the
+      // layer with the large time step will wait for multiple timesteps from the preceeding
+      // layer.
+      
+      // this is the time that layers are trying to reach (or get as close as possible)
+      auto next_time = manager->Simulation_Time_Object->next_timestep_epoch_time();
```
PhilMiller, line 396: It looks like `manager->Simulation_Time_Object` never has its `current_date_time_epoch` advanced in the revised code. That was (ludicrously) happening inside `get_timestamp`, which is no longer being called.

Is the current testing of this PR sufficient to see whether time is actually advancing appropriately, and the simulation state remains consistent with it?

src/NGen.cpp
```diff
@@ -327,74 +331,99 @@ int main(int argc, char *argv[]) {
...
-        r_c->set_et_params(pdm_et_data);
-        double response = r_c->get_response(output_time_index, 3600.0);
-        std::string output = std::to_string(output_time_index)+","+current_timestamp+","+
-                             r_c->get_output_line_for_timestep(output_time_index)+"\n";
-        r_c->write_output(output);
-        //TODO put this somewhere else.  For now, just trying to ensure we get m^3/s into nexus output
-        try{
-          response *= (catchment_collection->get_feature(id)->get_property("areasqkm").as_real_number() * 1000000);
-        }catch(std::invalid_argument &e)
+    for( int count = 0; count < manager->Simulation_Time_Object->get_total_output_times(); count++)
```
PhilMiller, line 384: It looks like `manager->Simulation_Time_Object` never has its `current_date_time_epoch` advanced in the revised code. That was (ludicrously) happening inside `get_timestamp`, which is no longer being called.

Is the current testing of this PR sufficient to see whether time is actually advancing appropriately, and the simulation state remains consistent with it?

PhilMiller:
I've gotten as much as I think I'm going to out of this PR in its present state. I think my last comment about advancing the simulation time stamp is of greatest concern.

include/core/HY_Features.hpp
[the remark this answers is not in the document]
hellkite500: Please add this back in if possible...

include/core/Layer.hpp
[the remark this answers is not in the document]
hellkite500: If this is debug information only, comment it out or use one of the QUITE macros around it (or even add a new DEBUG macro if that makes the most sense...)

include/core/LayerData.hpp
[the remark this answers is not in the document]
hellkite500: Will put this in an issue for refinement

include/core/SurfaceLayer.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+#include "Layer.hpp"
+
+namespace ngen
+{
+    class SurfaceLayer : public Layer
+    {
+        public:
+        
+        SurfaceLayer(
```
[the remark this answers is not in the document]
hellkite500: add a doc string for this constructor to describe the required inputs

include/core/SurfaceLayer.hpp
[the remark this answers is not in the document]
hellkite500: Comment this out for now or wrap it in a macro that we can disable

include/core/catchment/HY_Catchment.hpp
```diff
@@ -30,10 +30,40 @@ class HY_Catchment : public HY_HydroFeature
     conjoint_catchment({}),
     upper_catchment({}),
     lower_catchment({}),
-    realization(realization){}
+    realization(realization),
+    level(lv){}
     virtual ~HY_Catchment();
     const Nexuses& get_outflow_nexuses(){ return outflows; }
     std::shared_ptr<HY_CatchmentRealization> realization;
+
```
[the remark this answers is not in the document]
hellkite500: This can be raised in an issue for refinement, but should not be taken lightly...

include/core/network.hpp
```diff
@@ -206,12 +209,53 @@ namespace network {
...
+         *    std::cout << catchment;
+         * }
+         * @endcode
+         * 
+         * @param type The type of feature to filter for, i.e. 'cat', 'nex'
+         * @param target_level The level that filtered results should be in
+         * @param order What order to return results in
+         * @return auto 
+         */
+        auto filter(std::string type, int target_level, SortOrder order = SortOrder::Topological)
```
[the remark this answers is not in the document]
hellkite500: Has this been attempted?  If it works reducing this code complexity is worth while.

include/realizations/catchment/Formulation_Manager.hpp
hellkite500: This is a bug that needs to get fixed before we merge!

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -128,6 +129,53 @@ namespace realization {
...
+
+                            std::runtime_error r_error(message);
+                            
+                            throw r_error;
+                        }
+
+                        // add the layer to storage
+                        layer_storage.put_layer(layer_desc, layer_desc.id);
+
+                        // debuggin print to see parsed data
```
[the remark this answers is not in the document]
hellkite500: Remove or wrap with macro to disable

test/core/multilayer/MultiLayerParserTest.cpp
```diff
@@ -0,0 +1,61 @@
+
+#include "gtest/gtest.h"
+#include <Formulation_Manager.hpp>
+
+class MultiLayerParserTest : public ::testing::Test {
+    protected:
+
+        MultiLayerParserTest() : 
+            nexus_data_path("../data/multilayer/nexus_data.geojson"),
```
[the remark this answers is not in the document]
donaldwj: I think it existed only in my old working folder might be able to find it still, It could be gone if I removed the old directory after moving to local storage for the builds

src/NGen.cpp
donaldwj: This should probably be looked at with a multi time interval test. I dont think the standard tests include this.

include/realizations/catchment/Formulation_Manager.hpp
hellkite500: This likely works cause `get_optional<std::string>` will return an "empty object" of type `std::string` if the key doesn't exist...so `layer_units` is guaranteed to be initialized to a string, and the `""` if not key is present...there may still be some unthought of edge case that would make `layer_units` uninitialized...

include/core/SurfaceLayer.hpp
donaldwj: This has been done

include/core/Layer.hpp
donaldwj: old habits ...

donaldwj: Nevermind fixed

include/core/LayerData.hpp
donaldwj: done

include/core/LayerData.hpp
donaldwj: you are correct

include/core/LayerData.hpp
```diff
@@ -0,0 +1,59 @@
...
+        { 
+            return stored_layers.find(id) != stored_layers.end(); 
+        }
+
+        std::vector<int>& get_keys() { return keys; }
+
+        private:
+
+        std::unordered_map<int,LayerDescription> stored_layers;
+        std::vector<int> keys;
```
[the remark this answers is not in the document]
donaldwj: The map will be accessed a lot. Unorderedmap is O(1) vs O(lg(N)) of map, particularly when there will be a small number of keys. The use of vector is also not ideal because not all layer slots will be filled.

Currently layers range in id from -20 (lowest ground water) to 80 (upper atmosphere) but few ids will be used at any time. This means allowing retrieving a layer with its id requires a map (or an index array), since the list of keys is both small and will be needed by the main execution loop there is no reason not to build the key list while layers are added, and then store it for reference.

include/core/Layer.hpp
donaldwj: You can ask Matt on that one he wrote the simulation time class (corrected the typo)

include/core/Layer.hpp
donaldwj: Not many of them can be constant although in most cases the accessors could be. Catchment_data probably could be but I would have to check to see if the GeoJSON class works correctly with a const pointer.

include/realizations/catchment/Formulation_Manager.hpp
donaldwj: because it was a void function that does not throw

donaldwj: also you might want to place a new layer at some id, I this case we don't but there is the possibility that layers would be removed in some sort of transition for a highres pass.

include/core/SurfaceLayer.hpp
```diff
@@ -0,0 +1,87 @@
...
+                //Assumes dendridic, e.g. only a single downstream...it will consume 100%  of the available flow
+                cat_id = cat_ids[0];
+                }
+                else {
+                //This is a terminal node, SHOULDN'T be remote, so ID shouldn't matter too much
+                cat_id = "terminal";
+                }
+
+                std::cerr << "Requesting water from nexus, id = " << id << " at time = " <<current_time_index << ",  percent = 100, destination = " << cat_id << std::endl;
+                double contribution_at_t = features.nexus_at(id)->get_downstream_flow(cat_id, current_time_index, 100.0);
```
PhilMiller, line 63:
```suggestion
                // The call to `get_downstream_flow` updates the accounting of water flow out of the queried nexus - should be named something more like `withdraw_water`
                double contribution_at_t = features.nexus_at(id)->get_downstream_flow(cat_id, current_time_index, 100.0);
```

include/core/SurfaceLayer.hpp
```diff
@@ -0,0 +1,87 @@
...
+                #ifdef NGEN_MPI_ACTIVE
+                if (!features.is_remote_sender_nexus(id)) { //Ensures only one side of the dual sided remote nexus actually doing this...
+                #endif
+
+                //Get the correct "requesting" id for downstream_flow
+                const auto& nexus = features.nexus_at(id);
+                const auto& cat_ids = nexus->get_receiving_catchments();
+                std::string cat_id;
+                if( cat_ids.size() > 0 ) {
+                //Assumes dendridic, e.g. only a single downstream...it will consume 100%  of the available flow
```
PhilMiller, line 54:
```suggestion
                //Assumes dendritic, e.g. only a single downstream...it will consume 100%  of the available flow
```

include/core/Layer.hpp
```diff
@@ -0,0 +1,147 @@
...
+                //since we are operating on a 1 hour (3600s) dt, we need to scale the output appropriately
+                //so no response is m^2/hr...m^2/hr * 1hr/3600s = m^3/hr
+                response /= 3600.0;
+                //update the nexus with this flow
+                for(auto& nexus : features.destination_nexuses(id)) {
+                    //TODO in a DENDRIDIC network, only one destination nexus per catchment
```
PhilMiller, line 118:
```suggestion
                    //TODO in a DENDRITIC network, only one destination nexus per catchment
```

include/core/SurfaceLayer.hpp
mattw-nws: This is in #494 and in https://github.com/donaldwj/ngen/pull/6 .

include/core/Layer.hpp
mattw-nws: This is in #494 and https://github.com/donaldwj/ngen/pull/6 .

PR #507, opened 2023-03-31
Bmi conventions doc

mattw-nws:

## Additions

-

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

1.

- [ ] Linux

### DIFF
--- a/doc/BMIconventions.md
+++ b/doc/BMIconventions.md
@@ -83,19 +83,74 @@ Since the [BMI Documentation] simply states that, "Use of native language type n
     * `int`, `long`, `long long`, `int64`, `longlong`, `float`, `float64`, `long double`
     * also accepts `numpy.float64` and `np.float64` but this usage is discouraged! Please use the non-namespaced names above.
 
-## Array representation
+# Input and output published variables
 
-### Contiguousness
+The BMI documentation says that input variables should be declared for a model by returning the list of inputs via `get_input_var_names`, and outputs likewise via `get_output_var_names`. However, it does not specify any difference in behavior for variables published by these functions. In practice this has few technical implications--for instance, it is not explicitly forbidden to call `get_value` on an input variable, nor even to call `set_value` on an output variable. 
+
+Also, it is not forbidden to call `get_value` and `set_value` for variable names that appear in neither returned list. In the remainder of this document, such variables will be referred to as "unpublished" variables, and variables appearing in either `get_*_var_names` list as "published".
+
+The openness of the BMI in these areas both adds some danger and some flexibility, which ngen will mitigate and leverage with the conventions defined below.
+
+## Input/output variables and the time loop
+
+Variables published via `get_input_var_names` and `get_output_var_names` are assumed to be part of the function of the time loop execution of the simulation, e.g. forcing inputs or model quantity outputs. Other types of variables (configuration settings, file names, a version identifier, etc.) should not be published in these functions. See below for impacts of this assumption and potential pitfalls.
+
+## Validation of formulations for input variables
+
+For all model variables returned from `get_input_var_names`, a realization configuration *must* provide inputs for those variables (either by implicit name matching or explicit name mapping). If a module in a realization publishes a variable name as an input requirement, and an input source to provide the value cannot be found, validation will fail and the simulation will not run.
+
+## Validation of formulations for output variables
+
+Unlike input variables, variables with the names returned from `get_output_var_names` are not required to be consumed by another module or output writer, and an unused/unconnected output variable will not cause a validation failure. However, if a variable from a module is consumed in the realization config (by another module or an output writer), that variable *must* be listed in `get_output_var_names`--otherwise validation will fail *even if the source module could respond to a request for it via* `get_value`. In other words, you cannot use an unpublished variable as input to another module or to be captured as output.
+
+## Calling `set_value` on output variables
+
+The ngen process should not call `set_value` on an output variable. Attempting to create a formulation that connects forcing or another module's output to an input name that is in `get_output_var_names` risks undefined behavior.
+
+## Calling `get_value` on input variables
+
+The ngen process *may* call `get_value` on a variable that is returned from `get_input_var_names`. A model *should* return the last value provided to the input variable, unmodified, in this case. At a minimum, a model *must* not crash if this occurs, and should return some reasonable value, rather than throwing an exception or returning BMI_FAILURE (throwing an exception or returning BMI_FAILURE *is* however an appropriate and preferred response if no value has yet been set for the variable).
+
+## Input and output with the same name not permitted
+
+A single variable name cannot be published in both `get_var_input_names` and `get_var_output_names`--this will cause a validation error. Furthermore, it is not presently possible to work around this situation with variable name mapping, because <!-- TODO: This may change with a config file format refactor? --> a variable name mapping does not specify whether it is an input or output variable and could therefore not map only one or the other.
+
+# Unpublished variables
+
+As discussed above, it is not defined by the BMI specification that you may *only* call `get_value` and `set_value` on variable names returned by either `get_input_var_names` or `get_output_var_names`. We refer to variables that appear in neither list--but are understood by a BMI module--as "unpublished" variables. Because ngen uses the input and output variable listings to validate formulations, there are many use cases for such variables, for instance:
+
+* Initialization parameters
+* Calibratable parameters
+* State inspection (for serialization or debugging)
+* Metadata variables (spatial reference information, hyperparameters, etc.) 
+
+These and other uses may be applied by ngen. Specifically, ngen will make use of specified, reserved unpublished variables enumerated later in this document. Therefore, a module should be able to communicate via unpublished variables or at least not malfunction in the case of unpublished variables being used, as described below.
+
+Note that there is currently no discoverability mechanism for unpublished variables, they must be described in documentation (such as this document or a module's  documentation) and often manually implemented or configured for use (such as in a realization config file).
+
+## Unknown/unexpected variables passed to `set_value`
+
+Importantly, if an unexpected variable name is passed to `set_value` this *must* not cause an error or instability! Consider the unknown unpublished value to be "offered" and it can be ignored<!-- TODO: " (unless otherwise specified for specific variable names later in this document)"? We might have a required one at some point. -->. Generally, the preferred behavior in the case of an unknown variable is to do nothing and report success (i.e. throw no exception and return `BMI_SUCCESS`).
+
+This scenario should be avoided by ngen, but there may be cases where ngen does not yet know whether a module supports an unpublished variable and will try to set it. These attempts can be ignored if the module does not support or understand the variable. (Returning `BMI_SUCCESS` from `set_value` does *not* imply that the module understands/supports the variable, see below section on probing.)
+
+## Must support variable metadata functions
+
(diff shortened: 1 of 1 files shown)


doc/BMIconventions.md
```diff
@@ -0,0 +1,313 @@
...
+* Initialization parameters
+* Calibratable parameters
+* State inspection (for serialization or debugging)
+* Metadata variables (spatial reference information, hyperparameters, etc.) 
+
+These and other uses may be applied by ngen. Specifically, ngen will make use of specified, reserved unpublished variables enumerated later in this document. Therefore, a module should be able to communicate via unpublished variables or at least not malfunction in the case of unpublished variables being used, as described below.
+
+Note that there is currently no discoverability mechanism for unpublished variables, they must be described in documentation (such as this document or a module's  documentation) and often manually implemented or configured for use (such as in a realization config file).
+
+## Unknown/unexpected variables passed to `set_value`
```
hellkite500, line 131: Umm...I'm not sure I agree with this statement.  Would be happy to discuss in more detail.

doc/BMIconventions.md
```diff
@@ -0,0 +1,313 @@
...
+
+## Layout and Flattening
+
+### C ordering required
+
+The [BMI best practices](https://bmi.readthedocs.io/en/stable/bmi.best_practices.html) document states:
+
+> BMI functions always use flattened, one-dimensional arrays...It’s the developer’s responsibility to ensure that array information is flattened/redimensionalized in the correct order.
+
+However, strictly speaking, *how* a multi-dimensional array should be flattened into a one-dimensional one is never directly addressed. **For ngen, it is required that flattening happens as if the array was a C array, sometimes referred to as "row major order".** That is, if you create a multi-dimensional array in C (or a C array in C++) it will already be in the appropriate order and layout such that if the data is copied into a 1D array (or the pointer is passed as the result of `get_value_ptr`) it is already in the correct order/layout and will "just work".
```
hellkite500, line 159: It is not guaranteed to be contiguous though if you are using pointers.  Should probably note that caveat.

doc/BMIconventions.md
```diff
@@ -0,0 +1,313 @@
...
+
+# Compliance with the BMI v2.0 Standard
+
+BMI modules written in C, Fortran, and C++ *must* be compiled with the exact same header files as those used with ngen itself, and this means the BMI v2.0 header files provided in the CSDMS [`bmi-c`](https://github.com/csdms/bmi-c/blob/e6f9f8a0ab326218831000b4a5571490ebc21ea2/bmi.h), [`bmi-fortran`](https://github.com/csdms/bmi-fortran/blob/e34cd026f57cabd009ef0f662fc672baee66e442/bmi.f90), and [`bmi-cxx`](https://github.com/csdms/bmi-cxx/blob/be631dc510b477b3bc3eb3c8bbecf3d04ec4005c/bmi.hxx) repositories. Python models must inherit from the `Bmi` class in the [`bmipy`](https://pypi.org/project/bmipy/2.0/) package. Python and C++ classes can be subclasses of the standard BMI interface, but such subclasses must inherit from the standard versions listed here.
+
+This means among other things that it is not possible to "extend" the BMI interface simply by adding functions to the base interface or header files. In any case, even with subclassed BMI models in Python and C++ with additional functions, ngen would not call any such additional functions because it does not know that they exist.
```
hellkite500, line 13: "This means, among other things, that it is..." A little comma grammar...

doc/BMIconventions.md
mattw-nws: The header? Or the section comments? Sure, here, or hit me up on Slack.

doc/BMIconventions.md
mattw-nws: Ah, had to reread... okay, yes, if you create a C array of pointers that's different. Is there a name for the "correct" allocation method?

doc/BMIconventions.md
mattw-nws: My problem is usually more overuse of commas. This looks like a job for... EMDASH! 🦸‍♂️

doc/BMIconventions.md
hellkite500: Why should `set_value` not return `BMI_FAILURE`?  If ngen is using unpublished conventional variables, these are essentially OPTIONAL and it *should* explicitly check to ensure that `BMI_SUCCESS` was returned from `set_value` to ensure that it can properly assume the semantics are valid.  `BMI_FAILURE` would indicate that those assumptions are not valid.  That is how calibration parameters are handled right now -- `BMI_FAILURE` just assumes that variable is not calibratable.  When we get to grid conventions, it is almost imperative for `BMI_FAILURE` to be returned if a call to `set_value( 'grid_X_shape')` fails...this allows the model engine to know that the grid cannot be manipulated by the framework, and it can test if the model implements the correct grid functions and their values align in a meaningful fashion.

mattw-nws: Ah. Okay, this part I was less sure about. Generally, I wasn't sure how ngen would handle `set_value` returning `BMI_FAILURE` and whether it would consider that an error for a convention variable. You are probably right here, especially if the ngen code already expects these semantics.

The only thing left in my mind is, if the module supports the variable, but encounters an unexpected problem, what then? For C++/Python we can recommend specific exceptions but we're kindof out of options for C/Fortran. Probably no good answer, but it's a likely scenario. Will it be important to distinguish between support+failure vs. nosupport?

hellkite500: Nope, not without enumerated error codes...BMI 3.0 recommendation =)

doc/BMIconventions.md
```diff
@@ -0,0 +1,313 @@
...
+-->
+```Fortran
+real, dimension(0:3,0:2,0:1) :: var ! Note the reversal of the dimension sizes
+integer:: x,y,z
+real:: v = 0.0
+zloop: do z = 0, 1
+    yloop: do y = 0, 2
+        xloop: do x = 0, 3
+            v = v + 0.01
+            var(x,y,z) = v ! Note the indices ordering here
```
PhilMiller, line 208: `3, 2, 1` != `4, 3, 2` like I'd expect for equivalence to the other examples here

mattw-nws: Wait, what? It's been a while but I ended up running these examples a lot... you saw the `0:...` up in the `dimension(...)` call, right?

PhilMiller: Ah, I see now. I forgot that the way Fortran writes things ends up with closed intervals, not half-open intervals like C or Python.

doc/BMIconventions.md
```diff
@@ -0,0 +1,313 @@
...
+## `update` vs. `update_until`
+
+For the foreseeable future, ngen will always call `update_until(...)` and will not call a model's `update()` method. Per the [BMI documentation](https://bmi.readthedocs.io/en/stable/#update-until) for `update_until`, an absolute time is passed as the argument, not an incremental time; i.e., if the model start time is `0.0` and ngen is incrementing timesteps by 3600 seconds, the first three iterations will call `update_until` with the values `3600.0`, `7200.0`, and `10800.0`.
+
+At present, ngen will always pass a value greater than the previous value passed to `update_until` to any model instance (it will move only forward in time).
+
+## Time Units
+
+At present, all BMI modules must use seconds as the time units. The output of `get_time_units()` should be `seconds` and the value passed to `update_until(...)` should be interpreted as seconds past time `0` (see below). This is something that we intend to make more flexible in the future.
+
```
PhilMiller, line 38:
```suggestion
```
Obviated by #633

doc/BMIconventions.md
```diff
@@ -0,0 +1,313 @@
...
+#### Clarification for `get_value_at_indices` and `set_value_at_indices`
+
+The BMI documentation for [`get_value_at_indices`]() and [`set_value_at_indices`]() states that the `inds` parameter designates:
+
+> ...the locations specified by the one-dimensional array indices in the `inds` argument...Both `dest` and `inds` are flattened arrays....
+
+However if read perfectly literally, `inds` is by necessity already a one-dimensional array and therefore can't be flattened--rather, the implication is that the indexes in `inds` are *indexes into a flattened representation of the target variable array*. That is, in the example above, an `inds` array containing `[2, 21]` should return `[0.03, 0.22]`. However, if the flatteing is assumed to be done differently by different parts of the system, the wrong values may be retrieved or set.
```
PhilMiller, line 272:
```suggestion
However if read perfectly literally, `inds` is by necessity already a one-dimensional array and therefore can't be flattened--rather, the implication is that the indexes in `inds` are *indexes into a flattened representation of the target variable array*. That is, in the example above, an `inds` array containing `[2, 21]` should return `[0.03, 0.22]`. However, if the flattening is assumed to be done differently by different parts of the system, the wrong values may be retrieved or set.
```

doc/BMIconventions.md
mattw-nws, line 13:
```suggestion
This means&mdash;among other things&mdash;that it is not possible to "extend" the BMI interface simply by adding functions to the base interface or header files. In any case, even with subclassed BMI models in Python and C++ with additional functions, ngen would not call any such additional functions because it does not know that they exist. 
```

doc/BMIconventions.md
mattw-nws: Noting that follow-on discussion noted that BMI documentation is very vague about what should be passed to `update_until`, and whether the value returned by `get_time_units()` applies to `update_until()` ... recommend leaving this as is.

doc/BMIconventions.md
```diff
@@ -0,0 +1,313 @@
...
+
+These and other uses may be applied by ngen. Specifically, ngen will make use of specified, reserved unpublished variables enumerated later in this document. Therefore, a module should be able to communicate via unpublished variables or at least not malfunction in the case of unpublished variables being used, as described below.
+
+Note that there is currently no discoverability mechanism for unpublished variables, they must be described in documentation (such as this document or a module's  documentation) and often manually implemented or configured for use (such as in a realization config file).
+
+## Unknown/unexpected variables passed to `set_value`
+
+Importantly, if an unexpected variable name is passed to `set_value` this *must* not cause an error or instability! Consider the unknown unpublished value to be "offered" and it can be ignored<!-- TODO: " (unless otherwise specified for specific variable names later in this document)"? We might have a required one at some point. -->. Generally, the preferred behavior in the case of an unknown variable is to do nothing and report success (i.e. throw no exception and return `BMI_SUCCESS`).
```
mattw-nws, line 133:
```suggestion
Importantly, if an unexpected variable name is passed to `set_value` this *must* not cause an error or instability! Consider the unknown unpublished value to be "offered" and it can be ignored<!-- TODO: " (unless otherwise specified for specific variable names later in this document)"? We might have a required one at some point. -->. Generally, the preferred behavior in the case of an unknown variable is to do nothing (i.e. throw no exception) and return `BMI_FAILURE`&mdash;ngen will interpret this return value as indicating that the variable just set is not supported by the module.
```

doc/BMIconventions.md
mattw-nws, line 159:
```suggestion
However, strictly speaking, *how* a multi-dimensional array should be flattened into a one-dimensional one is never directly addressed. **For ngen, it is required that flattening happens as if the array was a contiguous C array, sometimes referred to as "row major order".** That is, if you create a contiguous (i.e. not using pointers) multi-dimensional array in C (or a C array in C++) it will already be in the appropriate order and layout such that if the data is copied into a 1D array (or the pointer is passed as the result of `get_value_ptr`) it is already in the correct order/layout and will "just work".
```

hellkite500 (approved):

PR #511, opened 2023-04-13
Implement bracketed unit parsing in forcing CSV headers

### LINKED ISSUE #510: Support bracketed units in CsvPerFeatureForcingProvider
Support CSV headers with units at the end in either `[]` or `()`

## Current behavior

- The entire header is used for the variable name--if any units in brackets are in the header, they become part of the variable name.

## Expected behavior

- CsvPerFeatureForcingProvider should:
  - identify if either `]` or `)` is the last character in the column name. If so...
  - capture the substring between the last character closing bracket and the matching open-bracket, making this the units string for the variable
    - if the appropriate opening bracket does not appear in the header, revert to the previous behavior (the ending bracket becomes part of the variable name)... in the future, this may throw an exception instead.
  - strip the bracketed substring from the variable name
  - strip any whitespace at the end of the remaining variable name (i.e. whitespace that was between the bracketed units and the rest of the name)

### Examples
| Header | Variable Name | Units string
|-|-|-
| `precip_rate` | `precip_rate` | [empty string] 
| `precip_rate[m/s]` | `precip_rate` | `m/s`
| `precip_rate [m/s]` | `precip_rate` | `m/s`
| `precip_rate [m/s)` | `precip_rate [m/s)` | [empty string]

1.

program--:
Resolves #510 by implementing unit parsing within brackets or parentheses appended to column names in forcing CSV headers.

## Additions

- Adds direct implementation in `CsvPerFeatureForcingProvider::read_csv`.
- Adds unit tests to check for header name resolution (i.e. removing unit declaration and whitespace), correct unit conversions, and correct parsing with non-well-known columns.
- Adds new test data file from cat-27115 with unit declarations, and new non-well-known column.

## Testing

- All existing and new unit tests pass locally on Linux with GCC version 12.2.1.

## Todos

- Add documentation for unit declaration syntax.

## Checklist

- [X] Linux


test/forcing/CsvPerFeatureForcingProvider_Test.cpp
```diff
@@ -164,3 +176,63 @@ TEST_F(CsvPerFeatureForcingProviderTest, TestGetAvailableForcingOutputs)
 
 }
 
+///Test CSV Units Parsing
```
hellkite500, line 179: Might be good to add a test for extra characters after the unit brackets, e.g. `var (unit) extra, ...`

include/forcing/CsvPerFeatureForcingProvider.hpp
```diff
@@ -307,7 +307,23 @@ class CsvPerFeatureForcingProvider : public data_access::GenericDataProvider
...
+                const auto var_name_close = var_name.back();
+                if (var_name_close == ']' || var_name_close == ')') {
+                    // found closing bracket/parenth
+
+                    const bool is_bracket = var_name_close == ']';
+                    const size_t var_name_open = is_bracket ? var_name.rfind('[') : var_name.rfind('(');
+                    if (var_name_open != std::string::npos) {
+                        // found matching opening bracket/parenth
+
+                        units = var_name.substr(var_name_open + 1);
```
hellkite500, line 320: My only concern is what happens with extra characters after the closing bracket/paren.

hellkite500 (commented):
Other than testing the case mentioned, the implementation looks reasonable to me.

include/forcing/CsvPerFeatureForcingProvider.hpp
program--: are we expecting anything after a unit declaration? My understanding was that unit declarations would be required to be at the end, or would not be parsed at all (see the outer `if` construct)

hellkite500: Not expecting it, but should be prepared...a warning, error, or something... Would not catching that case cause some weird behavior in the units that we don't expect or can't track down easily?

mattw-nws: The issue was originally written to treat this case as brackets that do not represent a unit.... in order for the unit to be recognized it would have to be the last thing in the string. We can revisit if we want.

The thought was that there might be (for some reason??) column names with another set of brackets representing something else. In that case, if they were not at the end, then the brackets and their contents would be ignored by the unit parser. If they were at the end, then the workaround could be to add an extra set of different brackets to the end, which would either have valid units or be empty and picked up as an empty units string (didn't cover this case in the issue though, probably).

test/forcing/CsvPerFeatureForcingProvider_Test.cpp
```diff
@@ -164,3 +176,63 @@ TEST_F(CsvPerFeatureForcingProviderTest, TestGetAvailableForcingOutputs)
...
+        const auto val = this->Forcing_Object_3->get_value(
+            CSVDataSelector(*ite, t, 3600, ""),
+            data_access::SUM
+        );
+        const std::string cerr_output = testing::internal::GetCapturedStderr();
+
+        if (ite - expected.begin() < 6) {
+            // we are expecting warnings, since udunits is trying to map some unit to no units
+            // if we don't get a warning, then the unit type was not parsed
+            EXPECT_NE(cerr_output, "");
```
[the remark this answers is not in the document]
program--: I agree... I'm not a fan of the way I implemented these cases haha.

One way that I think would be better is to expose a getter on `CsvPerFeatureForcingProvider` that returns a read-only reference to the `available_forcings_unit` map. Then, I can test for expected units directly. How does that sound?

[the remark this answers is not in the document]
mattw-nws: If we had such a method it would probably belong in `DataProvider` though... that may be a good idea or may be something we don't want to do--in any case then you'd have to implement it in `NetCDFPerFeatureDataProvider` also... not hard, but scope creep.

How about, since we know the units that it's supposed to be, we go ahead and provide a valid unit for conversion to `CSVDataSelector` for cases 0-5 and what _would_ be a valid unit for conversion for cases 6-8, then check if the result of `get_value` changes (in case 0-5 it should and in cases 6-8 it should not)? Make sense? Flaw?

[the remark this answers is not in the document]
program--: See commit 729e70a, this implements your suggestion. I've also removed the unit conversion checks at the end of this unit test, since they are redundant with what's tested in `TestForcingDataUnitConversion`. Thanks!

mattw-nws (approved):

PR #516, opened 2023-04-25
fix t-route module build to install numpy

hellkite500:
t-route integration test fails to build t-route missing numpy dependency.  This should ensure it is installed in the venv before attempting to build.

### DIFF
--- a/.github/workflows/module_integration.yml
+++ b/.github/workflows/module_integration.yml
@@ -145,7 +145,7 @@ jobs:
           pip install -U pip
           pip install -U setuptools
           pip install -r requirements.txt
-          pip install deprecated pyarrow tables geopandas
+          pip install deprecated pyarrow tables geopandas numpy
           if [ ${{ runner.os }} == 'macOS' ] 
           then
             export LIBRARY_PATH=/usr/local/lib/gcc/11/


hellkite500:
Really confused as to why this fails to find numpy, I was pretty sure it should have been there, but figured it was worth a try to add it explicity, but it doesn't seem to help...
```sh
Requirement already satisfied: numpy in /home/runner/work/ngen/ngen/.venv/lib/python3.10/site-packages (from -r requirements.txt (line 1)) (1.24.3)
```
```sh
      ModuleNotFoundError: No module named 'numpy'
```

mattw-nws:
Yeah it looks like restricting the pip version <23.1 fixes this… the following build message is relevant, but I’m unclear exactly what to do about it:

`  DEPRECATION: troute.network is being installed using the legacy 'setup.py install' method, because the '--no-binary' option was enabled for it and this currently disables local wheel building for projects that don't have a 'pyproject.toml' file. pip 23.1 will enforce this behaviour change. A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/11451`

hellkite500:
@mattw-nws I can't approve/review since I authored the original PR, but I think this is good enough for now, and I'll open an issue to track the above deprecation warning and remind us to clean this up.  Can you approve and we can merge this?

mattw-nws:
@aaraney do you recognize the problem here and what we're supposed to do about it? This same issue--I think--affects this workaround for ngen+hdf5+pytables in DMOD's Dockerfile: https://github.com/NOAA-OWP/DMOD/blob/master/docker/main/ngen/Dockerfile#L350 ... and it may already be broken.

PR #518, opened 2023-04-25
Implement proper handling of dirent in FormulationManager, fixes #517

### LINKED ISSUE #517: A few portability issues with the use of dirent for validating forcing file paths in model engine.
In [this directory iteration loop](https://github.com/NOAA-OWP/ngen/blob/7551590a415b89026559c1c570d4154e4746161b/include/realizations/catchment/Formulation_Manager.hpp#L417) in the FormulationManager, there are two potential issues.

First, [according to the documentation](https://www.gnu.org/software/libc/manual/html_node/Directory-Entries.html#:~:text=This%20member%20is%20a%20BSD%20extension.%20The%20symbol%20_DIRENT_HAVE_D_TYPE%20is%20defined%20if%20this%20member%20is%20available.), `d_type` is not guaranteed to be a member of the `dirent` struct on all systems, so this requires a macro guard using `#ifdef _DIRENT_HAVE_D_TYPE` to prevent compiler errors on such systems.

Second, `d_type` [is not guaranteed to be provided by all file systems](https://www.gnu.org/software/libc/manual/html_node/Directory-Entries.html#:~:text=The%20type%20is%20unknown.%20Only%20some%20filesystems%20have%20full%20support%20to%20return%20the%20type%20of%20the%20file%2C%20others%20might%20always%20return%20this%20value.).  As such, a fallback to `stat` is required for cases when `DT_UNKOWN` is returned.  Note that reading the `dirent` struct is more efficient, and should be tried first, but this fallback is needed as well.

hellkite500:
Fixes a bug that prevents forcing files from being properly discovered/validated on certain systems/file-systems.

## Changes

- Refactor directory iteration to fallback to `stat` when appropriate.

## Testing

1.  All unit tests tested locall

## Todos

- Should check any other places in the code potentially using `dirent` and `readdir` for similar issues.

## Checklist

- [x] Linux
- [x] MacOS

### DIFF
--- a/include/realizations/catchment/Formulation_Manager.hpp
+++ b/include/realizations/catchment/Formulation_Manager.hpp
@@ -6,6 +6,7 @@
 #include <tuple>
 #include <functional>
 #include <dirent.h>
+#include <sys/stat.h>
 #include <regex>
 
 #include <boost/property_tree/ptree.hpp>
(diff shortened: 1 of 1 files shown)


include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -411,19 +412,45 @@ namespace realization {
...
+                                return forcing_params(
+                                    path + entry->d_name,
+                                    provider,
+                                    simulation_time_config.start_time,
+                                    simulation_time_config.end_time
+                                );
+                            }
+                            else if ( entry->d_type == DT_UNKNOWN )
+                            #endif
+                            {
```
christophertubbs, line 432: This is throwing me off - if the _DIRENT_HAVE_D_TYPE condition doesn't pass, does this yield:

```c++
if( match ) {
    {
        struct stat st;
        if( stat((path+entry->d_name)>c_str(), &st) != 0) {...
    }
}
```

christophertubbs (approved):

include/realizations/catchment/Formulation_Manager.hpp
hellkite500: Yes, it might be a little easier to see from this SS where the IDE does some nice high/low lighting

image

donaldwj (commented):
Where is _DIRENT_HAVE_D_TYPE defined and is it a gcc extension? Ie will this work with intel compilers

hellkite500:
> Where is _DIRENT_HAVE_D_TYPE defined and is it a gcc extension? Ie will this work with intel compilers

This is in `dirent.h` (see [the documentation](https://www.gnu.org/software/libc/manual/html_node/Directory-Entries.html#:~:text=This%20structure%20may,macro%20_DIRENT_HAVE_D_RECLEN.) about the additional fields.)

PR #519, opened 2023-04-26
Test cleanup

hellkite500:
Some long overdue cleanup of some test classes and cases.  The changes here are mostly focused on implementing file path checking/substitution to allow the tests to discover required input data from different initial working directories.  Ultimately, this now allows `ctest` to run effectively and do both test discovery and execution.  So from the repo root directory, one can do the following:

Activate most all features for a build
```sh
make -B test_build -S . -DNETCDF_ACTIVE:BOOL=On -DBMI_C_LIB_ACTIVE:BOOL=On -DBMI_FORTRAN_ACTIVE:BOOL=On -DNGEN_ACTIVATE_PYTHON:BOOL=On -DNGEN_ACTIVATE_ROUTING:BOOL=On
```

Compile all targets (including activated tests for the various options)
```sh
cmake --build test_build --target all -j 8 
```

Discover and run all built tests
```sh
cmake --build test_build --target test
```

The builds and tests can also be done via `make`, e.g.
```sh
cd test_build
make test
```

What is important to note here, is that we have a couple standard test targets we test often, `test_unit` and `test_all`, which have 152 and 128 total possible tests, for a combined 280 tests.  (using the above build flags).

When you run the entirety of the discoverable tests using `ctest`, there are 822 tests discovered!  Even with the overlap in test_unit/test_all, there are clearly some test cases not be exercised frequently (some of where are fixed up in this PR).

This should have no regression on existing testing workflows, but should allow for more comprehensive ones, and we may even want to revisit the automated action runners and consider using the ctest interface to discover and run tests.

## Checklist

- [x] Linux
- [x] MacOS

### DIFF
--- a/include/realizations/catchment/AbstractCLibBmiAdapter.hpp
+++ b/include/realizations/catchment/AbstractCLibBmiAdapter.hpp
@@ -91,7 +91,12 @@ namespace models {
                 if (!utils::FileChecker::file_is_readable(bmi_lib_file)) {
                     //Try alternative extension...
                     size_t idx = bmi_lib_file.rfind(".");
-                    std::string alt_bmi_lib_file;                    
+                    std::string alt_bmi_lib_file;  
+                    if(bmi_lib_file.length() == 0){
+                        this->init_exception_msg =
+                                "Can't init " + this->model_name + "; library file path is empty";
+                        throw std::runtime_error(this->init_exception_msg);
+                    }                
                     if(bmi_lib_file.substr(idx) == ".so"){
                         alt_bmi_lib_file = bmi_lib_file.substr(0,idx) + ".dylib";
                     } else if(bmi_lib_file.substr(idx) == ".dylib"){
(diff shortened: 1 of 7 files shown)


donaldwj (approved):
I see no problems. On the exception handling commits I don't think we are losing information so long as the caught exception puts all needed information into its e.what() method. Admittedly there is no guarantee that all child exception classes will do this.

mattw-nws:
It looks like this was merged in spite of two failing tests, because the failing tests were believed to be fixed by [7550dfc](https://github.com/NOAA-OWP/ngen/pull/516/commits/7550dfcd3b6b9f04608c1fbc45f11d61417c463d) (in #516) ... but after that commit, the same tests in this PR would have failed for a different reason, which is now happening in #524.

I am going to roll back some of the pathing changes in the dependent files in `RoutingPyBindTest` to fix the tests, but this may make some of the goals of this PR unachieved--particularly the `cd test_build; make test` workflow. However, since these tests rely on static YAML files with relative paths in them, there is not a better answer at the moment--I continue to doubt whether the ability to run the tests from three different locations in the hierarchy is a worthwhile goal, for this and other reasons. At a minimum (and perhaps maximum), the tests should work when run from the repo root, and making them run from anywhere else incurs cost.

Issue #496, opened 2023-03-06
Hydrofabric GeoPackage support via GDAL


aaraney:
While investigating the hydrofabric’s schema to inform work related to model specific BMI configuration files, a few cross-cutting questions bubbled up that will likely impact NGEN development. The purpose of this issue is to document these questions, discover other questions, and hopefully find resolutions to these questions that assist in adding GeoPackage support to NGEN (related to #295). 

- Do GeoPackage’s support relationships? It appears several fields in the Hydrofabric could benefit from classic foreign key like constraints.
  - Can you specify foreign key constraints when creating a GeoPackage using standard tooling like GDAL and QGIS?
  - Can you specify that a column has a `UNIQUE NOT NULL` constrain when creating a GeoPackage using standard tooling like GDAL and QGIS?
  - Can you constrain that GeoPackage table schema’s are `STRICT`? GeoPackage’s are `sqlite3` databases after all and without something like a `STRICT` table constraint `sqlite3` entries are duck typed. 
- Can GDAL query tables in a GeoPackage that do not contain a `geom` field and are not in the `gpkg_contents` table?

In the expandable section below, ive included the hydrofabric geopackage schema dumped from the [v1.2 HUC 01 NGEN hydrofabric](https://nextgen-hydrofabric.s3.amazonaws.com/v1.2/nextgen_01.gpkg). 

<details>
<summary>
Hydrofabric 1.2 GeoPackage SQL schema
</summary>

Note, ive only included the tables relevant to the hydrofabric. Also, this schema deviates from the schema documented on the `hydrofabric`’s [GitHub.io](http://github.io/) page([source](https://github.com/NOAA-OWP/hydrofabric/blob/dfe0b30e1097a4b532a75285d4413743dc2e8d32/docs/schema.Rmd), [live](https://noaa-owp.github.io/hydrofabric/schema.html)).

```SQL
CREATE TABLE IF NOT EXISTS "flowpaths" (
    "fid" INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
    "geom" LINESTRING,
    "id" TEXT,
    "lengthkm" REAL,
    "main_id" MEDIUMINT,
    "member_comid" TEXT,
    "tot_drainage_areasqkm" REAL,
    "order" REAL,
    "realized_catchment" TEXT,
    "toid" TEXT
);

CREATE TABLE IF NOT EXISTS "divides" (
[...]

PR #522, opened 2023-04-26
GeoPackage input support

### LINKED ISSUE #295: GeoPkg Input Support
Enhance hydrofabric reading by supporting geopkg inputs directly.

## Current behavior

NGen only supports a single input format for hydrofabric, namely `geojson`.

## Expected behavior

As the size of the hydrofabric increases, it will be worth supporting other formats that are less disk and memory intensive.  Geopkg may be a good format to support, as this is generated natively by the hydrofabric generation in the first place.

## Notes

GeoPkg works with SQLite, so this brings, minimally, a SQLite dependency.  There are a few possible avenues to explore. 

*  We could use SQLite directly and  implement the [spec](https://www.geopackage.org/) required for reading the geo data from a geopkg input.

* A better option would be to use the exiting [libgpkg](https://github.com/luciad/libgpkg) as a depndency.  This is an (Apache 2.0 Liscenced)[http://www.apache.org/licenses/LICENSE-2.0] code, and would provide read/write capabilities.

* A third option would be to look into [SpatiaLite](https://www.gaia-gis.it/fossil/libspatialite/index). Versions >= 4.2.0 support GeoPkg, and would bring some additional spatial capabilities that could be leveraged.

program--:
This PR implements support for GeoPackage input via the SQLite3 C library based on the [OGC GeoPackage Encoding Standard](https://www.geopackage.org/spec/).

(resolves #295; related: #496)

> **Note**: the current additions are a bit of a mess, but will be cleaned up once this PR is no longer a draft.

## Additions

- Adds SQLite3 as a dependency.
- Adds a simple C++ wrapper over the SQLite3 C interface (`include/geopackage/SQLite.hpp`).
- Adds a new sub-library `geopackage` that contains the primary function `geopackage::read()`, which returns a `geojson::GeoJSON` (a.k.a. `std::shared_ptr<geojson::FeatureCollection>`) object, like `geojson::read`.
- Adds a WKB reader (`geopackage::wkb`). [ref: [OGC SFA Part 1](https://www.ogc.org/standard/sfa/), [GEOS WKB](https://libgeos.org/specifications/wkb/)]

## Testing

- [x] **gcc** (version 13.1.1)
- [x] **clang** (version 15.0.7)
- [x] **icx** (version 2023.0.0.20221201)

## Notes

- This implementation currently depends on `geojson` for feature classes.
- In the future (out of scope for this PR), it may be beneficial to decouple the feature classes implementation from the GeoJSON parser, since the feature classes themselves do not require use of `boost::property_tree`, and so that this implementation doesn't require linking to `geojson`.
- If GDAL is supported in the future, this implementation can be easily refactored -- and modified to be generic for any of the GDAL/OGR vector drivers.
- Once implemented, lines 262 and 266 should be able to be modified for use with either GeoJSON or GeoPackage, since they will return the same type of object: https://github.com/NOAA-OWP/ngen/blob/6ab044e3fd3883b86e45674032ef67946426ed2c/src/NGen.cpp#L262-L266

## Todos

- [x] `geopackage::read`
  * [x] `geopackage::build_feature`
  * [x] `geopackage::build_geometry`
  * [X] `geopackage::build_properties`
- [x] Implement Standard WKB reader
  * [x] Handle projections (i.e. Hydrofabric data typically is in EPSG:5070)
  * [x] Extend reader for GeoPackage blob WKB
- [x] Add SQLite dependency within cmake
- [x] Unit tests
  * [x] SQLite Wrapper
  * [x] GeoPackage
  * [x] WKB Reader
- [x] Integration tests (i.e. successfully use geopackage::read in place of the geojson variant)
  * [ ] Add a geopackage integration test to the existing github actions workflows
- [ ] Documentation
- [x] Make GeoPackage support conditional on if SQLite is found. Otherwise, fall back to GeoJSON.

## Checklist

- [X] Linux


program--:
This PR is just about ready for review. The last things are adding unit tests for `geopackage::read` and it's used functions, as well as run the ngen integration tests using `geopackage::read` instead of `geojson::read`. I'm expecting the need to refactor quite a bit of the work I've done here (especially the WKB reader), since it is messy 😅.

program--:
There is currently an issue in this PR that occurs when the CMake Build type is set to Release with GCC and Clang. For some reason, the `boost::geometry::srs::transformation<>` object throws an error when its constructor is called.

This does not occur when The build type is set to Debug, OR when ICX is used in either Debug or Release.

mattw-nws (commented):
Will look at this more, but did a cursory pass and don't see any show stoppers. Interested in others' feedback. I did run the software on the test case in `data/gauge_01073000` using both the geojson and geopkg inputs, which succeeded and I found no differences in the output, so that's positive.

src/geopackage/read.cpp
```diff
@@ -0,0 +1,108 @@
+#include "GeoPackage.hpp"
+
+#include <numeric>
+
+std::shared_ptr<geojson::FeatureCollection> geopackage::read(
```
hellkite500, line 5: might be a good idea to add some tests, especially with the id subsetting...

hellkite500 (changes requested):
Generally looks pretty good, but a couple things that might warrant further discussion/testing, and some minor cleanup comments.

src/geopackage/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+cmake_minimum_required(VERSION 3.10)
+
+
+string(COMPARE EQUAL "${CMAKE_CXX_COMPILER_ID}" "IntelLLVM" _cmp)
+if (NOT _cmp)
+    message(WARNING "[NGen::geopackage] geometry.cpp cannot be optimized with "
+                    "${CMAKE_CXX_COMPILER_ID} due to a suspected compiler/boost issue. "
```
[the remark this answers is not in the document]
mattw-nws: True... this will affect us in ops... on the other hand the de-optimization is limited to one file, and I doubt this will be a major time sink?

Also might be worth cross-checking with #559 if we upgrade our boost version.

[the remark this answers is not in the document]
PhilMiller: If we do end up needing to debug it, I've got more than my fair share of experience working out compiler bugs and relevant work-arounds.

src/NGen.cpp
```diff
@@ -259,11 +263,29 @@ int main(int argc, char *argv[]) {
...
+      #endif
+    } else {
+      nexus_collection = geojson::read(nexusDataFile, nexus_subset_ids);
+    }
     std::cout << "Building Catchment collection" << std::endl;
 
     // TODO: Instead of iterating through a collection of FeatureBase objects mapping to catchments, we instead want to iterate through HY_Catchment objects
-    geojson::GeoJSON catchment_collection = geojson::read(catchmentDataFile, catchment_subset_ids);
+    geojson::GeoJSON catchment_collection;
+    if (boost::algorithm::ends_with(catchmentDataFile, "gpkg")) {
```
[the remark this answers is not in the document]
program--: This is a temporary "workaround", my reasoning was that I did not want to immediately change the cli interface, so the intent wasn't directly to be able to support both GPKG and GeoJSON files on the same run. I imagine once we get around to refactoring the cli, this would be handled so that just the single GPKG path is passed.

src/geopackage/read.cpp
```diff
@@ -0,0 +1,108 @@
...
+    //
+    // this creates a string in the form:
+    //     WHERE id IN (?, ?, ?, ...)
+    // so that it can be bound by SQLite.
+    // This is safer than trying to concatenate
+    // the IDs together.
+    std::string joined_ids = "";
+    if (!ids.empty()) {
+        joined_ids = " WHERE id IN (?";
+        for (size_t i = 1; i < ids.size(); i++) {
```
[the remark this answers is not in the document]
program--: No, this should skip the first index and continue till the last ID. This creates a parameterized query string, such that if `ids = {1, 2, 3}`, then:

`i`         | `joined_ids`
----------- | ------------
Not Started | `" WHERE id in (?"`
1           | `" WHERE id in (?, ?"`
2           | `" WHERE id in (?, ?, ?"`
Done        | `" WHERE id in (?, ?, ?)"`

If `ids = {1}`, then the loop is never executed, so `joined_ids` will become `" WHERE id in (?)"`.

include/geopackage/WKB.hpp
```diff
@@ -0,0 +1,403 @@
...
+        polygon.inners().resize(count);
+    }
+
+    auto outer = read_linestring(buffer, index, order);
+    polygon.outer().reserve(outer.size());
+    for (auto& p : outer) {
+        polygon.outer().push_back(p);
+    }
+
+    for (uint32_t i = 1; i < count; i++) {
```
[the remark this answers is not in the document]
program--: Ah good catch, this might be an off-by-one. `boost::geometry` does some interesting indexing for polygon rings, so there might've been a reason, but I might've also made a mistake here. I'll do some investigating and add a comment/note if it should start at 1. Thanks!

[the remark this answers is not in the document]
program--: You were correct here, I'm not sure why I started at 1 and incremented to count. An incoming commit will resolve this and adds a test for reading polygons with holes, such that the inner rings start at index 0.

include/geopackage/SQLite.hpp
```diff
@@ -0,0 +1,273 @@
...
+
+/**
+ * Smart pointer (unique) type for sqlite3 database
+ */
+using sqlite_t = std::unique_ptr<sqlite3, sqlite_deleter>;
+
+/**
+ * Smart pointer (shared) type for sqlite3 prepared statements
+ */
+using stmt_t = std::shared_ptr<sqlite3_stmt>;
```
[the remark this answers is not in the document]
hellkite500: I thought I removed this comment cause I looked that up and realized what was happening, haha

src/geopackage/feature.cpp
```diff
@@ -0,0 +1,97 @@
...
+    if (wkb_type == geojson::FeatureType::Point) {
+        const auto& pt = boost::get<geojson::coordinate_t>(geometry);
+        bounding_box[0] = pt.get<0>();
+        bounding_box[1] = pt.get<1>();
+        bounding_box[2] = pt.get<0>();
+        bounding_box[3] = pt.get<1>();
+    }
+    
+    switch(wkb_type) {
+        case geojson::FeatureType::Point:
```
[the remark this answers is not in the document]
program--: The reason for this was readability, since if I construct the bounding box within the switch, it would look like this (due to variable initialization bypassing):
```
bounding_box[0] = boost::get<geojson::coordinate_t>(geometry).get<0>();
bounding_box[1] = boost::get<geojson::coordinate_t>(geometry).get<1>();
bounding_box[2] = boost::get<geojson::coordinate_t>(geometry).get<0>();
bounding_box[3] = boost::get<geojson::coordinate_t>(geometry).get<1>();
```

In any case, I can do this, but I'm not sure that the loss in readability is worth it?

cmake/modules/FindSQLite3.cmake
```diff
@@ -0,0 +1,16 @@
...
+
+find_library(SQLITE3_LIBRARY NAMES sqlite3)
+mark_as_advanced(SQLITE3_LIBRARY)
+if(SQLITE3_LIBRARY)
+	add_library(sqlite3 SHARED IMPORTED)
+	set_property(TARGET sqlite3 PROPERTY IMPORTED_LOCATION "${SQLITE3_LIBRARY}")
+endif()
+
+include(FindPackageHandleStandardArgs)
+FIND_PACKAGE_HANDLE_STANDARD_ARGS(SQLite3 DEFAULT_MSG SQLITE3_LIBRARY SQLITE3_INCLUDE)
```
PhilMiller, line 16: Could you set your editor to include newlines at the end of files, so Git doesn't get annoyed?

program--: Yeah, sorry about that. I'm going through now to fix missing newlines

cmake/modules/FindSQLite3.cmake
```diff
@@ -0,0 +1,16 @@
+message("Looking for SQLite3...")
```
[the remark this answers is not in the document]
PhilMiller: Note that you'll need to rebase this branch for the code to match our documented requirement.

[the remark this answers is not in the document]
program--: Will remove the added CMake module in favor of the built-in one. Thanks!

include/geopackage/SQLite.hpp
```diff
@@ -0,0 +1,273 @@
+#ifndef NGEN_GEOPACKAGE_SQLITE_H
+#define NGEN_GEOPACKAGE_SQLITE_H
```
PhilMiller, line 2: I'm mildly concerned about naming a file we own simply `SQLite.hpp`, since that could look like it's code from the library in stack traces or whatever. Would you mind prefixing its name?

PhilMiller:
Would it make sense to move all of the added SQLite support code over to utilities, since nothing in it is actually specified to GeoPackage?

include/geopackage/SQLite.hpp
```diff
@@ -0,0 +1,273 @@
...
+
+/**
+ * Get a runtime error based on a function and code.
+ *
+ * @param f String denoting the function where the error originated
+ * @param code sqlite3 result code
+ * @param extra additional messages to add to the end of the error
+ * @return std::runtime_error
+ */
+inline std::runtime_error sqlite_error(const std::string& f, int code, const std::string& extra = "")
```
PhilMiller, line 54: Is there a particular reason to inline error reporting code? Usually it should be the coldest code path

include/geopackage/SQLite.hpp
PhilMiller: In that vein, I notice that all of the class and function names declared here are prefixed with `sqlite_`. It would be preferable to have everything in a namespace like `ngen::sqlite`, and then the individual prefixes could be dropped.

include/geopackage/SQLite.hpp
```diff
@@ -0,0 +1,273 @@
...
+struct sqlite_deleter
+{
+    void operator()(sqlite3* db) { sqlite3_close_v2(db); }
+    void operator()(sqlite3_stmt* stmt) { sqlite3_finalize(stmt); }
+};
+
+/**
+ * Smart pointer (unique) type for sqlite3 database
+ */
+using sqlite_t = std::unique_ptr<sqlite3, sqlite_deleter>;
```
PhilMiller, line 39: This type alias has pretty limited use, and only in code private to `class sqlite` itself. Let's move it inside that, or drop it entirely.

program--:
> Would it make sense to move all of the added SQLite support code over to utilities, since nothing in it is actually specified to GeoPackage?

Maybe? Although, GeoPackage is the only portion of NGen that is using SQLite, so in that case it would also make sense to keep it together with the GeoPackage code. I think it would make sense to move it *if* other portions of NGen made use of SQLite.

src/NGen.cpp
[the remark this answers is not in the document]
mattw-nws: Indeed.... either the CLI or config file refactor is needed to "fix" this. The HF (or the individual components of it) might belong in the realization config in a multi-layer world, because (at least eventually) different layers will use other hydrofabrics or discretizations that don't really need a hydrofabric for definition... So while the HF discretization would probably be a special case used for multiprocessing division, it would be just one of possibly many discretizations specified within a simulation.

include/geopackage/SQLite.hpp
program--: No particular reason -- I think I was working in-header at the time, but never moved that out. In my next push I'll include a change that defines this only within the source files, since it isn't really needed outside of the `sqlite`/`sqlite_iter` impls

include/geopackage/SQLite.hpp
program--: Added suggestion in 49bd6d9.

include/geopackage/SQLite.hpp
program--: File name change is implemented in 31fb4b1. I am gong to hold off on encasing the sqlite definitions inside an `ngen` namespace, since I think there is a larger reorg issue there with the other sub-libraries in NGen, and that should be addressed separately (I think)

src/geopackage/read.cpp
program--: 6a0bb77 adds a test for the subsetting. There already is a test for reading a GeoPackage using this function (and for projection), but without adding more GeoPackage files, there isn't much more to test (as far as I can think of). Since the output is a pointer to a `geojson::FeatureCollection`, the tests for that are handled within the geojson tests.

include/geopackage/NGen_SQLite.hpp
```diff
@@ -0,0 +1,235 @@
...
+
+    /**
+     * Construct a new sqlite object from a path to database
+     *
+     * @param path File path to sqlite3 database
+     */
+    sqlite(const std::string& path);
+
+    sqlite(sqlite& db);
+    sqlite& operator=(sqlite& db);
```
PhilMiller, line 174: Given that these are implemented as move operations, and don't leave the LHS and RHS in an equivalent state, I think they should actually be ` = delete;`. If there are places in the code that are calling them and consequently fail to compile, those should be fixed with `std::move` to make the consumption/change-of-ownership effect explicit.

program--: Oh shoot, you're right. These should be deleted.

src/geopackage/sqlite/database.cpp
```diff
@@ -0,0 +1,124 @@
...
+    this->conn = std::move(db.conn);
+    this->stmt = db.stmt;
+}
+
+sqlite& sqlite::operator=(sqlite&& db)
+{
+    this->conn = std::move(db.conn);
+    this->stmt = db.stmt;
+    return *this;
+}
```
PhilMiller, line 50: Should these set the RHS `stmt` member to `nullptr`? Just saying `this->stmt = std::move(db.stmt);` would accomplish that, since `std::shared_ptr::{shared_ptr,operator=}(shared_ptr&&)` does that.

At that point, we're effectively just writing out `sqlite(sqlite&&) = default;` in long form, and should just do that.

src/geopackage/sqlite/iterator.cpp
```diff
@@ -0,0 +1,152 @@
...
+sqlite_iter::sqlite_iter(stmt_t stmt)
+  : stmt(stmt)
+{
+    // sqlite3_column_type requires the last result code to be
+    // SQLITE_ROW, so we need to iterate on the first row.
+    // TODO: need to test how this functions if the last result code
+    //       was SQLITE_DONE.
+    this->next();
+    this->column_count = sqlite3_column_count(this->ptr());
+    this->column_names = std::vector<std::string>();
```
PhilMiller, line 33: Default constructors have already run for any data member not named in the initializer list. There's no need to repeat initialization (and it may be a pessimization)

mattw-nws:
> Would it make sense to move all of the added SQLite support code over to utilities, since nothing in it is actually specified to GeoPackage?

Possibly... the question might be whether we anticipate using SQLite for anything _other_ than GeoPKG... or whether we encourage or discourage doing so. This could go in a direction where we remove the direct SQLite dependency in favor of GDAL w/GPKG support--in which case we'd be stuck with it if other things use the format. I don't think we know yet which thing we want. I'm slightly inclined to keep the two tied right now to suggest this dependency shouldn't be used elsewhere, unless/until we make a conscious decision to expand its use.

PhilMiller:
> > Would it make sense to move all of the added SQLite support code over to utilities, since nothing in it is actually specified to GeoPackage?
> 
> Possibly... the question might be whether we anticipate using SQLite for anything _other_ than GeoPKG... or whether we encourage or discourage doing so. This could go in a direction where we remove the direct SQLite dependency in favor of GDAL w/GPKG support--in which case we'd be stuck with it if other things use the format. I don't think we know yet which thing we want. I'm slightly inclined to keep the two tied right now to suggest this dependency shouldn't be used elsewhere, unless/until we make a conscious decision to expand its use.

Ok, thanks for the context. That makes sense.

include/geopackage/NGen_SQLite.hpp
```diff
@@ -0,0 +1,236 @@
...
+class sqlite
+{
+  private:
+    /**
+     * Smart pointer (unique) type for sqlite3 database
+     */
+    using sqlite_t = std::unique_ptr<sqlite3, sqlite_deleter>;
+  
+    sqlite_t conn = nullptr;
+    stmt_t   stmt = nullptr;
```
PhilMiller, line 161: Why does `class sqlite` have a `stmt` member at all? It doesn't do anything with it besides store it. Having it muddies ownership of the underlying `sqlite3_stmt` object, which could otherwise be kept in a `unique_ptr` in `sqlite_iter`

src/geopackage/sqlite/database.cpp
program--: I think these definitions were a remnant of earlier iterations when there was a bit more complexity to the query/statement relationship. I agree that these should just be `= default`.

src/geopackage/sqlite/iterator.cpp
```diff
@@ -0,0 +1,152 @@
...
+{
+    sqlite3_reset(this->ptr());
+    this->iteration_step = -1;
+    this->iteration_finished = false;
+    return *this;
+}
+
+void sqlite_iter::close()
+{
+    this->~sqlite_iter();
```
PhilMiller, line 97: Writing an explicit call to a class's own destructor is a code smell. This is primarily a round-about way of trying to say `sqlite3_finalize(ptr())`, right? It doesn't even really do that, because the `sqlite` instance we go it from still has a reference to it.

include/geopackage/WKB.hpp
```diff
@@ -0,0 +1,144 @@
...
+#include "JSONGeometry.hpp"
+#include <boost/geometry/srs/epsg.hpp>
+#include <boost/geometry/srs/transformation.hpp>
+
+namespace bg = boost::geometry;
+
+namespace geopackage {
+
+/**
+ * A recursive WKB reader
```
PhilMiller, line 14: For the benefit of those of us who are new to the project / GIS stuff in general, it would be useful to note that WKB stands for "well-known binary", which is currently specified in an ISO standard.

Also potentially useful to note whether the flavor handled is strictly what's described in the standard, or something like EWKB or some other application's adapted/extended format.

src/geopackage/feature.cpp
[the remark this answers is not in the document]
hellkite500: Write an inline function for the logic and put that call in the switch?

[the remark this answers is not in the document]
program--: 🤦🏾 Ah duh, that makes sense. I'll include this in the next push, thanks!

include/geopackage/WKB.hpp
program--: One thing to note is that this implementation **partially** follows the OGC specification (see: https://www.ogc.org/standard/sfa/), not the ISO spec (I'm not sure on the differences as I haven't read the ISO version, but most software follows the OGC spec). I'll add a note, but this is a strict WKB (i.e. not EWKB or TWKB) implementation.

I say partially because I only implement the subset of geometry types that we should expect, i.e. the common ones: point, linestring, polygon, and corresponding multi-geometries.

src/geopackage/wkb.cpp
```diff
@@ -0,0 +1,306 @@
...
+    utils::copy_from(buffer, index, count, order);
+
+    multipoint_t mp;
+    mp.resize(count);
+
+    for (auto& point : mp) {
+        const byte_t new_order = buffer[index];
+        index++;
+
+        uint32_t type;
```
PhilMiller, line 80: Do I understand correctly that `type` is being read out of the buffer and ignored? If so, do we need some sort of guard that its value matches expectations?

src/geopackage/wkb.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+    geometry g;
+    switch(type) {
+        case 1:  g = read_point(buffer, index, order); break;
+        case 2:  g = read_linestring(buffer, index, order); break;
+        case 3:  g = read_polygon(buffer, index, order); break;
+        case 4:  g = read_multipoint(buffer, index, order); break;
+        case 5:  g = read_multilinestring(buffer, index, order); break;
+        case 6:  g = read_multipolygon(buffer, index, order); break;
+        default: g = point_t{std::nan("0"), std::nan("0")}; break;
```
PhilMiller, line 156: Should the `default` case also raise an exception, rather than returning an invalid value?

include/geopackage/GeoPackage.hpp
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @param[in] gpkg_path Path to GPKG file
+ * @param[in] layer Layer name within GPKG file to create a collection from
+ * @param[in] ids optional subset of feature IDs to capture (if empty, the entire layer is converted)
+ * @return std::shared_ptr<geojson::FeatureCollection> 
+ */
+std::shared_ptr<geojson::FeatureCollection> read(
+    const std::string& gpkg_path,
+    const std::string& layer,
+    const std::vector<std::string>& ids
```
PhilMiller, line 58: I'm not thrilled with an API where "empty subset list" means "give me everything", but it looks like that's matching what our geojson reader already does, so that's probably unreasonable to change.

My preferred alternative would probably be to take something like `std::variant` that distinguishes the "give me everything" case from a specified subset.

include/utilities/EndianCopy.hpp
```diff
@@ -0,0 +1,58 @@
...
+ * @tparam S a primitive type
+ * @param src a vector of bytes
+ * @param index an integral type tracking the starting position of @param{dst}'s memory
+ * @param dst output primitive
+ * @param order endianness value (0x01 == Little; 0x00 == Big)
+ */
+template<typename S>
+void copy_from(const std::vector<uint8_t>& src, int& index, S& dst, uint8_t order)
+{
+    std::memcpy(&dst, &src[index], sizeof(S));
```
PhilMiller, line 26:
```suggestion
    std::memcpy(&dst, &src[index], sizeof(dst));
```
This spelling doesn't require the reader to look up and see that `decltype(dst)` is `S`

include/utilities/EndianCopy.hpp
```diff
@@ -0,0 +1,58 @@
...
+    // copy resolved endianness into double
+    std::memcpy(&dst, &tmp, sizeof(double));
+
+    // above call to copy_from handles index
+    // incrementing, so we don't need to.
+}
+
+}
+
+#endif // NGEN_ENDIANCOPY_H
```
PhilMiller, line 58: Still missing newline here

src/geopackage/CMakeLists.txt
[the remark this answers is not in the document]
PhilMiller: Also, do I understand correctly that this code is used in startup, and not while we're repeatedly time-stepping?

src/geopackage/feature.cpp
```diff
@@ -0,0 +1,97 @@
...
+            return std::make_shared<geojson::MultiPolygonFeature>(geojson::MultiPolygonFeature(
+                boost::get<geojson::multipolygon_t>(geometry),
+                id,
+                properties,
+                bounding_box,
+                std::vector<geojson::FeatureBase*>(),
+                std::vector<geojson::FeatureBase*>(),
+                {}
+            ));
+        default:
```
PhilMiller, line 86: This is supposed to be the case for `geojson::FeatureType::GeometryCollection`, right? Please name the case, so that the `switch` is explicitly exhaustive, and the `default` can actually raise an error or return a clearly invalid result (e.g. `nullptr`)

include/geopackage/WKB.hpp
PhilMiller: Thanks. Those are both details worth recording in the comments

src/geopackage/read.cpp
```diff
@@ -0,0 +1,108 @@
...
+            joined_ids += ", ?";
+        }
+        joined_ids += ")";
+    }
+
+    // Get number of features
+    sqlite_iter query_get_layer_count = db.query("SELECT COUNT(*) FROM " + layer + joined_ids, ids);
```
PhilMiller, line 49: Is `layer` something we potentially need to worry about as untrusted input (e.g. if there's some web front-end for launching a simulation)? I see that it's passed as a query parameter in the implementation of `has_table`, but concatenated into the query string here. I would *not* rely on the affirmative return from `has_table` to tell us that it's safe to use like this.

PhilMiller: It's also parameterized down below, on line 67

PhilMiller: And then concatenated again on 72

src/geopackage/read.cpp
```diff
@@ -0,0 +1,108 @@
...
+        );
+
+        features.push_back(feature);
+        query_get_layer.next();
+    }
+
+    // get layer bounding box from features
+    //
+    // GeoPackage contains a bounding box in the SQLite DB,
+    // however, it is in the SRS of the GPKG. By creating
```
PhilMiller, line 91: I couldn't figure out from the Boost::Geometry documentation - what's SRS stand for? It only seems to appear as `namespace boost::geometry::srs` which contains `class spheroid`

src/geopackage/read.cpp
program--: Table names can't be parameterized in the `FROM` clause, as far as I know. The one below parameterizes `layer` as part of the `WHERE` clause, which is the same in `has_table`.

src/geopackage/read.cpp
```diff
@@ -0,0 +1,108 @@
...
+    for (const auto& feature : features) {
+        const auto& bbox = feature->get_bounding_box();
+        min_x = bbox[0] < min_x ? bbox[0] : min_x;
+        min_y = bbox[1] < min_y ? bbox[1] : min_y;
+        max_x = bbox[2] > max_x ? bbox[2] : max_x;
+        max_y = bbox[3] > max_y ? bbox[3] : max_y;
+    }
+
+    const auto fc = geojson::FeatureCollection(std::move(features), {min_x, min_y, max_x, max_y});
+    return std::make_shared<geojson::FeatureCollection>(fc);
```
PhilMiller, line 107: Written like this, there will be an extra copy of the already-constructed `FeatureCollection` instance, since copy-elision can't time travel into the future to construct the local instance `fc` in the heap space allocated by `make_shared`.

```suggestion
    return std::make_shared<geojson::FeatureCollection>(std::move(features), {min_x, min_y, max_x, max_y});
```

Instead, pass the original constructor arguments to `make_shared`, and skip calling any copy constructor.

src/geopackage/sqlite/database.cpp
```diff
@@ -0,0 +1,111 @@
...
+sqlite_iter sqlite::query(const std::string& statement)
+{
+    sqlite3_stmt* stmt;
+    const auto    cstmt = statement.c_str();
+    const int     code  = sqlite3_prepare_v2(this->connection(), cstmt, statement.length() + 1, &stmt, NULL);
+
+    if (code != SQLITE_OK) {
+        // something happened, can probably switch on result codes
+        // https://www.sqlite.org/rescode.html
+        throw sqlite_error("sqlite3_prepare_v2", code, "(query: " + statement + ")");
```
PhilMiller, line 60: Again on the "if this is part of an internet-accessible-service" front, concatenating a statement (especially one we know has failed to execute) into what may be rendered as output could result in a vulnerability.

src/geopackage/read.cpp
program--: SRS stands for **Spatial Reference System**, it's the same thing as CRS, which is **Coordinate Reference System**. Here's a set of slides that's useful for showing boost's use of srs transformations: [SRS Transformations](https://archive.fosdem.org/2019/schedule/event/geo_spatialboostgeometry/attachments/slides/3306/export/events/attachments/geo_spatialboostgeometry/slides/3306/Boost_Geometry_SRS_Transformations.pdf).

Unfortunately (in my opinion), boost's namespacing of the SRS functions is really poor. The primary one that you'll need to be concerned with (as far as this PR) is `boost::geometry::srs::transformation`, which is defined in `<boost/geometry/srs/transformation.hpp>`.

src/geopackage/sqlite/database.cpp
program--: This is a good point, this should've been debug output (at most), or removed entirely. Will adjust :+1:

src/geopackage/wkb.cpp
program--: Ah, good catch. That's another remnant of older code. We should have a guard there to check that `type` is as expected.

include/utilities/EndianCopy.hpp
PhilMiller: Also, I think this should preferentially be written
```suggestion
    std::memcpy(&dst, src.data() + index, sizeof(dst));
```
to emphasize that we're not actually subscripting the vector for its element at `index`

src/geopackage/read.cpp
PhilMiller: In that case, if `layer` might be untrusted input, we need to very strictly limit its contents to a sequence of characters that represent a single valid (potential) table name and nothing more.

include/geopackage/NGen_SQLite.hpp
program--: Implemented changes in 914ec44: removed `stmt` member, changed `stmt_t` alias to a `unique_ptr`.

src/geopackage/sqlite/iterator.cpp
program--: Implemented changes in 914ec44: removed `sqlite_iter::close()`, this function wasn't being used anywhere, so I'm not sure why I implemented it (or why I implemented it to call the destructor) -- might've been old code

include/geopackage/GeoPackage.hpp
program--: This is something we can potentially change when we get around to refactoring the feature handling.

include/geopackage/WKB.hpp
program--: Added in 987d3a9.

src/geopackage/wkb.cpp
program--: Implemented guards in d0bf396.

src/geopackage/read.cpp
program--: Ah, good catch. I didn't realize at the time that `std::make_shared` did forwarding.

src/geopackage/sqlite/database.cpp
program--: Removed statement output in a6c57e0.

src/geopackage/CMakeLists.txt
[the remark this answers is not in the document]
program--: @PhilMiller Yes, that is correct. It's used when reading in the features at the beginning, i.e. here: https://github.com/NOAA-OWP/ngen/blob/e8f46463d73e3652cde57d9f5346b516655f0c6a/src/NGen.cpp#L266-L287

src/geopackage/feature.cpp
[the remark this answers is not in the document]
program--: Implemented in e8f4646.

src/geopackage/sqlite/database.cpp
```diff
@@ -0,0 +1,108 @@
...
+      errmsg += " ";
+      errmsg += extra;
+    }
+
+    return std::runtime_error(errmsg);
+}
+
+sqlite::sqlite(const std::string& path)
+{
+    sqlite3* conn;
```
PhilMiller, line 31: Tiny remark: It's not a great idea to name a local variable to shadow a member variable. That is a case where qualifying `this->conn` makes a difference, though.

src/geopackage/wkb.cpp
```diff
@@ -0,0 +1,339 @@
...
+        case 2:
+            return read_linestring(buffer, index, order);
+        case 3:
+            return read_polygon(buffer, index, order);
+        case 4:
+            return read_multipoint(buffer, index, order);
+        case 5:
+            return read_multilinestring(buffer, index, order);
+        case 6:
+            return read_multipolygon(buffer, index, order);
```
PhilMiller, line 186: Please use the named enumeration values from `wkb_geom_t`

include/geopackage/WKB.hpp
```diff
@@ -0,0 +1,149 @@
...
+    wgs84(uint32_t srs, const bg::srs::transformation<>& tr)
+        : srs(srs)
+        , tr(tr) {};
+
+    geojson::geometry operator()(point_t& g);
+    geojson::geometry operator()(linestring_t& g);
+    geojson::geometry operator()(polygon_t& g);
+    geojson::geometry operator()(multipoint_t& g);
+    geojson::geometry operator()(multilinestring_t& g);
+    geojson::geometry operator()(multipolygon_t& g);
```
PhilMiller, line 140: Given how they're implemented, the really idiomatic thing to do would be to make these all `virtual`, and have a subclass for the `srs == 4326` case. But, that's probably not actually worth doing here and now.

src/geopackage/sqlite/database.cpp
mattw-nws: Goes back to the need for a more production-ready logging capability... the statement should be included in one logging path and the message returned to the user in another (maybe the exception text, maybe not).

src/geopackage/geometry.cpp
```diff
@@ -0,0 +1,85 @@
...
+            max_prj = geojson::coordinate_t{max.get<0>(), max.get<1>()};
+            min_prj = geojson::coordinate_t{min.get<0>(), min.get<1>()};
+        } else {
+            prj.forward(max, max_prj);
+            prj.forward(min, min_prj);
+        }
+
+        // assign the projected values to the bounding_box parameter
+        bounding_box.clear();
+        bounding_box.resize(4); // only 4, not supporting Z or M dims
```
PhilMiller, line 63: Does the code check in some way that it's only getting 2D input, and Z and M dimensions don't appear?

src/geopackage/geometry.cpp
```diff
@@ -0,0 +1,85 @@
...
+        if (indicator == 2 || indicator == 3) {
+            index += 2 * sizeof(double);
+        } else if (indicator == 4) {
+            index += 4 * sizeof(double);
+        }
+    }
+
+    if (!is_empty) {
+        const std::vector<uint8_t> geometry_data(geometry_blob.begin() + index, geometry_blob.end());
+        auto wkb_geometry = wkb::read(geometry_data);
```
PhilMiller, line 79: Not something to change for this PR, but this would be another excellent use for `std::span` rather than constructing a `vector` that copies the data

src/geopackage/geometry.cpp
program--: Implicitly, yes. The `wkb_geom_t` integers are specifically 2D geometry. Geometries with Z, M, and ZM are (respectively), +1000, +2000, and +3000 from `wkb_geom_t`. So if a 3 or 4 coordinate-dimensional geometry were given in the GPKG, it would fail at the initial switch statement in `geopackage::read`.

src/geopackage/read.cpp
PhilMiller: I won't hold things up over this, since I have the least context to make a good decision.

If we do want to sanitize `layer`, I'd propose a two-layer white-list check before the `has_table` call up above
* composed only of alpha-numeric characters, `-`, and `_` (i.e. matches regular expression `[-A-Za-z0-9_]+`), and
* does not begin with `sqlite_` (because SQLite uses that for special internal tables)

I omitted `.` from my suggestion of acceptable characters because that's what SQLite uses for a "schema name" separator. We could also maybe work around that by always prefixing `layer` as a table name with a scheme name that we specify, so that subsequent `.` don't have any additional semantic.

test/geopackage/WKB_Test.cpp
```diff
@@ -0,0 +1,262 @@
...
+    void TearDown() override {}
+    
+    std::map<std::string, std::vector<uint8_t>> wkb_bytes;
+};
+
+TEST_F(WKB_Test, wkb_point_test) // also tests endianness
+{  
+    const auto geom_big    = boost::get<wkb::point_t>(wkb::read(this->wkb_bytes["point_big"]));
+    const auto geom_little = boost::get<wkb::point_t>(wkb::read(this->wkb_bytes["point"]));
+    EXPECT_NEAR(geom_big.get<0>(), geom_little.get<0>(), 0.000001);
```
PhilMiller, line 150: Stylistic thing: it's easier to read tolerances specified in scientific notation, so as not to have to count repeated `0`s

test/geopackage/WKB_Test.cpp
```diff
@@ -0,0 +1,262 @@
...
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x40
+        };
+    }
+
+    void TearDown() override {}
+    
+    std::map<std::string, std::vector<uint8_t>> wkb_bytes;
+};
+
+TEST_F(WKB_Test, wkb_point_test) // also tests endianness
```
PhilMiller, line 146: Re endian-ness, do you have some literal values in the above fixture both ways around, with the indicator to switch between them? If so, that's worth stating explicitly. I can't quite parse floating point from literal bytes in my head 😃

test/geopackage/WKB_Test.cpp
```diff
@@ -0,0 +1,262 @@
...
+        EXPECT_NEAR(poly.outer()[i].get<1>(), expected_outer[i].second, 0.0001);
+    }
+
+    const std::vector<std::pair<double, double>> expected_inner = {
+        {2, 2},
+        {3, 2},
+        {3, 3},
+        {2, 3},
+        {2, 2}
+    };
```
PhilMiller, line 221: I'm not sure if I fully understood what I read, but aren't inner polygon boundaries supposed to run clockwise, to maintain orientation?

test/geopackage/SQLite_Test.cpp
```diff
@@ -0,0 +1,111 @@
...
+        "table_name",
+        "data_type",
+        "identifier",
+        "description",
+        "last_change",
+        "min_x",
+        "min_y",
+        "max_x",
+        "max_y",
+        "srs_id"
```
PhilMiller, line 61: Does SQLite's API guarantee that columns are returned in a specified order?

test/geopackage/WKB_Test.cpp
program--: You're correct. That is the case typically (or at least all polygons in a given implementation should orient the same way). Boost doesn't care for orientation (as far as I know), so order is given as read. Most geometry writers will handle orientation and (sometimes) topology validation, so that felt out of scope to handle in our case.

In this test case specifically, we aren't checking for topological correctness, just value correctness.

test/geopackage/WKB_Test.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+    void SetUp() override {
+        this->wkb_bytes["point"] = {
+            0x01,
+            0x01, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x40,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x10, 0x40
+        };
+
+        this->wkb_bytes["point_big"] = {
```
PhilMiller, line 23:
```suggestion
        this->wkb_bytes["point_little_endian"] = {
            0x01,
            0x01, 0x00, 0x00, 0x00,
            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x40,
            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x10, 0x40
        };

        this->wkb_bytes["point_big_endian"] = {
```

test/geopackage/WKB_Test.cpp
PhilMiller: That was exactly what I meant. Now that you've said it, I can see that they're equivalent up to byte-reversal of each piece. I suggested renamings so that it's more self-evident

test/geopackage/WKB_Test.cpp
PhilMiller: Is there anything in the broader application that relies on that orientation? If so, is that being checked at a higher level, or does it expect the WKB (and geojson) readers to ensure the validity of what they return?

program--: If by broader application you mean `ngen`, then no I don't think so. The geometry validity is a hydrofabric responsibility I think, so that isn't something we would need to be concerned with. As far as the validity of outputs from WKB/geopackage/geojson writers, assuming our inputs are the hydrofabric, then we are relying on the validity of GDAL, which I think is good enough to not worry about.

@hellkite500 might have a better answer than me (because I could be completely wrong haha).

test/geopackage/SQLite_Test.cpp
program--: I'm actually not sure if the C library guarantees column ordering -- that's a good question. Intuitively, I think columns are indexed based off the meta table describing the table schema, and these names are accessed using `sqlite_column_name()` with the column index as an argument, so possibly?

src/geopackage/wkb.cpp
program--: Implemented in 894e8ab.

test/geopackage/WKB_Test.cpp
program--: I didn't know that C++ supported scientific notation, thanks! Implemented this in 5419798.

test/geopackage/WKB_Test.cpp
program--: Implemented in 5531a89.

test/geopackage/WKB_Test.cpp
program--: Implemented in 5531a89 based on your suggestion below.

src/geopackage/sqlite/database.cpp
program--: I may have a tendency to do this (on accident) :) but I guess it's a good thing I'm on the side of prefacing all members with `this->` 😛

src/geopackage/sqlite/database.cpp
program--: Maybe https://github.com/gabime/spdlog could be something we look at? Not sure if there's any better logging implementations, but that's the one I'm familiar with (and it's MIT license + header only)

src/geopackage/read.cpp
program--: I've implemented the two-layer white-list check in 8cbd253. Before the `has_table` call, a check for `sqlite_` and invalid characters is done in a call to the function `check_table_name`.

src/geopackage/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+    message(WARNING "[NGen::geopackage] wkb.cpp cannot be optimized with "
+                    "${CMAKE_CXX_COMPILER_ID} due to a suspected optimizer/boost issue:\n"
+                    "https://github.com/NOAA-OWP/ngen/issues/567.\n"
+                    "Use IntelLLVM if optimization for this source file is required.")
+    # !! Required due to optimizer issue with either clang or
+    # !! boost::geometry::srs::transformation
+    set_source_files_properties(wkb.cpp PROPERTIES COMPILE_FLAGS -O0)
+endif()
+
+add_library(geopackage STATIC geometry.cpp
```
PhilMiller, line 15: Is there a reason to specify `STATIC`?

program--: No specific reason, I think I just copied what the geojson CMakeLists had. If it's better to put a different option then I'll change it to that!

PhilMiller: You can just drop it, and the build will produce whatever CMake has been configured to do in that build.

PhilMiller:
To the extent that I'm qualified to review this, it looks good to me. The code seems to do what it says on the label, and is written well.

src/geopackage/read.cpp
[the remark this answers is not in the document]
hellkite500: I see, I just wasn't expanding the string in my head enough....really need a white board or chalk board next to my computer haha

include/geopackage/WKB.hpp
hellkite500: We can postulate this in an issue as a future enhancement if we don't want to refactor it on this PR.

test/geopackage/WKB_Test.cpp
hellkite500: This is an interesting question, especially when it comes to topology validation.  I can definitely see a use case that takes a catchment polygon (or set of polygons) and wants to do something like create a dynamic mesh using the boundaries as pilot points.  In this case, topology is quite important.  Nothing in our current and short/medium term goals, however, is going to be doing something that sophisticated.  

Assuming the inputs are always generated by a specific workflow/tool isn't a great assumption either.  The general intent is that anyone can generate/produce a hydrofabric of their liking and we can simulate on it.  I would say handling this validation is likely out of scope for this PR, but it shouldn't be entirely ignored.  I think some in code comments stating the assumptions are worthwhile, and probably an issue to track/document our ultimate decision on how much validation we perform is warranted.

include/geopackage/WKB.hpp
PhilMiller: Let's leave it alone, and consider it if we need to account for more than the one or two cases we have now.

program--: This might also factor into #567 somewhat due to the transformation calls.

I think in general we will only have the 2 cases: (1) geometry is in EPSG:4326, and (2) geometry is not in EPSG:4326.

Though, this brings up another point of if we want to work in EPSG:4326 in-memory, or transform to a numerically better projection for our expected domain (i.e. EPSG:5070 for CONUS). There's probably a lot more to consider with that though, so I agree making an issue is the better approach for this.

test/geopackage/WKB_Test.cpp
program--: I think creating an issue on this topic is worthwhile.

My assumption is less on whether its generated by a specific workflow/tool, but more of if the inputs are generated using _standard-conforming_ writers and representations.

If the inputted geometry is already invalid, that shouldn't be an `ngen` responsibility, since that requires modifying the geometry before processing, which also requires inferring the user's intent (i.e. we fix a divide's topology, but now its covering is inaccurate -- in this case a warning might be useful).

However, if _we_ are generating geometry, then I think that is a different conversation. At that point, validation should be considered (if not necessary).

Regardless, I think it's something worth discussing though! 😄

PhilMiller: Given the breadth of possible sources of data, I'm strongly inclined toward saying that we should implement internal validation of whatever we're reading, and enforce it sooner rather than later. Otherwise, we're going to end up committed to supporting those invalid inputs, even as we want to do things that would benefit from stricter assumptions, and so we'll have to write clean-up code.

None of that argues for doing it in this PR, though. Please open an issue for it and link it here.

PhilMiller: Also, please make these test cases strictly valid inputs

program--: updated in bd30664.

hellkite500: Trust but validate....I think they call that `0 trust` these days haha

hellkite500 (approved):

PR #524, opened 2023-05-09
Add stdexcept include to UnitsHelper header

### LINKED ISSUE #523: UnitsHelper source/header requires additional headers with libstdc++ version 13.1.0
`libstdc++` (GNU C++) version 13 raises issues with string comparison, and std::runtime_error not being found, when compiling target `core_mediator`. `libc++` (LLVM C++) compiles as expected.

## Current behavior
```
...
[build] ngen/src/core/mediator/UnitsHelper.cpp:42:24: error: no member named 'runtime_error' in namespace 'std'
[build]             throw std::runtime_error("Unable to convert " + in_units + " to " + out_units);
[build]                   ~~~~~^
[build] ngen/src/core/mediator/UnitsHelper.cpp:60:17: error: invalid operands to binary expression ('const std::string' (aka 'const basic_string<char>') and 'const std::string')
[build]     if(in_units == out_units){
[build]        ~~~~~~~~ ^  ~~~~~~~~~
...
```

## Expected behavior
The `core_mediator` target should compile.

> Assumes CMake project is already configured, and the system default is to use `libstdc++` with **clang** or **icx**.

```sh
# This should fail
cmake --build build --target core_mediator
```

Add the following at around `CMakeLists.txt:47` to force `libc++`
```cmake
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")
```
and rerun
```sh
# This should compile
cmake --build build --target core_mediator
```

## Possible Solution
Add an include directive for `<stdexcept>` (which also includes `<string>`) to `include/core/mediator/UnitsHelper.hpp`.

I'm guessing either `<map>`, `<memory>`, or `<mutex>` stopped using `<string>` and `<stdexcept>` so they aren't included by association anymore, but still are in `libc++`. (Unless this is only an issue in my environment...)

program--:
This PR resolves #523 by including `<stdexcept>` in `UnitsHelper.hpp`. In `libstdc++` version 13.1.0, the `<mutex>` header removed an include to `<system_error>`, which also removed includes to `<stdexcept>` and `<string>`. This caused `std::string` and `std::runtime_error` to raise errors on compilation of target `core_mediator`.

## Additions

- Adds `<stdexcept>` to `include/core/mediator/UnitsHelper.hpp`.

## Testing

- `core_mediator` compiles successfully and all unit tests pass locally using `libstdc++` and `libc++` with GCC version 13.1.0, clang 15.0.7, and icx 2023.

## Checklist

- [x] Linux

### DIFF
--- a/include/core/mediator/UnitsHelper.hpp
+++ b/include/core/mediator/UnitsHelper.hpp
@@ -2,6 +2,7 @@
 #include <map>
 #include <memory>
 #include <mutex>
+#include <stdexcept>
 #include "all.h"
 
 #ifndef NGEN_UNITSHELPER_H


mattw-nws:
This looks good... can you make sure you're rebased on latest master? From the test results it looks like you're missing [a5d1102](https://github.com/NOAA-OWP/ngen/commit/a5d110259964df3f0a1d767ac3619c5cb8a0b5f7).

mattw-nws:
> This looks good... can you make sure you're rebased on latest master? From the test results it looks like you're missing [a5d1102](https://github.com/NOAA-OWP/ngen/commit/a5d110259964df3f0a1d767ac3619c5cb8a0b5f7).

Strike that... that's not it... rather it looks like ce0ce46 broke this test... that's odd... @hellkite500 can you double-check that?

hellkite500:
Looking at error that has cropped up and doing a little digging, this seems to be an upstream bug/issue in the backend arrow library

https://github.com/apache/arrow/issues/15054

mattw-nws:
> Looking at error that has cropped up and doing a little digging, this seems to be an upstream bug/issue in the backend arrow library
> 
> [apache/arrow#15054](https://github.com/apache/arrow/issues/15054)

So, the segfault, yes... I hadn't noticed that, actually, but that's definitely odd and likely the above issue.

However... the preceeding message is indicative of it not finding the gpkg file, which I think worked until  c6509cb ... I'm pretty sure the workflow runs the test at the repo root and not in `cmake_build/test`, which would make it not find the gpkg at a `../../` location. IOW, the test would still fail without the segfault.

Hopefully if it finds the file and doesn't exit on that path (probably not closing something in a `finally` block, which triggers the pyarrow crash), the test will succeed.

mattw-nws:
Okay, you should be able to rebase on master now and we can get this merged.

mattw-nws (approved):

PR #528, opened 2023-05-17
Fixing test_routing_pybind workflow failures

mattw-nws:
See [comment in #519](https://github.com/NOAA-OWP/ngen/pull/519#issuecomment-1551370076) and discussion in #524 , rolls back some changes in 7550dfc to fix failing tests, and excludes pyarrow 12.0.0 because of apache/arrow#15054 (v11.0.0 doesn't have the behavior and hopefully 13.0.0 will fix it).

## Additions

-

## Removals

-

## Changes

- rolled back pathing changes in 7550dfc in `ngen_routing_config_unit_test.yaml`
- updated workflow YAML to exclude pyarrow version 12.0.0 specifically

## Testing

1.

## Notes

-

## Todos

- Make sure when pyarrow 13.0.0 is released that the segfault bug goes away.

## Checklist

1.

- [X] Linux
- [X] macOS

### DIFF
--- a/.github/workflows/module_integration.yml
+++ b/.github/workflows/module_integration.yml
@@ -145,7 +145,7 @@ jobs:
           pip install 'pip>=23.0,<23.1'
           pip install -U setuptools
           pip install -r requirements.txt
-          pip install deprecated pyarrow tables geopandas numpy
+          pip install deprecated 'pyarrow!=12.0.0' tables geopandas numpy
           if [ ${{ runner.os }} == 'macOS' ] 
           then
             export LIBRARY_PATH=/usr/local/lib/gcc/11/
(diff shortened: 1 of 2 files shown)


program-- (approved):
This looks good to me!

I think this brings up something we need to account for when dealing with config files -- if `ngen` is available via PATH, then we may want to handle configuration paths as relative to the configuration file (as opposed to relative to the user's cwd), or at least document a defined behavior (if it is not documented already), but that's for a different discussion :)

mattw-nws:
Yes, it would be particularly nice for configuration files with paths in them to be able to handle paths as relative to the parent file--in this case you'd have to file that issue with NOAA-OWP/t-route 😄 . Interestingly, however, we also generate realization config "files" as strings in some of our tests... so that becomes an edge case that would add some complexity.

PR #533, opened 2023-06-12
In-memory, multi-dimensional structure

program--:
This PR provides an in-memory structure for storing multi-dimensional unstructured/structured grid data. This structure is similar to NetCDF's in-memory format, but provides developer-friendly semantics (i.e. creating a grid scheme, without immediately writing to disk, which is what NetCDF does).

The goal of this implementation is to provide an IR between model outputs and disk output, such that model outputs will be put into this MD structure, then the MD structure is used to write out to a file or endpoint, in a generic sense.

## Additions
This PR uses the `ngen` top-level namespace.

### `ngen::mdarray`

This `mdarray` implementation represents an *N*-dimensional array using a **stride**-based backing vector in column-major form (the first dimension is contiguous).

This class implements the following interface:
- `ngen::mdarray::at` - Access a value in an `mdarray`.
- `ngen::mdarray::insert` - Insert a value into an `mdarray`.
- `ngen::mdarray::allocate` - Allocates space within the `mdarray` using `std::vector::resize`.
- `ngen::mdarray::size` - Returns the total size of the `mdarray` (aka the product of all dimensions).
- `ngen::mdarray::max_size` - Effectively the same as `ngen::mdarray::size` currently.
- `ngen::mdarray::begin` - Iterator at the beginning of the `mdarray`.
- `ngen::mdarray::end` - Iterator at the end of the `mdarray`.
- `ngen::mdarray::rank` - Returns the rank of the `mdarray` (aka number of dimensions).
- `ngen::mdarray::shape` - Returns the dimensions of the `mdarray`.
- `ngen::mdarray::index` - Indexes a given list of multi-dimensional indices into a stride address. (i.e. `{ X1, X2, X3, ... } -> Y`)
- `ngen::mdarray::deindex` - Converts a stride address into a list of multi-dimensional indices. (i.e. `Y -> { X1, X2, X3, ... }`)

### `ngen::mdframe`

The `mdframe` implementation is a structure containing multiple **labeled** dimensions and `mdarray`'s. This structure (somewhat) mimics the conceptual format of NetCDF. In particular, we can think of an `mdframe` as a *multimap* between dimension tuples to multi-dimensional arrays. In other words, it is similar to an R data.frame or pandas dataframe where the column types are multi-dimensional arrays.

Consider the dimensions: x, y, z; and the variables:
- `v1<float>(x)`
- `v2<double>(x, y)`
- `v3<int>(x, y, z)`

Then, it follows that the corresponding `mdframe` (spanned over x):

  x |  v1 |      v2 |      v3
--- | --- | ------- | -------
  0 | f_0 | \[...\]\_0 | \[\[...\]\]\_0
  1 | f_1 | \[...\]\_1 | \[\[...\]\]\_1
... | ... |  ...    |   ...
  n | f_n | \[...\]\_n | \[\[...\]\]\_n

where:
- `v1: Vector of floats,   rank 1 -> [...]`
- `v2: Matrix of doubles,  rank 2 -> [[...]]`
- `v3: Tensor of integers, rank 3 -> [[[...]]]`

Alternatively, we can project down to a 2D representation by
unpacking the dimensions, such that:

```
     dimensions ||                variables
=============== || ========================
  x |   y |   z ||   v1 |     v2 |       v3
--- | --- | --- || ---- | ------ | --------
  n |   m |   p || s[n] | d[n,m] | i[n,m,p]
```

This class implements the following interface:
- `ngen::mdframe::get_dimension` - Get a dimension.
- `ngen::mdframe::has_dimension` - Check for a dimension.
- `ngen::mdframe::add_dimension` - Add a dimension.
- `ngen::mdframe::get_variable` - Get a variable by name.
- `ngen::mdframe::operator[]` - Same as `ngen::mdframe::get_variable`.
- `ngen::mdframe::has_variable` - Check for a variable.
- `ngen::mdframe::add_variable` - Add a variable.
- `ngen::mdframe::to_csv` - Handler for writing to CSV.
- `ngen::mdframe::to_netcdf` - Handler for writing to NetCDF.

## Testing

- All unit tests pass locally using GCC 13.1.1, Clang 15.0.7, and ICPX 2023.1.0.

## Todos

- CSV Handler may need a logic refactor. It correctly works (as far as the unit test), so it may not be necessary, but just a thought to keep in mind.

## Checklist

- [x] Linux

### DIFF
--- a/include/utilities/mdframe/variable.hpp
+++ b/include/utilities/mdframe/variable.hpp
@@ -265,6 +265,6 @@ struct variable {
 
 } // namespace detail
 
-} // namespace io
+} // namespace ngen
 
 #endif // NGEN_IO_MDFRAME_VARIABLE_HPP
--- a/include/utilities/traits.hpp
+++ b/include/utilities/traits.hpp
@@ -104,6 +104,6 @@ struct type_list{
 };
 
 } // namespace traits
-} // namespace io
+} // namespace ngen
 
 #endif // NGEN_IO_TRAITS_HPP


PhilMiller:
Quick suggestion to address your note about row vs column major order confusion: Have a unit test that creates `mdarray` instances of various dimensions, of extent 2 in each dimension. Assert for each that `a.index(0,0,...) +1 == a.index(1,0,0,...)`.

program--:
> Quick suggestion to address your note about row vs column major order confusion: Have a unit test that creates `mdarray` instances of various dimensions, of extent 2 in each dimension. Assert for each that `a.index(0,0,...) +1 == a.index(1,0,0,...)`.

This is now added as part of commit 39bc4d8, thanks!

PhilMiller:
A high-level recommendation - don't use a super-generic name like `io` as the namespace name. Ideally, we'd have a top-level namespace for the project, like `namespace ngen` (which is also not a great name), and then stuff living inside it. The expedient resolution would be naming it something like `namespace ngen_io` for now.

src/io/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+endif()
+
+add_library(io ${io_sources})
+add_library(NGen::io ALIAS io)
+target_include_directories(io PUBLIC ${PROJECT_SOURCE_DIR}/include/io)
+target_link_libraries(io Boost::boost)
+
+if(NETCDF_ACTIVE)
+    target_link_libraries(io ${NETCDF_LIBRARIES})
+endif()
```
PhilMiller, line 14: Missing EOF new-line. Git complains about this because lots of utilities handle this poorly -- e.g. try `cat src/io/CMakeLists.txt` and notice that your prompt is printed right after the `endif()`

include/io/mdarray.hpp
```diff
@@ -0,0 +1,274 @@
...
+     *
+     * @example
+     * mdarray<int> x(2); // rank 2
+     * x.at({0, 0});      // get the value at (0, 0) in a 2D array.
+     * x.at({1, 1}) = 3;  // set the value at (1, 1) in a 2D array.
+     *
+     * @param n Index list
+     * @return reference 
+     */
+    reference at(const ilist& n)
```
PhilMiller, line 49: I'm bothered by this using `std::vector` as the argument type for indexing, because it means that most every call site is going to have to construct a temporary to pass, including heap allocation and de-allocation.

I don't immediately have a concrete suggestion for a complete replacement, but one possibility would be taking `std::initializer_list<size_t>`

I *think* that if we had `std::span` (alas, C++20) or an analogue available, we could just take that, and both `vector`s and braced initializer lists would be accepted, with no notable overhead.

I plan on proposing the import of `boost::span` into our codebase to better support passing data between models and layers of the framework, so I'm glad to see there's another prospective use case.

include/io/mdarray.hpp
```diff
@@ -0,0 +1,274 @@
...
+
+        size_type index = this->index(n);
+    
+        if (index > this->max_size())
+            throw std::out_of_range("out of range");
+        
+        if (index > this->size() || this->size() == 0)
+            this->m_data.resize(index + 1);
+    
+        this->m_data.at(index) = value;
```
PhilMiller, line 84:
```suggestion
        this->m_data[index] = value;
```
There's no reason to ask the underlying `vector` to repeat the bounds-check we ensured would pass just above.

include/io/mdarray.hpp
```diff
@@ -0,0 +1,274 @@
...
+    const_reference at(const ilist& n) const
+    {
+        return this->m_data.at(this->index(n));
+    }
+
+    void insert(const ilist& n, value_type value)
+    {
+        for (size_type i = 0; i < this->m_shape.size(); i++) {
+            if (n[i] >= this->m_shape[i])
+                throw std::out_of_range("out of range");
```
PhilMiller, line 73: The two `out_of_range` exceptions should report different messages. Ideally, they would include the actual values in question - it's an error case, so we probably don't need to worry about their efficiency in assembling those strings.

include/io/mdarray.hpp
```diff
@@ -0,0 +1,274 @@
...
+    {
+        return this->m_shape.size();
+    }
+
+    /**
+     * Get the shape of this mdarray.
+     * 
+     * @return ilist
+     */
+    ilist shape() const noexcept
```
PhilMiller, line 150: This is also going to force allocation of a short-lived `vector` on the callers. Ideally, we'd return `span<const size_t>`. I doubt callers would have life-time/aliasing issues with that in a case where `m_shape` changes.

include/io/mdarray.hpp
```diff
@@ -0,0 +1,274 @@
...
+     * 
+     * @param idx Address index (aka flat index)
+     * @return ilist 
+     */
+    ilist deindex(size_type idx) const
+    {
+        if (idx == 0) { // trivial case
+            return ilist(this->rank(), 0);
+        }
+
```
PhilMiller, line 184:
```suggestion
```
This isn't a special case in which the general code behaves incorrectly. We're not really saving anything performance-wise by having this, so it's just added code complexity.

src/io/mdframe_handler_netcdf.cpp
```diff
@@ -0,0 +1,75 @@
...
+ */
+MDARRAY_VISITOR(mdarray_putvar, void)
+{
+    MDARRAY_VISITOR_TEMPLATE_IMPL(T& v, netCDF::NcVar& var) -> void
+    {
+        for (auto it = v.begin(); it != v.end(); it++) {
+            var.putVar(it.mdindex(), *it);
+        }
+    }
+};
```
PhilMiller, line 20: Is this the only use of `mdarray::iterator`? Are any others anticipated?

If it's the only one, I think I'd prefer just a loop over the linear index range, and explicit calls to `deindex`

include/io/mdarray.hpp
PhilMiller: Or, the caller could pass in a `span` that they want filled in.

include/io/mdarray.hpp
```diff
@@ -0,0 +1,274 @@
...
+        return index;
+    }
+
+    /**
+     * Retrieve the multi-dimensional index from a given address index.
+     * 
+     * @param idx Address index (aka flat index)
+     * @return ilist 
+     */
+    ilist deindex(size_type idx) const
```
PhilMiller, line 179: In this case, we should definitely be taking an 'output argument' that we'll write the result into, rather than constructing and returning a temporary `vector`. The use in `iterator::mdindex()` called in a loop by `mdarray_putvar` will be painful otherwise, since it will do that for every single element

include/io/mdarray.hpp
program--: I think, when this member is used, the vector argument will almost always be constructed before calling. The tests don't necessarily reflect that, but the handlers (for example) do.

I agree that a `std::span` impl would be better served here, but I'm not sure that a `std::initializer_list<size_t>` would work well here (since this prevents a vector from being passed in, I think?)

include/io/mdarray.hpp
program--: Thanks! I always forget that `.at` does a bounds check.

include/io/mdarray.hpp
program--: Ah, I think this is something I forgot to fix. I realize I can return a reference, since `shape` is a member variable also I think.

include/io/mdarray.hpp
mattw-nws: See also #576 , since this is still open.

CMakeLists.txt
```diff
@@ -111,7 +111,7 @@ if(LSTM_TORCH_LIB_ACTIVE)
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${TORCH_CXX_FLAGS}")
 endif()
 
-find_package(Boost 1.72.0 REQUIRED)
+find_package(Boost 1.72.0 REQUIRED COMPONENTS filesystem)
```
hellkite500, line 114: Just noticed this come through and I wanted to weigh in on the inclusion of `boost:: filesystem`. So far we have been careful to use header only boost components, and the discussion of `boost:: filesystem` in the past led us to steer clear of it to keep the binary dependency out of the equation. We can revisit this topic, but I wanted to get this warning up before we got to invested.

src/io/mdframe_handler_netcdf.cpp
program--: > Note: this chunk of code is now in `include/utilities/mdframe/handler_netcdf.hpp`

Currently, I think it is, but I think there is a use-case for iterator-based access if the `mdarray` is being used outside of the context of an output handler. I haven't gotten a chance to _find_ that use-case, but I'm currently expecting it

On the other hand, it's quite possible that use-case never appears, and if that's the case then I'll change this over as necessary

include/utilities/mdarray/mdarray.hpp
```diff
@@ -0,0 +1,206 @@
...
+     * Index a multi-dimensional set of indices to a single address index.
+     * 
+     * @param n 
+     * @return size_type 
+     */
+    size_type index(const boost::span<const size_t> n) const
+    {
+        size_type index = 0, stride = 1;
+        for (size_type k = 0; k < this->rank(); k++) {
+            index += n[k] * stride;
```
PhilMiller, line 176: Should there be an error check of `n[k] < m_shape[k]`? Could just be `assert` for now, to automatically optimize it out with `-DNDEBUG`

include/utilities/mdarray/mdarray.hpp
```diff
@@ -0,0 +1,206 @@
...
+        size_type stride = 1;
+        for (size_type k = 0; k < this->rank(); k++) {
+            n[k] = idx / stride % this->m_shape[k];
+            stride *= this->m_shape[k];
+        }
+    }
+
+  private:
+    std::vector<size_type> m_shape;
+    container_type         m_data;
```
PhilMiller, line 201: I'm a bit lost about the relationship between `m_data.size()` and the values in `m_shape`. The shape is fixed at construction, while `m_data` can grow or shrink arbitrarily with calls to `allocate`. Could you add assertions of the invariants that are actually supposed to hold, and maybe document the enforced relationship in the comments of whichever of `size`, `max_size`, or `allocate` makes sense?

include/utilities/cartesian.hpp
```diff
@@ -0,0 +1,39 @@
...
+#define NGEN_UTILITIES_CARTESIAN_HPP
+
+#include <vector>
+
+#include "span.hpp"
+
+namespace ngen {
+
+/**
+ * @brief
```
PhilMiller, line 11: Without scribbling out a lot of algebra, I can't directly work out what this function does. Please document it more extensively.

I'm also concerned about the mixed iterative/recursive structure, as well as the combinatorial size of output that it implies.

Style-wise, it's writing to elements of `index` and extending `output`. It looks like `index` is just a bit of scratch memory for use during the overall operation. Given that, it would be preferable to rename this routine something like `cartesian_indices_helper`, and have a `cartesian_indices` routine that allocates `index` internally and passes it down, so that the caller doesn't need to. That split could potentially also be an opportunity address some of my concerns about the mixed iterative/recursive structure.

PhilMiller: Finally, as a big alternative to actually materializing a container that holds all of the indices, you could make more of a generator object the can spit out the indices one by one, without first generating and storing all of them.

program--: > This comment is just for context

This function takes in a `shape` span, (i.e. `{3, 1, 2}`), a buffer `index`, the current `dimension`, and the actual `output`.

In general what this does is generates all the possible indices that can occur from the given `shape`. For example, calling `cartesian_indices` with the above shape would result in:
```
output = {
  { 0, 0, 0 },
  { 0, 0, 1 },
  { 1, 0, 0 },
  { 1, 0, 1 },
  { 2, 0, 0 },
  { 2, 0, 1 }
}
```

I agree that there's probably a better way, and a generator object may be the way to go (especially as the `mdframe` contents get larger)

include/utilities/mdframe/handler_csv.hpp
```diff
@@ -0,0 +1,102 @@
...
+#define NGEN_MDFRAME_HANDLERS_CSV_HPP
+
+#include <fstream>
+
+#include "mdframe.hpp"
+#include "cartesian.hpp"
+
+namespace ngen {
+
+inline void mdframe::to_csv(const std::string& path, bool header) const
```
PhilMiller, line 11: The implementations of this and `to_netcdf` should be in .cpp files under `src/`, not in headers. Among other things, that would push includes of `cartesian.hpp` and `<netcdf>` out of headers being included by other code.

include/utilities/mdarray/mdarray.hpp
program--: An incoming commit handles this by only providing `size` in the public interface. On construction, an `mdarray` will allocate as much space as it needs (via the `std::vector` size constructor) and we'll leave optimizing this to a future concern. The `allocate` member function is removed, and the `max_size` member is converted into a private member inlined at construction.

include/utilities/cartesian.hpp
program--: After working on this for a bit, I don't think there's a good solution for splitting the iteration and recursive components (at least not from what I could see). Due to the cartesian product, there will be a need for iteration (in this case, we iterate to make recursive calls), and the recursion makes it succinct. Since this is an output handler (and _should_ only be called at the very end of a run), I'm not sure that refactoring the inner details is worth it.

> One thing to note about `output` as well is that its size is known before this function is ever called (since it's just the product of the elements of `shape`). So, in theory, we could handle it outside of this if it is a concern.

However, modifying it so that only the shape and output vector of vectors are needed as inputs is possible (which I'll add onto the next push).

include/utilities/mdarray/iterator.hpp
```diff
@@ -0,0 +1,81 @@
...
+    using difference_type   = std::ptrdiff_t;
+    using value_type        = mdarray::value_type;
+    using pointer           = mdarray::pointer;
+    using reference         = mdarray::reference;
+
+    iterator(const mdarray& ref, size_type idx)
+        : m_ref(ref)
+        , m_idx(idx){};
+
+    const_reference operator*() const
```
PhilMiller, line 21: I'm a tad surprised `const_reference` compiles - it's not defined above, so I assume it's pulled in from the enclosing `class mdarray`

PhilMiller: And I checked that it is used in the test, below

PhilMiller: It would probably be good to define it explicitly in here, too.

include/utilities/mdarray/mdarray.hpp
```diff
@@ -0,0 +1,234 @@
...
+
+    /**
+     * Retrieve the multi-dimensional index from a given address index.
+     * 
+     * @param[in] idx Address index (aka flat index)
+     * @param[out] n outputted index
+     */
+    void deindex(size_type idx, boost::span<size_type> n) const
+    {
+        // TODO: Need a bounds check?
```
PhilMiller, line 202: I think `assert(n.size() == this->rank())`?

PhilMiller: Ah, and `idx < this->size()`, like in `index()`

include/utilities/mdframe/dimension.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+        static std::size_t apply(const dimension& d) noexcept
+        {
+            return dimension::hash{}(d);
+        }
+    };
+
+    dimension()
+        : m_name()
+        , m_size() {};
```
PhilMiller, line 35:
```suggestion
    dimension() = default;
```

include/utilities/mdframe/dimension.hpp
```diff
@@ -0,0 +1,86 @@
...
+        : m_name(name)
+        , m_size(static_cast<std::size_t>(-1)) {};
+    
+    dimension(const std::string& name, std::size_t size)
+        : m_name(name)
+        , m_size(size) {};
+
+    dimension(const dimension& d)
+        : m_name(d.m_name)
+        , m_size(d.m_size) {};
```
PhilMiller, line 47:
```suggestion
    dimension(const dimension& d) = default;
```

include/utilities/mdframe/dimension.hpp
```diff
@@ -0,0 +1,86 @@
...
+        this->m_name = d.m_name;
+        this->m_size = d.m_size;
+        return *this;
+    }
+
+    dimension& operator=(dimension&& d) {
+        this->m_name = std::move(d.m_name);
+        this->m_size = std::move(d.m_size);
+        return *this;
+    }
```
PhilMiller, line 63:
```suggestion
    dimension(dimension&& d) = default;
    dimension& operator=(const dimension& d) = default;
    dimension& operator=(dimension&& d) = default;
```

include/utilities/traits.hpp
```diff
@@ -0,0 +1,109 @@
...
+     *
+     * For example, for a type_list<int, std::string> with @c{Container} -> std::vector:
+     *
+     *   boost::variant<std::vector<int>, std::vector<std::string>>
+     * 
+     * @tparam Container Type of container to hold
+     *                   each type in within the variant.
+     */
+    template<template<typename> class Container>
+    using variant_container = boost::variant<Container<Ts>...>;
```
PhilMiller, line 84: Funny I should see this here - I was just thinking about wanting exactly such a thing in #593!

include/utilities/mdframe/dimension.hpp
```diff
@@ -0,0 +1,86 @@
...
+    
+    bool operator==(const dimension& rhs) const
```
PhilMiller, line 65: Why compare hashes of `m_name` rather than `m_name` directly? The cost is actually higher, since the hash is reading the whole name and computing something from it.

Same applies to `variable` and anywhere else this pattern appears

include/utilities/mdframe/visitors.hpp
```diff
@@ -0,0 +1,96 @@
...
+    std::string operator()(const T& v) const
+    {
+        return std::to_string(v);
+    }
+};
+
+
+} // namespace visitors
+} // namespace detail
+} // namespace io
```
PhilMiller, line 94:
```suggestion
} // namespace ngen
```

include/utilities/mdframe/mdframe.hpp
```diff
@@ -0,0 +1,273 @@
...
+    void to_csv(const std::string& path, bool header = true) const;
+
+#if NETCDF_ACTIVE
+    /**
+     * Write this mdframe to a NetCDF file.
+     * 
+     * @param path File path to the output NetCDF
+     */
+    void to_netcdf(const std::string& path) const;
+#endif
```
PhilMiller, line 264: For purposes of things like ccache, it's slightly preferable to not conditionalize the declaration, and either leave the definition out, so linking fails, or have it error at runtime.

Not necessary to change for this PR, though.

src/utilities/mdframe/handler_netcdf.cpp
```diff
@@ -0,0 +1,78 @@
...
+    netCDF::NcFile output{path, netCDF::NcFile::replace};
+
+    std::unordered_map<std::string, netCDF::NcDim> dimmap;
+    std::unordered_map<std::string, netCDF::NcVar> varmap;
+
+    for (const auto& dim : this->m_dimensions)
+        dimmap[dim.name()] = output.addDim(dim.name(), dim.size());
+
+    for (const auto& pair : this->m_variables) {
+        netCDF::NcType* type = nullptr;
```
PhilMiller, line 41: What's with `type` being a pointer? All the use looks suitable for just being `netCDF::NcType`.

Its documentation looks perfectly safe to copy those around; is there some subtlety to `netCDF::NcFile::addVar` that distinguishes instances of these type objects, such that a copy of `ncInt` isn't equivalent to the original?

PhilMiller: (I've only ever use netCDF from Fortran code)

src/utilities/mdframe/handler_csv.cpp
```diff
@@ -0,0 +1,130 @@
...
+            boost::span<size_type> index_view{ index_buffer.data(), var.rank() };
+            boost::span<size_type> vd_view{vd_index[var.name()]};
+            for (size_type i = 0; i < var.rank(); i++)
+                index_view[i] = index.at(vd_view[i]);
+
+            decltype(auto) value = var.at(index_view);
+            output_line += value.apply_visitor(visitor) + ",";
+        }
+
+        output_line.pop_back();
```
PhilMiller, line 123: This is probably irrelevant, but I think there's an edge case here if `variable_subset` is empty, and hence there's no final `,` to be popped off the line

PhilMiller: I guess you could guard against that by adding an early return above the loop over `indices`, if you care to.

src/utilities/mdframe/handler_netcdf.cpp
program--: This was so I could set `type` once instead of having an `if-elseif` branch that called `addVar` in each branch. `NcType` is polymorphic, and `type` here is set to point to the already initialized version of the different `NcTypes` (i.e. see `netCDF::NcInt` vs `netCDF::ncInt`).

src/utilities/mdframe/handler_csv.cpp
program--: Good catch, a logic error should probably be thrown in that case, since if `variable_subset` is empty, that means there are no output variables.

PhilMiller: Is that an error case, or just a degenerate case?

src/utilities/mdframe/handler_netcdf.cpp
PhilMiller: Ok, I see now. Assigning an arbitrary value of `NcType type` might not be valid code. Thanks.

src/utilities/mdframe/handler_csv.cpp
program--: I would assume an error case (though maybe that's the wrong assumption), since if there are no output variables, then there is no contents in the resulting CSV.

include/utilities/mdframe/mdframe.hpp
program--: Okay, that makes sense -- I wasn't sure if there were any issues with this. Either in this PR or a future one, I'll move the conditional to the source file so that if NetCDF isn't available, then this function will throw a runtime error.

include/utilities/mdframe/dimension.hpp
program--: I think my initial thought was that the hash would be derived from not only `m_name`, but also the dimension size, but it probably makes more sense semantically to derive equality from only dimension names, since we don't want multiple dimensions with the same name in an `mdframe`.

include/utilities/mdframe/dimension.hpp
program--: Good catch, 🤦🏾 I guess I didn't realize these (and the other suggested commits above) were default constructors initially.

CMakeLists.txt
```diff
@@ -85,6 +85,9 @@ if(LSTM_TORCH_LIB_ACTIVE)
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${TORCH_CXX_FLAGS}")
 endif()
 
+find_package(Boost 1.72.0 REQUIRED)
+
+
```
PhilMiller, line 90: Looks like a conflict resolution artifact - this got moved up to to line 76 to be next to the Boost variable handling

program--: Shoot, thought I handled them correctly. Thanks for the catch 🤦🏾 haha

include/utilities/mdframe/variable.hpp
```diff
@@ -0,0 +1,270 @@
...
+    // multi-dimensional vector associated with this variable
+    mdarray_variant m_data;
+
+    // References to dimensions that this variable spans
+    std::vector<dimension> m_dimensions;
+};
+
+} // namespace detail
+
+} // namespace io
```
PhilMiller, line 268:
```suggestion
} // namespace ngen
```

include/utilities/traits.hpp
```diff
@@ -0,0 +1,109 @@
...
+     * supported within this type list.
+     *
+     * @tparam S type to check
+     */
+    template<typename S, typename Tp>
+    using enable_if_supports = std::enable_if_t<is_supported<S>::value, Tp>;
+};
+
+} // namespace traits
+} // namespace io
```
PhilMiller, line 107:
```suggestion
} // namespace ngen
```

PhilMiller (approved):

Issue #441, opened 2022-09-07
NextGen not finding Python packages in virtual environments


peckhams:
## Current behavior

After compiling NextGen on a Mac and successfully running all tests, NexGen is not finding Python packages/modules in active virtual environments (either conda or venv).  The error message is:  ModuleNotFoundError.  This appears to be an instance of a known problem when Python is embedded in an executable via pybind11, as described here:
https://stackoverflow.com/questions/56904149/embedding-python-with-pybind11-virtual-environment-doesnt-work
Another possible solution can be found here:
https://docs.python.org/3.11/library/sys_path_init.html#embedded-python
A work-around is to provide the paths to all Python packages using the environment variable PYTHONPATH.

## Expected behavior

NextGen should be able to find Python packages that have been installed in a virtual environment when it is run with that environment active.  It should not be necessary to set PYTHONPATH.  It should also not be necessary to first install the package in a virtual environment and then compile NextGen with that environment active.

## Setup for the tests

0.  Install NextGen by following these online docs:
    https://noaa-owp.github.io/ngen/index.html
    https://noaa-owp.github.io/ngen/md_doc__d_e_p_e_n_d_e_n_c_i_e_s.html
    https://noaa-owp.github.io/ngen/md__i_n_s_t_a_l_l.html
    https://noaa-owp.github.io/ngen/md_doc__b_u_i_l_d_s__a_n_d__c_m_a_k_e.html
    https://noaa-owp.github.io/ngen/md_doc__b_m_i__m_o_d_e_l_s.html
    Also see the "Install ngen" section of this Google doc:
    https://docs.google.com/document/d/17e6HZtSx9Wn6QHnYaLhoHcaCpwrj0rJ3bTh40lbriBQ/edit#heading=h.lzjs5neq2b50

1.  There is a test project folder called "bmi-python-module" that contains a Python
    package folder called "bmihydro", that contains a BMI model source code module
    called "toybmi" (i.e. toybmi.py) that defines a class called "toybmi".
    Copy it into the folder:  ngen/extern/
    (Note that the tests give the same result with the new LSTM Python package.)

2.  Make sure bmihydro is not installed in the conda base environment:
       (base) % pip list
    If it is, uninstall it with:
       (base) % pip uninstall bmihydro

3.  Check whether the environment variable PYTHONPATH has been set:
       (base) % echo $PYTHONPATH
    If it has, this path will be prepended to Python's "sys.path", which will affect
[...]

PR #540, opened 2023-06-22
Remove heuristics for finding Python and NumPy in favor of baseline find_package

PhilMiller:
The basic CMake FindPython functionality works well in modern enough versions. The heuristics may have predated the ability to specify `find_package(Python ... COMPONENTS ... NumPy)`.

FindPython docs for reference: https://cmake.org/cmake/help/latest/module/FindPython.html

## Additions

N/A

## Removals

N/A

## Changes

- Rely on CMake to pick up the active python installation (system or virtual environment) found by `find_package(Python ...)`
- Update the minimum version of CMake to ensure that will work

## Testing

1. Manual testing on my system with a mix of conformations of virtual environments and packages installed in the system environment

## Screenshots

N/A

## Notes

- Developers on other teams may need to be informed that they'll need to `activate` their virtual environment before configuring with `cmake`, rather than just having `ngen/venv` present

## Checklist

- [x] Update documentation accordingly

- [x] Linux
- [x] macOS

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -214,65 +214,6 @@ endif()
 # If Python set to be active, run steps to make sure it is included properly, and set compiler definition to true
 if (${NGEN_ACTIVATE_PYTHON})
     add_compile_definitions(ACTIVATE_PYTHON=true)
-    # If there is a virtual environment directory then look for several things there
-    if(EXISTS ${PROJECT_SOURCE_DIR}/.venv OR EXISTS ${PROJECT_SOURCE_DIR}/venv)
-        if (EXISTS ${PROJECT_SOURCE_DIR}/.venv)
-            set(PY_VENV_DIR ${PROJECT_SOURCE_DIR}/.venv)
-        else()
-            set(PY_VENV_DIR ${PROJECT_SOURCE_DIR}/venv)
-        endif()
-        # Start with looking for NumPy
-        file(GLOB_RECURSE NP_COMMON_HEADER ${PY_VENV_DIR}/*npy_common.h)
-        get_filename_component(NP_HEADER_DIR ${NP_COMMON_HEADER} DIRECTORY)
-        get_filename_component(NP_INCLUDE_DIR ${NP_HEADER_DIR} DIRECTORY)
-        set(Python_NumPy_INCLUDE_DIR ${NP_INCLUDE_DIR})
-        message(INFO " Found numpy include path  ${Python_NumPy_INCLUDE_DIR}")
-        # Also try to get the real python root
-        message(INFO " Before Python exec  ${PYTHON_EXECUTABLE}")
-        get_filename_component(PYTHON_EXECUTABLE ${PY_VENV_DIR}/bin/python REALPATH)
-        message(INFO " After Python exec  ${PYTHON_EXECUTABLE}")
-        get_filename_component(PYTHON_BIN ${PYTHON_EXECUTABLE} DIRECTORY)
-        get_filename_component(PY_ROOT_DIR ${PYTHON_BIN} DIRECTORY)
-        set(Python_ROOT_DIR "${PY_ROOT_DIR}")
-        message(INFO " Python root dir to search is  ${Python_ROOT_DIR}")
-        set(Python_FIND_VIRTUALENV ONLY)
-
-        # use, i.e. don't skip the full RPATH for the build tree
-        #set(CMAKE_SKIP_BUILD_RPATH FALSE)
-        #set(CMAKE_SKIP_BUILD_RPATH TRUE)
-        #set(CMAKE_INSTALL_RPATH "/Library/Developer/CommandLineTools/Library/Frameworks")
-        #message(INFO " ORIGIN is ${ORIGIN}")
-        #SET(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
-
-        # add the automatically determined parts of the RPATH
-        # which point to directories outside the build tree to the install RPATH
-        #set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
-        #set(CMAKE_INSTALL_RPATH_USE_LINK_PATH FALSE)
-
-        # the RPATH to be used when installing, but only if it's not a system directory
-        #list(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES "${CMAKE_INSTALL_PREFIX}/lib" isSystemDir)
-        #if("${isSystemDir}" STREQUAL "-1")
-        #    set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
-        #endif("${isSystemDir}" STREQUAL "-1")
-
-        # Alternatively, if there is something in the ENV for the user, try that
-    elseif(DEFINED ENV{Python_NumPy_INCLUDE_DIR})
-        set(Python_NumPy_INCLUDE_DIR $ENV{Python_NumPy_INCLUDE_DIR})
-    endif()
-
-#    if(DEFINED ENV{PYTHON_INCLUDE_DIR})
-#        set(PYTHON_INCLUDE_DIR $ENV{PYTHON_INCLUDE_DIR})
-#        message(INFO " Found 'PYTHON_INCLUDE_DIR' from ENV ($ENV{PYTHON_INCLUDE_DIR})")
-#    else()
-#        message(INFO " Did not find 'PYTHON_INCLUDE_DIR' in ENV")
-#    endif()
-#    if(DEFINED ENV{PYTHON_LIBRARIES})
-#        set(PYTHON_LIBRARIES $ENV{PYTHON_LIBRARIES})
-#        message(INFO " Found 'PYTHON_LIBRARIES' from ENV ($ENV{PYTHON_LIBRARIES})")
-#    else()
-#        message(INFO " Did not find 'PYTHON_LIBRARIES' in ENV")
-#    endif()
-#    include_directories(${PYTHON_INCLUDE_DIR})
 
     find_package(Python 3.6.8 REQUIRED COMPONENTS Interpreter Development NumPy)
     #message(INFO " the final Py lib is ${Python_LIBRARY}")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -202,11 +202,11 @@ endif()
 # First, define NGEN_ACTIVATE_PYTHON, if not already defined
 #   Use NGEN_ACTIVATE_PYTHON environment variable if available, or default to true
 if (NOT (DEFINED NGEN_ACTIVATE_PYTHON))
-    message(INFO " NGEN_ACTIVATE_PYTHON not defined")
     if (DEFINED ENV{NGEN_ACTIVATE_PYTHON})
-        message(INFO " set as $ENV{NGEN_ACTIVATE_PYTHON}")
+        message(INFO " NGEN_ACTIVATE_PYTHON set to '$ENV{NGEN_ACTIVATE_PYTHON}' from the environment")
         set(NGEN_ACTIVATE_PYTHON $ENV{NGEN_ACTIVATE_PYTHON})
     else()
+        message(INFO " NGEN_ACTIVATE_PYTHON set to 'true' by default")
         set(NGEN_ACTIVATE_PYTHON true)
     endif()
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.14)
 
 if(MPI_ACTIVE)
    set (CMAKE_CXX_COMPILER "mpicxx")
--- a/doc/BMI_MODELS.md
+++ b/doc/BMI_MODELS.md
@@ -279,7 +279,7 @@ An example implementation for an appropriate BMI model as a **C++** shared libra
 
 ### Enabling Python Integration
 
-Python integration is controlled with the CMake build flag `NGEN_ACTIVATE_PYTHON`, however this currently defaults to "On"--you would need to turn this off if Python is not available in your environment. See [the Dependencies documentation](DEPENDENCIES.md#python-3-libraries) for specifics on Python requirements, but in summary you will need a working Python environment with NumPy installed. You can set up a Python environment anywhere with the usual environment variables, but note that ngen will look for a Python environment in the build directory named `.venv` or `venv` by default. The appropriate Python environment should be active in the shell when ngen is run.
+Python integration is controlled with the CMake build flag `NGEN_ACTIVATE_PYTHON`, however this currently defaults to "On"--you would need to turn this off if Python is not available in your environment. See [the Dependencies documentation](DEPENDENCIES.md#python-3-libraries) for specifics on Python requirements, but in summary you will need a working Python environment with NumPy installed. You can set up a Python environment anywhere with the usual environment variables. The appropriate Python environment should be active in the shell when ngen is run.
 
 For Python BMI models specifically, you will also need to install the [bmipy](https://github.com/csdms/bmi-python) package, which provides a base class for Python BMI models.
 
--- a/doc/DEPENDENCIES.md
+++ b/doc/DEPENDENCIES.md
@@ -10,7 +10,7 @@
 | [Boost (Headers Only)](#boost-headers-only) | external | `1.72.0` | headers only library |
 | [Udunits libraries](https://www.unidata.ucar.edu/software/udunits) | hybrid (external or submodule) | >= 2.0 | Can be installed via package manager; if not installed, will be automatically downloaded as submodule, and build will be attempted. |
 | [MPI](https://www.mpi-forum.org) | external | No current implementation or version requirements | Required for [multi-process distributed execution](DISTRIBUTED_PROCESSING.md) |
-| [Python 3 Libraries](#python-3-libraries) | external | \> `3.6.8` | Can be [excluded](#overriding-python-dependency). Requires ``numpy`` package |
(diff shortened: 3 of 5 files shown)


PhilMiller:
Draft mode until I go over the documentation and update it to match.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.14)
```
mattw-nws, line 1: Hmm... There is this: https://github.com/NOAA-OWP/ngen/blob/ccce9dc1f2915b90719bff4f0f9b061367220cd2/docker/CENTOS_4.8.5_NGEN_RUN.dockerfile#L9 
... That's centos7 with GCC 4.8.5 I think, not centos4, thankfully... but I don't think we can support GCC 4.x anymore... Probably that dockerfile needs to be removed.

doc/BMI_MODELS.md
```diff
@@ -279,7 +279,7 @@ An example implementation for an appropriate BMI model as a **C++** shared libra
...
+Python integration is controlled with the CMake build flag `NGEN_ACTIVATE_PYTHON`, however this currently defaults to "On"--you would need to turn this off if Python is not available in your environment. See [the Dependencies documentation](DEPENDENCIES.md#python-3-libraries) for specifics on Python requirements, but in summary you will need a working Python environment with NumPy installed. You can set up a Python environment anywhere with the usual environment variables. The appropriate Python environment should be active in the shell when ngen is run.
```
mattw-nws, line 282: So, should we pivot this PR just a little and also get rid of this? : https://github.com/NOAA-OWP/ngen/blob/ccce9dc1f2915b90719bff4f0f9b061367220cd2/include/utilities/python/InterpreterUtil.hpp#L217-L234
I have been reluctant to because it could theoretically break someone's working setup. However, for everything there is a season.

CMakeLists.txt
PhilMiller: Need at least GCC 5 for C++14:
https://en.cppreference.com/w/cpp/compiler_support/14

Probably a better idea to go with slightly newer, based on thoroughly explored compiler bugs in this space:
https://kokkos.github.io/kokkos-core-wiki/requirements.html
Kokkos 3.x also required C++14, and set GCC 5.3 as a minimum due to bugs in earlier versions.

doc/BMI_MODELS.md
PhilMiller: Oof, I hadn't figured there was associated code doing the same. I'll take a look, but I'm kinda inclined to be a bit more cautious about that angle.

doc/DEPENDENCIES.md
```diff
@@ -75,7 +75,7 @@ However, a [CMake build system](BUILDS_AND_CMAKE.md#generating-a-build-system) m
 
 ### Version Requirements
 
-Currently, a version of CMake >= `3.12.0` is required.
+Currently, a version of CMake >= `3.14.0` is required.
```
mattw-nws, line 78: Yeah, anywhere we need to support < 3.14? Looks like our on-prem environment (RHEL7) gets 3.17. @hellkite500 ?

PhilMiller: Any such environment would have already reported failed configuration if it had Python support activated, since the `NumPy` component didn't exist in earlier versions.

CMakeLists.txt
PhilMiller: Can we leave that broader cleanup for another issue?

mattw-nws:
General comment, we do still require the NumPy headers (or pybind11 does, I think)... so what's the thought behind removing the documentation of the necessity of installing that?

CMakeLists.txt
PhilMiller: Also, CentOS itself is obsolescent. RedHat sunset the unbadged rebuild a couple years back and replaced it with CentOS Stream, which is quite different from the RHEL-compatible thing. AlmaLinux took up the mantle, but even that's now at risk:
https://www.phoronix.com/news/AlmaLinux-With-RHEL-Source-Path

mattw-nws: Yeah we're painfully aware. Another of the dockerfiles starts with Rocky, and we use Rocky in DMOD.

PhilMiller:
They're part of the pip package, and picked up by CMake without additional guidance:

```
$ find test-venv/lib/python3.11/site-packages/numpy/ | grep '.h$'
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/ndarrayobject.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/utils.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/__multiarray_api.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/npy_1_7_deprecated_api.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/oldnumeric.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/experimental_dtype_api.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/libdivide/libdivide.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/old_defines.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/ufuncobject.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/npy_common.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/npy_cpu.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/ndarraytypes.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/npy_no_deprecated_api.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/npy_3kcompat.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/npy_os.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/npy_endian.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/npy_interrupt.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/_numpyconfig.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/_neighborhood_iterator_imp.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/numpyconfig.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/__ufunc_api.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/arrayscalars.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/random/distributions.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/random/bitgen.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/npy_math.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/halffloat.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/arrayobject.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/_dtype_api.h
test-venv/lib/python3.11/site-packages/numpy//core/include/numpy/noprefix.h
test-venv/lib/python3.11/site-packages/numpy//f2py/src/fortranobject.h
```


```
(venv) Phils-Mac-mini:ngen phil$ cmake -B test-numpy -S . -DBOOST_ROOT=/opt/homebrew/opt/boost\@1.76/
-- The C compiler identification is AppleClang 14.0.0.14000029
-- The CXX compiler identification is AppleClang 14.0.0.14000029
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Compiler: /Library/Developer/CommandLineTools/usr/bin/cc
-- Compiler: /Library/Developer/CommandLineTools/usr/bin/c++
-- No Boost root: 
-- Found Boost: /opt/homebrew/opt/boost@1.76/lib/cmake/Boost-1.76.0/BoostConfig.cmake (found suitable version "1.76.0", minimum required is "1.72.0")  
INFO UDUNITS_ACTIVE not defined
Looking for UDUNITS-2...
-- Could NOT find UDUNITS2 (missing: UDUNITS2_LIBRARY UDUNITS2_INCLUDE) 
-- Updating submodule /Users/phil/Code/noaa/ngen/extern/UDUNITS-2
INFO - Linking libudunits2 to path "/Users/phil/Code/noaa/ngen/test-numpy/UDUNITS-2/build/lib/libudunits2.dylib"
INFO NGEN_ACTIVATE_PYTHON set to 'true' by default
-- Found Python: /Users/phil/Code/noaa/ngen/venv/bin/python3.11 (found suitable version "3.11.4", minimum required is "3.6.8") found components: Interpreter Development NumPy Development.Module Development.Embed 
Python_NumPy_INCLUDE_DIRS = /Users/phil/Code/noaa/ngen/venv/lib/python3.11/site-packages/numpy/core/include
-- pybind11 v2.10.1 
-- Performing Test HAS_FLTO
-- Performing Test HAS_FLTO - Success
-- Performing Test HAS_FLTO_THIN
-- Performing Test HAS_FLTO_THIN - Success
-- Found Boost: /opt/homebrew/opt/boost@1.76/lib/cmake/Boost-1.76.0/BoostConfig.cmake (found version "1.76.0")  
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Found Python: /Users/phil/Code/noaa/ngen/venv/bin/python3.11 (found version "3.11.4") found components: Interpreter 
-- Configuring done (1.6s)
-- Generating done (0.1s)
-- Build files have been written to: /Users/phil/Code/noaa/ngen/test-numpy
```
(I added the print of `Python_NumPy_INCLUDE_DIRS` to demonstrate that it worked)

CMakeLists.txt
mattw-nws: > Can we leave that broader cleanup for another issue?

Yes, at least as far as any CMake minimum version stuff is concerned, based on the comment elsewhere about NumPy needing 3.14 anyway.

doc/BMI_MODELS.md
mattw-nws: Yeah, I've been cautious about removing this also. But with the above documentation change the code referenced above becomes a COMPLETELY undocumented and rather odd behavior. How to reconcile?

CMakeLists.txt
PhilMiller: Opened #544 to do that clean-up

doc/BMI_MODELS.md
mattw-nws: Also related is https://github.com/NOAA-OWP/ngen/issues/441#issuecomment-1318890579

mattw-nws:
> They're part of the pip package, and picked up by CMake without additional guidance...

Interesting. This probably makes sense but I wonder about the plethora of Python setups... we should probably check this out with a conda env before merging at a minimum, and maybe with OS-installed numpy package (e.g. Debian-style). But yes it will probably work fine.

doc/BMI_MODELS.md
PhilMiller: The 'Expected Behavior' in the issue description there is pretty strong. Let's discuss on Monday, after I've had a chance to look more deeply?

CMakeLists.txt
```diff
@@ -202,77 +202,18 @@ endif()
...
-#    endif()
-#    if(DEFINED ENV{PYTHON_LIBRARIES})
-#        set(PYTHON_LIBRARIES $ENV{PYTHON_LIBRARIES})
-#        message(INFO " Found 'PYTHON_LIBRARIES' from ENV ($ENV{PYTHON_LIBRARIES})")
-#    else()
-#        message(INFO " Did not find 'PYTHON_LIBRARIES' in ENV")
-#    endif()
-#    include_directories(${PYTHON_INCLUDE_DIR})
 
     find_package(Python 3.6.8 REQUIRED COMPONENTS Interpreter Development NumPy)
```
robertbartel, line 218: If the build itself isn't detecting virtual environments any more, it wouldn't hurt to test what happens if we try to regenerate a CMake build without the virtual environment still activated.  My main thought there involves IDEs that may do this for you, perhaps automatically whenever a _CMakeLists.txt_ file is edited.  We might want to add something in the docs about trouble scenarios.

robertbartel (changes requested):
There is at least one very small thing that needs adjustment in the docs on the CMake version dependency.  I did respond (separately) to Matt's comment on the possible change in _InterpreterUtil.hpp_ that need consideration. Beyond those, I had a few thoughts on making sure we've tested certain scenarios.

CMakeLists.txt
PhilMiller: My expectation is that CMake's cache of the previous configuration will sustain it. That would need testing, which I'll do.

doc/DEPENDENCIES.md
```diff
@@ -10,7 +10,7 @@
 | [Boost (Headers Only)](#boost-headers-only) | external | `1.72.0` | headers only library |
```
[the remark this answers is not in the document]
PhilMiller: Good catch

CMakeLists.txt
PhilMiller: Tested - configured with virtual environment active, deactivated it, `touch CMakeLists.txt` to force re-running CMake, and ran `make` and `ctest` successfully

PhilMiller: If some IDE is blowing away CMake configuration and regenerating it in full each time, I would see that as a problem.

mattw-nws: Okay, yes, I have done the above in at least one of my `launch.json` files... So, this is an "ergonomics" concern, but there's a fairly ready and sensible solution... it may break people's setups though so since it's a change, the workaround might bear mentioning in documentation somewhere.... "Note that if you are using an IDE to build ngen, you will need to activate any needed venv (or set `VIRTUAL_ENV` and `PATH` environment variables as if it was activated) as part of the build process--see your IDE's documentation for how to accomplish this." or similar.

robertbartel: I'll unresolve this for now for the sake of tracking.  

Indeed, we can't account for every situation, but we can document suggestions and "gotchas" for people to pay attention to.

doc/BMI_MODELS.md
mattw-nws: > I am not sure whether the currently-active Python environment or the build system's Python environment would be used when running a built executable (or worse, a mix of both).

My understanding (possibly flawed) is that currently the build-time impacts of the activated venv are:
1. The Python runtime for that Python version is what gets embedded with pybind11... so the _version_ of Python for that build is fixed from there forward--running that executable with an activated venv made with a different version of Python would probably cause Bad Things™️.
2. The numpy headers that are found at that time are compiled in. So in theory running that executable with a different venv containing a different version of NumPy _might_ blow up... there are probably a lot of unpredictable caveats there.

Notably what is _not_ included above is that the path to the venv that was present at build time is nowhere compiled in... generally this is probably a good thing (debate?) and not something we'd want to change. Even if we did, besides the scenario of moving that venv aside and putting a different one in it's place--which we can't control--it's possible to upgrade a venv in-place to a newer version of Python.... so we're not really accomplishing much meaningful.

> do we need to...more robustly ensure the right environment gets loaded at runtime?

The "right" environment--if this means the venv present at compilation time--is arguably indeterminable _ex post facto_, for reasons including those described above. We _could_ (and maybe should) inspect the present venv at startup and try to use some heuristics (exchanging one set of heuristics for another now) to determine whether the presently loaded Python environment (which might _not_ be a venv) is valid for the ngen build... e.g. Python micro version different = WARN, Python minor version different = FAIL? Similar checks on NumPy?

mattw-nws: > or worse, a mix of both...

Also, the "mix of both" is probably the biggest problem with the _current_ code... it is at least possible (if not likely) to run with one version of Python active (e.g. the system Python) and for ngen to pick up package paths from another version (e.g. the version used to create whatever is at the path `.venv`).

Lastly, attempting to "fix" (somehow) the venv that was run at build time as the one that must be used at runtime would interfere with what was trying to be done in #441 --essentially, what the submitter wanted there was the ability to switch venvs from run to run trying different versions of a module (and combinations of dependencies of that module, I think)...which seems like a useful use case and something we should try not to prevent... though it's impossible to support without caveats (all must use same Python version, for starters).

robertbartel: Yes, I am fairly confident that Python itself and numpy (that ngen uses) are compiled into ngen.  You touch on my concern, though.  If the same environment isn't used at runtime, that opens the door to discrepancies with either the required version of Python or numpy for a BMI module imported from the runtime Python environment.

robertbartel (approved):

PhilMiller:
For posterity, @mattw-nws @robertbartel and I discussed this, and decided to go ahead with the build system change here, and to separately consider the runtime behaviors that this brings into consideration.

mattw-nws:
Belated note, yes I did have to do this in Visual Studio Code (VSCode) via `.vscode/settings.json` after this PR, to get the IDE's cmake integration to find a virtual environment it was finding automatically before:

```json5
{
    "cmake.environment": {
        "VIRTUAL_ENV":"/path/to/.venv",
        // ...
    }
}

PhilMiller:
I think the 'right' thing to do would rather be to define `Python_ROOT_DIR` in a CMake user presets file. So, don't pretend that all the other stuff the venv drops in the environment is also set, just point CMake at the installation and let it do its thing.

PhilMiller:
https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html

PR #542, opened 2023-06-23
Pin pyarrow to avoid crash in 12.x

mattw-nws:
See PR #528 ... it was expected that the next release of pyarrow would make this problem go away. However, looking closer at https://github.com/apache/arrow/issues/15054 and https://github.com/apache/arrow/pull/33858 , it appears that the underlying issue supposedly _was fixed_ in the 12.0.0 release... so the fact that we're still seeing it is confusing. Either way, this problem should probably now be referred to t-route, created issue https://github.com/NOAA-OWP/t-route/issues/614 .

## Additions

- 

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

1.

- [x] Linux

### DIFF
--- a/.github/workflows/module_integration.yml
+++ b/.github/workflows/module_integration.yml
@@ -145,7 +145,7 @@ jobs:
           pip install 'pip>=23.0,<23.1'
           pip install -U setuptools
           pip install -r requirements.txt
-          pip install deprecated 'pyarrow!=12.*' tables geopandas numpy
+          pip install deprecated 'pyarrow==11.0.0' tables geopandas numpy
           if [ ${{ runner.os }} == 'macOS' ] 
           then
             export LIBRARY_PATH=/usr/local/lib/gcc/11/
--- a/.github/workflows/module_integration.yml
+++ b/.github/workflows/module_integration.yml
@@ -145,7 +145,7 @@ jobs:
           pip install 'pip>=23.0,<23.1'
           pip install -U setuptools
           pip install -r requirements.txt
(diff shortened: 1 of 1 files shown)


.github/workflows/module_integration.yml
```diff
@@ -145,7 +145,7 @@ jobs:
           pip install 'pip>=23.0,<23.1'
           pip install -U setuptools
           pip install -r requirements.txt
-          pip install deprecated 'pyarrow!=12.0.0' tables geopandas numpy
+          pip install deprecated 'pyarrow!=12.*' tables geopandas numpy
```
PhilMiller, line 148:
```suggestion
          pip install deprecated 'pyarrow==11.0.0' tables geopandas numpy
```
I suggest pinning a known-working version, rather than trying to chase exclusions of presently non-working versions. Then we can stick an issue in the backlog to revisit the selected version sometime later.

mattw-nws: Yeah especially now that I think it _should_ be working in 12.x, I was going to come back and do this instead.

mattw-nws: As a side note I'm never again going to put `deprecated` as the first package after `pip install` because if you don't know that's a package it really looks like part of the pip command.

PhilMiller: I'll admit that threw me off a bit when I saw it.

PhilMiller (approved):

PR #564, opened 2023-07-10
#362: Remove repeated searches for Boost that were picking up mixed versions

### LINKED ISSUE #362: Setting ${BOOST_ROOT} doesn't seem to help -- Boost has to be system searchable
Using Ubuntu 20.04, the default Boost library (`sudo apt install libboost-dev-all`) is v1.71. NGEN requires 1.72 (or 1.72+ presumably), so a custom Boost installation is needed. So I pulled the latest library from https://www.boost.org/users/download/

These cmake lines seems to indicate that the script will respect a custom BOOST library installed somewhere: 
https://github.com/NOAA-OWP/ngen/blob/895708ae3052ce35687a16cf8c199873f46af262/CMakeLists.txt#L86-L97

But just below that line, this line performs an explicit search for the specified Boost library version, ignoring the $BOOST_ROOT environment variable: 
https://github.com/NOAA-OWP/ngen/blob/895708ae3052ce35687a16cf8c199873f46af262/CMakeLists.txt#L109

Commenting the explicit test exposes embedded tests for system searchability of the Boost libraries, not (apparently) respecting the $BOOST_ROOT (e.g., there are errors about a $Boost_DIR that come from more deeply embedded calls to `find_package`.)

I tried more explicitly specifying the path, as shown here, but that did not solve the later errors. 
https://stackoverflow.com/questions/49816206/cmake-find-package-specify-path

Ultimately, I just built and installed the custom library at the system level -- it seems like BOOST_ROOT wasn't helping too much.

PhilMiller:
Having multiple calls to `find_package(Boost)`, some specifying a required version and others not, led to CMake searching for the package multiple times, and flipping imports between found instances depending on which one was found last (or maybe within the calling scope). This could result in a quietly and inscrutably broken build, with ABI mismatches and ODR violations.

Address these concerns by removing all of the `find_package` calls other than the one with the specified version at the root of the package.

Fixes #362

## Additions

- N/A

## Removals

- N/A

## Changes

- Drop `find_package(Boost)` calls from subsidiary packages' `CMakeLists.txt`

## Testing

1.

## Screenshots

See #362 

## Checklist

### DIFF
--- a/src/core/catchment/CMakeLists.txt
+++ b/src/core/catchment/CMakeLists.txt
@@ -13,9 +13,6 @@ target_include_directories(core_catchment PUBLIC
         ${PROJECT_SOURCE_DIR}/include/core/catchment/giuh
         )
 
-# TODO: consider setting a minimum or required version
-find_package(Boost)
-
 add_subdirectory("giuh")
 
 target_link_libraries(core_catchment PUBLIC
--- a/src/core/catchment/giuh/CMakeLists.txt
+++ b/src/core/catchment/giuh/CMakeLists.txt
@@ -13,7 +13,6 @@ target_include_directories(core_catchment_giuh PUBLIC
         ${PROJECT_SOURCE_DIR}/include/core/catchment/giuh
         )
 
-find_package(Boost)
 target_link_libraries(core_catchment_giuh PUBLIC
         Boost::boost                # Headers-only Boost
         )
--- a/src/core/hydrolocation/CMakeLists.txt
+++ b/src/core/hydrolocation/CMakeLists.txt
@@ -13,9 +13,6 @@ target_include_directories(core_hydrolocation PUBLIC
         ${PROJECT_SOURCE_DIR}/include/utilities
         )
 
-# TODO: consider setting a minimum or required version
-find_package(Boost)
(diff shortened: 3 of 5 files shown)


mattw-nws:
Ugh! While, technically, I don't think we could get into an ABI mismatch situation if we were only using header-resident components (right?) yeah I can see how this could lead to major headaches while doing iterative builds during development if there were mismatches in API versions picked up over time. @hellkite500, is there any reason to keep the `find_boost` calls in the subdirectory CMakeLists.txt files? I don't think there was really ever an intent for those "libraries" to be able to be built independently...?

PhilMiller:
If any types in those headers differed between versions and were passed across function calls, then you'd have an ABI mismatch.

However, even if there were just distinct callers to function templates that came from different versions, you'd still have violations of the [One Definition Rule](https://en.cppreference.com/w/cpp/language/definition). Having actually debugged a failure resulting from an ODR violation, it's **not fun**

PhilMiller:
Also, if you look at my notes on #362, I was getting version mismatches *in a single, clean build*

hellkite500 (approved):

PR #576, opened 2023-07-17
#559: Import definitions of data() and span() from Boost Core

### LINKED ISSUE #559: Get some implementation of `std::span`
`std::span` describes array data that the recipient can access and modify, but without any allocation or transfer of ownership. It was standardized in C++20, while we're using C++14 currently. For at least the BMI model interaction and `mdarray`/`mdframe` IO stuff, it would be highly desirable to have an implementation available.

PhilMiller:
Fixes #559 

## Additions

- span.hpp: copy of Boost 1.82 Core span.hpp
- boostcore_data.hpp: copy of Boost 1.82 Core data.hpp (named like this to avoid `.gitignore`)

## Testing

1. CI covers the case of referencing a copy of Boost 1.72, prior to its implementation of `span`

## Notes

-

## Todos

- [x] Test against Boost 1.78
- [x] Test against Boost 1.82

## Checklist


include/utilities/span.hpp
```diff
@@ -0,0 +1,412 @@
...
+#include <boost/core/span.hpp>
+#else
+
+#ifndef BOOST_CORE_SPAN_HPP
+#define BOOST_CORE_SPAN_HPP
+
+#if __has_include(<boost/core/data.hpp>)
+#include <boost/core/data.hpp>
+#else
+#include "boostcore_data.hpp"
```
PhilMiller, line 19: Named like this to avoid tripping over the `.gitignore` line reading `boost_*`

PhilMiller:
CI testing is apparently covering the case where we pull in our own copy of the minimum required Boost 1.72, which has implementations of neither `data()` nor `span`. We should also test against Boost 1.78, which first adds a `span` implementation, and at least `1.82`, where it also provides `data`

include/utilities/span.hpp
mattw-nws: Hmm.... that should probably be `/boost_*` instead? Or `/boost_*/` if that works to match only directories? Purpose was to exclude boost distributions which may be untarred at the root of the repo. Probably should match only directories and probably only match at the root of the working copy. I'd consider this in-scope to fix if we think it's a better solution, which I'm thinking it might be.

src/NGen.cpp
```diff
@@ -3,6 +3,8 @@
 #include <string>
 #include <unordered_map>
 
+#include <utilities/span.hpp>
```
mattw-nws, line 6: Noted 🤨 but sure if we want to make sure it compiles for now.

mattw-nws (approved):

src/NGen.cpp
PhilMiller: To memorialize our Slack conversation - that was the intention.

include/utilities/span.hpp
PhilMiller: I don't expect the `data.hpp` header to ever be of interest to include on its own, so I don't think it's a big deal to just leave it. It's not going to make anything else worse, or otherwise cost us anything, I don't think.

PhilMiller:
I would consider this ready for review and potential merge if it gets tested against newer Boost versions. As noted, I'm away the rest of this week, so if we want to keep this moving, having someone else test it would be good. Someone else testing it would probably also help guard against me inadvertently testing in a way that's not representative of others' needs.

program-- (approved):
This PR was tested by modifying `geojson::read(2)` to accept a `boost::span<std::string>` as second argument, then compiling the `ngen` target using the below matrix, and running a simple execution to verify output is the same as the pre-`span` code.

The macros for included headers were manually checked with each boost version to ensure they pulled the correct header (i.e. either the vendored version, or directly from boost)

**All matrix entries compiled, ran, and outputted successfully**

Changing the parameter tests a few things from the boost implementation:
1. `std::vector` conversion to `boost::span`
2. `boost::span` methods (specifically `.empty()`, `.begin()`, and `.end()`)
3. `std` algorithms on `boost::span` (specifically `std::find`)

Compiler | Version| Boost Version
--- | --- | ---
clang | 14.0.6 | 1.75.0
clang | 14.0.6 | 1.78.0
clang | 14.0.6 | 1.82.0
clang | 15.0.7 | 1.75.0
clang | 15.0.7 | 1.78.0
clang | 15.0.7 | 1.82.0
gcc | 13.1.1 | 1.75.0
gcc | 13.1.1 | 1.78.0
gcc | 13.1.1 | 1.82.0
IntelLLVM | 2023.1.0 | 1.75.0
IntelLLVM | 2023.1.0 | 1.78.0
IntelLLVM | 2023.1.0 | 1.82.0

Checking with the above matrix was automated using this script:
```bash
#!/usr/bin/env bash
set -e

# Change as needed -------------------------
C_COMPILERS=(
    "/usr/lib/llvm14/bin/clang-14"
    "/usr/bin/clang"
    "/usr/bin/gcc"
    "/opt/intel/oneapi/compiler/2023.1.0/linux/bin/icx"
)

CXX_COMPILERS=(
    "/usr/lib/llvm14/bin/clang++-14"
    "/usr/bin/clang++"
    "/usr/bin/g++"
    "/opt/intel/oneapi/compiler/2023.1.0/linux/bin/icpx"
)

# Folders containing boost headers. In this case,
# they are located in the root of the `ngen` folder.
BOOST_VERSIONS=(
    "boost_1_75_0"
    "boost_1_78_0"
    "boost_1_82_0"
)
# ------------------------------------------

for i in "${!C_COMPILERS[@]}"; do
    c_path="${C_COMPILERS[i]}"
    cxx_path="${CXX_COMPILERS[i]}"
    c_name=$(basename ${c_path})
    cxx_name=$(basename ${cxx_path})
    echo "Using C Compiler: ${c_path}"
    echo "Using C++ Compiler: ${cxx_path}"

    for boostver in "${BOOST_VERSIONS[@]}"; do
        echo "Cleaning build directory"
        rm -r build
    
        echo "Using Boost Version: ${boostver}"
    
        checksum_name="${c_name}_${boostver}.checksum"

        echo -e "Configuring CMake build:\n\tC     : ${c_path}\n\tCXX   : ${cxx_path}\n\tBOOST : ${boostver}"

        cmake \
            --no-warn-unused-cli \
            -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE \
            -DCMAKE_BUILD_TYPE:STRING=Release \
            -DCMAKE_C_COMPILER:FILEPATH=${c_path} \
            -DCMAKE_CXX_COMPILER:FILEPATH=${cxx_path} \
            -DBOOST_ROOT=${boostver} \ # CMakeLists.txt was modified to force this (i.e. prevent finding in system paths)
            -S . \
            -B build \
            -G Ninja 

        echo "Building CMake project"
        cmake --build build --config Release --target ngen --

        echo "Running example NGen run"
        ./build/ngen \
            data/catchment_data.geojson "cat-67,cat-52" \
            data/nexus_data.geojson "nex-68,nex-34" \
            data/example_realization_config.json

        echo "Outputting hashes"
        sha256sum cat-52.csv > ${checksum_name}
        sha256sum cat-67.csv >> ${checksum_name}
        sha256sum nex-34_output.csv >> ${checksum_name}
        sha256sum nex-68_output.csv >> ${checksum_name}
        rm cat-*.csv
        rm nex-*.csv
    done
done
```

Seems good to me, assuming this testing is sensible!

PhilMiller:
@program-- Thank you for the **very** thorough (and thoroughly documented) testing.

Added more of the team for awareness that this is becoming available in the codebase.

Anywhere that something is passed as `std::vector` or `std::vector const&`, consider using `span` instead. Keep in mind that `span` does not own the memory it references. The allocation must outlive use of the `span`. So, rather than returning a `span`, consider passing one as an output argument from the caller, designating where output should be written.

Issue #535, opened 2023-06-20
Fail execution if `output_variables` has invalid entries


ajkhattak:
## Current behavior
 - In the `output_variables` list at the multi_bmi level, if the user provides a variable that does not exist in the BMIs, then the framework ignores that high-level `output_variables` list and outputs the variables of the vary last BMI instead of `throwing an error `

## Expected behavior
 - If a variable listed in the `output_variables` at the multi_bmi level does not exist, the framework should stop the execution and let the user know that the variable(s) provided does not exist in any of the BMI output variables

## Screenshots
If all variables in the `output_variables` exist in the BMIs, the output looks like this
<img width="347" alt="Screen Shot 2023-06-20 at 7 02 09 PM" src="https://github.com/NOAA-OWP/ngen/assets/15165757/d31c9835-0936-4c00-a859-af48382ec0aa">

and when a variable `var_DNE` does not exist, the output looks like this -- a mismatch between the header size and the data points.
<img width="1272" alt="Screen Shot 2023-06-20 at 7 02 59 PM" src="https://github.com/NOAA-OWP/ngen/assets/15165757/b92ee7a2-7c1d-48eb-92df-e94bc4daee2b">

The realization file used in the above examples is attached here (CFE+PET+SLoTH example).
[realization_cfe_pet.json.zip](https://github.com/NOAA-OWP/ngen/files/11808608/realization_cfe_pet.json.zip)

mattw-nws:
@stcui007 ... Per discussion we had earlier, I double-checked and in this example file the `output_variables` and `output_header_fields` are in the _right_ place--that is, the top-level of the multi-BMI formulation, _not_ on the last BMI module inside the multi-BMI. So this is a valid issue and the example is good.

stcui007:
@Matthew Williamson - NOAA Federal ***@***.***>   I did
run some tests with the provided config file last week. So we want to make
the change to the ngen code so that it complies with the Expected behavior?

PR #577, opened 2023-07-19
Bug fix: invalid ouput_variable entry in bmi_multi module

stcui007:
Fix invalid entry error that causes incorrect output header

## Additions
 -

## Removals

-

## Changes

src/realizations/catchment/Bmi_Multi_Formulation.cpp
test/realizations/catchments/Bmi_Multi_Formulation_Test.cpp

## Testing

1.

## Notes

-

## Todos

-

## Checklist

- [ x] PR has an informative and human-readable title
- [ x] Any _change_ in functionality is tested

1.

- [x ] Linux

### DIFF
--- a/src/realizations/catchment/Bmi_Multi_Formulation.cpp
+++ b/src/realizations/catchment/Bmi_Multi_Formulation.cpp
@@ -100,11 +100,15 @@ void Bmi_Multi_Formulation::create_multi_formulation(geojson::PropertyMap proper
             out_vars[i] = out_vars_json_list[i].as_string();
         }
         set_output_variable_names(out_vars);
+        set_glob_bmi_model_type(get_output_variable_names());
+        //set_sub_bmi_model_type(get_output_variable_names());
     }
     // Otherwise, for multi BMI, the BMI output variables of the last nested module should be used.
     else {
         is_out_vars_from_last_mod = true;
         set_output_variable_names(modules.back()->get_output_variable_names());
+        set_glob_bmi_model_type(get_output_variable_names());
+        //set_sub_bmi_model_type(get_output_variable_names());
     }
     // TODO: consider warning if nested module formulations have formulation output variables, as that level of the
     //  config is (at present) going to be ignored (though strictly speaking, this doesn't apply to the last module in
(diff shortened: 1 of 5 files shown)


src/realizations/catchment/Bmi_Multi_Formulation.cpp
```diff
@@ -303,6 +303,23 @@ string Bmi_Multi_Formulation::get_output_line_for_timestep(int timestep, std::st
...
 
+        std::cout << "Multi: timestep = " << timestep << std::endl;
+        for (int i = 0; i < output_var_names.size(); ++i) {
+            std::cout << "Multi: output_var_names," << output_var_names[i] << std::endl;
+        }
+
+        // Remove invalid output variable name entry in the globle bmi-multi module
+        std::vector<std::string>::iterator it;
+        int last_index = modules.size() - 1;
+        auto module = modules[last_index];
```
mattw-nws, line 314: Actually this is wrong. The list of valid variable names for a multi_bmi formulation is _not_ found just in the last BMI module in the list; rather it is the union of all names of all BMI modules, including the forcing DataProvider. (Using the outputs from the last module is only relevant as the _default_ set of names used when nothing is explicitly specified, as done further down in this function.)

Perhaps this is why the test below was disabled? Does it attempt to retrieve a variable that is not from the last BMI module in the multi_bmi?

mattw-nws (changes requested):
Beyond what's discussed in the inline review comments, the approach here is actually done in the wrong place, for several reasons.

First, while #535 specifically addresses a mult_bmi formulation, the `output_variables` configuration can be applied to any kind of formulation and the problematic result behavior (headers mismatching the data columns) will occur for other formulation types also--so this needs to be tackled in Bmi_Module_Formulation, somehow, even if it requires common code/a method in that class that is then called somewhere in BMI_[language]_Formulation subclasses. The bmi_multi case might require extra consideration/code because of the unique way that a list of valid variable names is derived (not from a single BMI module), but still the issue affects non-multi formulations.

Also, as discussed in the code comments, this should not be checked every time an output line is processed. First because that will run far more times than necessary, and second because it will only get checked after the simulation has started running at least one timestep. I believe all of the formulation types pre-compute the list of available variables when the formulation is created--this would be the much better place to check if the custom `output_variables` list is valid.

stcui007:
First, I agree that the code as it is currently written runs every time
step. That can be easily fixed.

Also, in its current form in the BMI_Multi_formulation.cpp, it catches the
last module's output header in config jason file for output. I think that's
the module the get_output_line_timestep() uses for writing output. I have
tested and it seems to be the case.

The reason that test 3_a (which is for python) is currently disabled is
because the Bmi_Multi_Formulation_Test.cpp has a different output variable
list than that in the extern/test_bmi_py. That triggers an exception to be
thrown by design in Bmi_Multi_Formulation.cpp.

mattw-nws:
> Also, in its current form in the BMI_Multi_formulation.cpp, it catches the last module's output header in config jason file for output. I think that's the module the get_output_line_timestep() uses for writing output.

I mentioned this in the last review comment, but it captures the last module's output fields as the _default_ set of variables if nothing else is explicitly specified (via `output_variables` on the multi_bmi itself). But this needs to validate if all the variables specified in `output_variables` are in the set of _valid_ variables for the formulation, not if they are all in the _default_ set of variables.

As a concrete example, `output_variables` could contain `water_potential_evaporation_flux`, which comes from the PET module, which is not the _last_ BMI module in the example realization config in #535, but would be a valid option to have in `output_variables`--the code in this PR would throw an exception in this case. Also, it would be valid to include `atmosphere_water__liquid_equivalent_precipitation_rate` in `output_variables`, which typically comes from the forcing data, not from the last BMI module--this code would throw an exception in that case also.

src/realizations/catchment/Bmi_Multi_Formulation.cpp
stcui007: It is possible we may get an incorrect header in the output if we check all the modules.

mattw-nws: How? _Checking_ all of the modules to see if a given variable is valid does not cause all of the modules' outputs to appear in the output--what causes them to appear in the output is if the user has explicitly called for them using `output_variables`.

PR #580, opened 2023-07-24
#536: Add AddressSanitizer flags to CI tests

PhilMiller:
Additions and changes to the codebase should not be allowed to introduce new errors detected by runtime dynamic analysis by AddressSanitizer.

## Additions

- Address Sanitizer is now activated on all tests run in CI
- Geopackage units tests are run in CI (they were previously missing)

## Changes

- Fixed several errors detected by ASan in GeoPackage code and its tests
- Suppress ASan report from zombie GIUH code

## Testing

This expands what CI covers

## Notes

- Leak detection in ASan is disabled across the board, because there are a number of small leaks that bite every test, which then cause it to abort rather than reporting the test results

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows project standards (link if applicable)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

### Target Environment support

- [x] Linux
- [x] macOS

PR #583, opened 2023-07-25
Use `span` to avoid copying WKB geometry data

### LINKED ISSUE #582: Demo usage of `span`

PhilMiller:
Fixes #582 

## Changes

- Convert internal GeoPackage interfaces from passing around `const std::vector<uint8_t>&` to passing `const boost::span<const uint8_t>`, so that callers won't need to have a `vector` in hand to pass as an argument
- Avoid constructing and copying into a temporary `vector` with almost the entire geometry blob, since we can now pass a `span` referring to the source of the copy

## Testing

1. `test_geopackage` runs equally successfully before and after (with fixes from #580)

## Checklist

### DIFF
--- a/src/geopackage/geometry.cpp
+++ b/src/geopackage/geometry.cpp
@@ -75,7 +75,8 @@ geojson::geometry geopackage::build_geometry(
     }
 
     if (!is_empty) {
-        const std::vector<uint8_t> geometry_data(geometry_blob.begin() + index, geometry_blob.end());
+        const boost::span<const uint8_t> geometry_data(geometry_blob.data() + index,
+                                                       geometry_blob.data() + geometry_blob.size());
         auto wkb_geometry = wkb::read(geometry_data);
         geojson::geometry geometry = boost::apply_visitor(pvisitor, wkb_geometry);
         return geometry;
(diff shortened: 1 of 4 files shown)


src/geopackage/geometry.cpp
```diff
@@ -75,7 +75,8 @@ geojson::geometry geopackage::build_geometry(
     }
 
     if (!is_empty) {
-        const std::vector<uint8_t> geometry_data(geometry_blob.begin() + index, geometry_blob.end());
+        const boost::span<const uint8_t> geometry_data(geometry_blob.data() + index,
+                                                       geometry_blob.data() + geometry_blob.size());
```
PhilMiller, line 79: Note to readers: `span` cannot be constructed from opaque iterators, only (begin pointer, end pointer) or (begin pointer, size)

PhilMiller: (or from other spans of convertible types, or from whole containers or ranges or types that it knows how to convert)

program-- (approved):
Looks good to me, and I think it's a fairly easy-to-understand set of changes for everyone else! 😄

src/geopackage/geometry.cpp
PhilMiller: The latter argument could also have been written as 
```suggestion
                                                       geometry_blob.size() - index);
```
But that requires more thought to understand, than just specifying the end address to mean 'everything from index onward'

PR #588, opened 2023-07-26
Allow passing feature properties to BMI models through `model_params`

### LINKED ISSUE #554: Pass catchment hydrofabric attributes to BMI models as `model_params`
Especially wIth GeoPKG hydrofabric support in #522, add the ability to pass attributes from the features to the BMI model stack. The foremost use-case is to pass `area_sqkm` to a lumped module that needs it to convert rainfall depth to volume, but also we may store other model-specific or general attributes/parameters in the hydrofabric where appropriate.

These parameters/values should be passed only once, before the input variables (gotten from `get_input_var_names`) are passed--at about the same time as any `model_params` from the realization config are passed, and using a similar mechanism. A model accepting such a parameter should respond properly to metadata functions for that parameter such as `get_var_type` and this information should determine how the value is passed to the module.

Preferably, some mechanism should be designed such that the value is only attempted to be set on a BMI model that needs it--possibly using the existing `model_params` feature with an object value (instead of a numeric value) that indicates that the source of the parameter is the hydrofabric. Some way of mapping the hydrofabric attribute name to a model parameter name--which may be different--is also desirable, e.g. leverage `variables_names_map` or allow aliasing information in a JSON key within `model_params`.

This feature is focused on lumped catchment models defined in the hydrofabric. Future similar use-cases to keep in mind:
 * Nexus attributes may be passed to the BMI version of a routing module (e.g. roughness coefficient)
 * We may in the future, for gridded models, have a (non-hydrofabric) grid parameter file (e.g. NetCDF file) with one or more parameter variables which would need to be made available to the model using a similar idiom.

## Current behavior

No information from the hydrofabric is made available to BMI models

## Expected behavior

A BMI model (including nested models in a multi-BMI stack) should be able to have attributes from the hydrofabric passed to it.

1.

program--:
This PR resolves #554 by providing a new value specification for formulation model parameters (`model_params`).

An *external model parameter* is described as:
```
<parameter name>: { "source": <source> [, "from": <mapping>] }
```
Where the following represent:
- `<parameter name>`: the name the BMI model will use
- `<source>`: This PR only supports a source of `"hydrofabric"`, but this scaffolds out the possibly of supplying a path to a separate file that can be joined on demand.
- `<mapping>`: A mapping name that signifies what the property is named as within `<source>`

An example of what this could look like is as follows:
```json
{
  "model_params": {
    "area_sqkm": {
      "source": "hydrofabric",
      "from": "area"
    },
    "some_static_param": 3.0,
    "length_km": {
      "source": "hydrofabric"
    }
  }
}
```

In this example, `area` and `length_km` are defined within the hydrofabric feature, and are passed to the BMI model as `area_sqkm` and `length_km`, respectively.

> **Note**: notice that `length_km` does not specify a `"from"` key, so we assume it is named the same within the hydrofabric feature's properties. If it **does not** exist within the feature properties, it is ignored, and removed from `model_params`.

## Additions

- Adds a member function `parse_external_model_params` to `realization::Formulation_Manager` that performs the parsing/modification of `model_params`.
- Adds a small section of logic to `Formulation_Manager::read` that checks for a `model_params` key when iterating the catchment-specific formulations.
- Adds an additional set of unit tests to the formulation manager tests that verifies this PR's functionality.

## Testing

All unit tests pass locally in `Release` with GCC 13.1.1, Clang 15.0.7, and IntelLLVM 2023.

## Todos

- [X] Parse and modify `model_params` for global formulations.
- [X] Add tests for global formulation `model_params`.
- [X] Parse and modify `model_params` for catchment-specific multi-BMI formulations.
- [X] Add tests for catchment-specific multi-BMI `model_params`.
- [x] Parse and modify `model_params` for global multi-BMI formulations.
- [x] Add tests for global multi-BMI `model_params`.

## Checklist

- [X] Linux

### DIFF
--- a/extern/test_bmi_cpp/include/test_bmi_cpp.hpp
+++ b/extern/test_bmi_cpp/include/test_bmi_cpp.hpp
@@ -139,18 +139,24 @@ class TestBmiCpp : public bmi::Bmi {
         bool use_input_array, use_output_array;
    
         std::vector<std::string> input_var_names = { "INPUT_VAR_1", "INPUT_VAR_2" };
-        std::vector<std::string> output_var_names = { "OUTPUT_VAR_1", "OUTPUT_VAR_2" };
+        std::vector<std::string> output_var_names = { "OUTPUT_VAR_1", "OUTPUT_VAR_2", "OUTPUT_VAR_4", "OUTPUT_VAR_5" };
+        std::vector<std::string> model_var_names = { "MODEL_VAR_1", "MODEL_VAR_2" };
         std::vector<std::string> input_var_types = { "double", "double" };
-        std::vector<std::string> output_var_types = { "double", "double" };
+        std::vector<std::string> output_var_types = { "double", "double", "double", "double" };
+        std::vector<std::string> model_var_types = { "double", "double" };
         std::vector<std::string> input_var_units = { "m", "m" };
-        std::vector<std::string> output_var_units = { "m", "m/s" };
+        std::vector<std::string> output_var_units = { "m", "m/s", "m", "m" };
+        std::vector<std::string> model_var_units = { "m", "m" };
         std::vector<std::string> input_var_locations = { "node", "node" };
-        std::vector<std::string> output_var_locations = { "node", "node" };
+        std::vector<std::string> output_var_locations = { "node", "node", "node", "node" };
+        std::vector<std::string> model_var_locations = { "node", "node" };
 
         std::vector<int> input_var_item_count = { 1, 1 };
-        std::vector<int> output_var_item_count = { 1, 1 };
+        std::vector<int> output_var_item_count = { 1, 1, 1, 1 };
+        std::vector<int> model_var_item_count = { 1, 1 };
         std::vector<int> input_var_grids = { 1, 1 };
-        std::vector<int> output_var_grids = { 1, 1 };
+        std::vector<int> output_var_grids = { 1, 1, 1, 1 };
+        std::vector<int> model_var_grids = { 1, 1 };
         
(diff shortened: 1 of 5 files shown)


program--:
`test_unit` is currently failing since it doesn't look like `test_bmi_cpp` gets built in that workflow. I think the unit tests for this PR probably need to move into `Bmi_Cpp_Formulation_Test.cpp`, but I need to make sure that's appropriate first.

mattw-nws:
> `test_unit` is currently failing since it doesn't look like `test_bmi_cpp` gets built in that workflow. I think the unit tests for this PR probably need to move into `Bmi_Cpp_Formulation_Test.cpp`, but I need to make sure that's appropriate first.

I have this fixed in a parallel PR... looks like I need to make a subset of that PR into its own PR anyway. I'll try to get this out today.

mattw-nws:
#594 should fix the unit test issue.

program--:
Moving back to draft -- accidentally committed upstream changes from SoilMoistureProfiles (which I'm not even sure how those got pulled on my local side).

mattw-nws:
Hmm. Fun fact... this worked:

```
                                "model_params": {
                                    "sloth_2_var_C(1,double,m,node,sloth_1_var_A)": 0.0,
                                    "sloth_2_var_D(1,double,kg)": 4.0,
                                    "sloth_2_var_E(1,double,L)": 5.0,
                                    "area_sqkm(1,double,m^2)": {"source": "hydrofabric", "from": "areasqkm" }
                                },
```

mattw-nws (approved):

PR #593, opened 2023-07-31
Basic abstractions for core framework entities

PhilMiller:
## Additions

-

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

1.

- [ ] Linux


include/core/Coupling.hpp
```diff
@@ -0,0 +1,89 @@
...
+    { }
+
+    void transfer() override
+    {
+        SpMV(transfer_weights_(), from_.storage(), to.storage());
+    }
+
+    shared_ptr<SourceVariable> from_;
+    shared_ptr<SinkVariable> to_;
+    SparseMatrix transfer_weights_;
```
PhilMiller, line 88: Given what's available on WCOSS, it may actually be easiest to define this from a NumPy type, since that will have a better-than-naive implementation

include/core/ModelVariable.hpp
```diff
@@ -0,0 +1,47 @@
...
+struct SourceVariable : public virtual ModelVariable
+{
+    shared_ptr<DataProvider> provider;
+
+};
+
+
+struct BmiSourceVariable : public virtual SourceVariable
+{
+    // constraint: dynamic_cast<Bmi_Formulation*>(provider) != nullptr
```
PhilMiller, line 32: This should try to use BMI's `get_value_ptr` to arrange `storage()`, and fall back to allocating a buffer that it will load into

PhilMiller:
My high-level idea of how this would play out in the rest of the framework is that
* initialization would construct a set of `Discretization`s, then a set of `Variable`s coming from the models it instantiates on those `Discretization`, then a set of `Coupling` between variables
* Timestepping would `update()` all the models in a layer, then call `transfer()` on all of the `Coupling`s from the `Variable`s produced by those models

include/core/Discretization.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+struct Discretization
+{
+    virtual ~Discretization() = default;
+
+    enum GridType { hydrofabric_locations, spatial_structured, spatial_unstructured };
+    virtual GridType grid_type() = 0;
+
+    // Must cover an identified geographic domain
+    virtual GeoDivide divide() = 0;
```
PhilMiller, line 11: Or maybe omit this, if we'd have to get into weird constructive geometry expressions, like "hydro-locations across the CONUS domain, minus those within a region of interest modeled by a different formulation"

include/core/Discretization.hpp
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+struct Discretization
+{
+    virtual ~Discretization() = default;
+
+    enum GridType { hydrofabric_locations, spatial_structured, spatial_unstructured };
```
PhilMiller, line 7: Do we expect these to end up one-to-one with concrete implementations of `Discretization`?

include/core/Discretization.hpp
```diff
@@ -0,0 +1,34 @@
...
+    virtual GridType grid_type() = 0;
+
+    // Must cover an identified geographic domain
+    virtual GeoDivide divide() = 0;
+
+    virtual size_t size() = 0;
+    virtual int rank() = 0;
+    virtual std::span<Foo> get_points() = 0;
```
PhilMiller, line 15: I think technically, we may want to distinguish between a discretization over `Catchment`, one over a designated `FlowPath` within said `Catchment`, and the outflow `Nexus` that those feed. Maybe even template `Discretization` over those possibilities as well as `GeometricPoint`, so that `get_points` can return the right kind of collection, rather than numbers with arbitrarily different meanings.

If we distinguished a `Discretization<Catchment>` from a matching `Discretization<Nexus>`, we'd need to be a bit more sophisticated in expressing the 'simple' one-to-one coupling cases, but I think that may be worthwhile. Additionally, if the number or ordering of those actually differs in the input data, then we can't just map those one-to-one anyway.

include/core/Coupling.hpp
```diff
@@ -0,0 +1,97 @@
...
+        // scaling should be set to address that
+    }
+
+    void transfer() override
+    {
+        // Short-circuit the case of shared storage
+        if (dynamic_pointer_cast<ModelVariable>(from_) == dynamic_pointer_cast<ModelVariable>(to_))
+            return;
+
+        boost::span<double> src = from_->storage(), dst = to_->storage();
```
hellkite500, line 47: If these data are coming from BMI models, then assuming `double` types isn't sufficient.

include/core/Discretization.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+template <typename T>
+struct ConcreteDiscretization : public Discretization
+{
+    using PointType = T;
+    virtual boost::span<PointType> get_points() = 0;
+};
+
+// One point per catchment/nexus/whatever set elements
+struct HydroLocationDiscretization : public ConcreteDiscretization<int>
```
hellkite500, line 31: Might need to disscuss this in more detail.  I'm tyring to understand the relationship of this struct to a `HydroLocation` as it relates to HY_Features and the conceptual model as well as the hydrofabric data model we have.

include/core/Discretization.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+#include "span.hpp"
+
+using GeoDivide = int;
+
+struct Discretization
+{
+    virtual ~Discretization() = default;
+
+    enum GridType { hydrofabric_locations, spatial_structured, spatial_unstructured };
```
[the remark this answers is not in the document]
PhilMiller: From our discussion - yes, I should/will align this with BMI grid types. I will want to 'subtype' the `points` grid type to specify what each point is actually associated with - e.g. a nexus, a catchment, an arbitrary coordinate, etc.

include/core/Discretization.hpp
PhilMiller: As discussed, this is somewhat mis-named. I'll correct to follow the bit about 'points' grids above

include/core/Coupling.hpp
```diff
@@ -0,0 +1,97 @@
...
+                              double area_sqkm)
+        : DirectCoupling(from, to, timestep.count()*area_sqkm*unit_correction)
+    {
+        //assert(from->element_dimensionality() == length/time);
+        //assert(to->element_dimensionality() == length*length*length);
+    }
+};
+
+template <typename T>
+using SparseMatrix = int;
```
PhilMiller, line 74: Maybe Scipy.sparse, or maybe something from Boost Numeric/uBLAS

include/core/Discretization.hpp
```diff
@@ -0,0 +1,134 @@
...
+        vector,
+        unstructured,
+        structured_quadrilateral,
+        rectilinear,
+        uniform_rectilinear
+    };
+    virtual GridType grid_type() const = 0;
+
+    // Must cover an identified geographic domain
+    virtual GeoDivide divide() const = 0;
```
program--, line 24: Are discretizations only associated to catchments, and are those catchment's divide realizations the return of this member?

_if not_ (and this is mostly semantics), I wouldn't call this a `divide`, since that term is specific to catchment boundaries. If the intention of this member is to retrieve the geographic domain of discretizations over arbitrary geographic domains, this should be probably be called `extent` and return something like a `GeoExtent`.

PhilMiller: Good catch on the semantics. I think `extent` is right, since I intend it to be possible for a `Discretization` to cover the entirety of a gridded model's operational domain, spanning distinct catchments.

Issue #562, opened 2023-07-06
Remove internal ET capability


mattw-nws:
Key stakeholders have indicated that ET calculation should not be a framework capability, there is a [BMI module that serves this purpose](https://github.com/NOAA-OWP/evapotranspiration), and use of the internal ET calc method is actively discouraged. The requirement to continue to support it for legacy purposes is causing unnecessary zombie code to creep into major architectural features like #497, increasing complexity and reducing maintainability. This is a roadmap issue to discuss removing the internal ET support in a near-future breaking change.

## Current behavior

ET calculation is a 1st-class feature provided to models including legacy and BMI modules, including significant pipelining code within BMI formulations.

## Expected behavior

ET calc code and the pipelining is removed from the architecture.

## Steps to replicate behavior (include URLs)

1.

PR #594, opened 2023-08-02
Refactor tests to remove dependencies on internal models

mattw-nws:
Shifts unit tests to use BMI test mock models instead of legacy internal models. This lays the groundwork for #562 and additional legacy code removal

## Additions

- `test_bmi_cpp` is now used for several unit tests in Formulation_Manager_Test.cpp
- The cmake build system now automatically builds `test_bmi_cpp` without any additional flags
- Formulation_Manager_Test now uses some `#define`s from Bmi_Formulation.hpp and AorcForcing.hpp (the latter was the case already but it wasn't explicitly included)

## Removals

- Formulation_Manager_Test no longer uses `tshirt`, `tshirt_c`, `simple_lumped` or anything that requires `et_calc`

## Changes

- `test_bmi_cpp` is now built as a dependency. The build should be the same as the usual way (e.g. the artifacts go to the same place) and shouldn't interfere with manual builds either.

## Testing

1. All automated tests pass, separately tested with the removed components removed from the codebase.

## Notes

-

## Todos

- Possibly allow C++ BMI formulations to be directly compiled in and use `Bmi_Formulation` instead of `Bmi_Module_Formulation`, which would remove the required weirdness of building the plugin shared library version of test_bmi_cpp.

## Checklist

1.

- [X] Linux


test/realizations/Formulation_Manager_Test.cpp
```diff
@@ -248,49 +230,28 @@ const std::string EXAMPLE_1 = "{ "
...
+                "\"destroy_function\": \"bmi_model_destroy\","
+                "\"uses_forcing_file\": false"
+              "} "
             "} "
-          "} "
-        "} "
       "], "
       "\"forcing\": { "
-          "\"file_pattern\": \".*{{ID}}.*.csv\", "
+          "\"file_pattern\": \".*{{ID}}.*.csv\", " // DIFF from Ex.1
```
PhilMiller, line 254: Are these comments placed to get `git diff` to behave sensibly? That's the only possibility I can infer from them

extern/test_bmi_cpp/CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ target_include_directories(testbmicppmodel PRIVATE ../bmi-cxx/ )
 set_target_properties(testbmicppmodel PROPERTIES VERSION ${PROJECT_VERSION})
 
 set_target_properties(testbmicppmodel PROPERTIES PUBLIC_HEADER include/test_bmi_cpp.hpp)
+set_target_properties(testbmicppmodel PROPERTIES CXX_VISIBILITY_PRESET default)
```
PhilMiller, line 30: From my perspective, this is a stop-gap toward actually export-marking the relevant exported symbols explicitly, to demonstrate the practice that we should be encouraging from real models that we integrate.

PhilMiller:
Not for this PR, but those embedded configurations would be an ideal use of C++ raw string literals
https://en.cppreference.com/w/cpp/language/string_literal

test/realizations/Formulation_Manager_Test.cpp
mattw-nws: Ah... those are I suppose vestiges from a676c5d, where I was trying to capture how the examples differed from one another, to determine what the tests were actually testing and hence what needed to be preserved. Still somewhat useful as road markers, since the four giant strings aren't easily comparable. (I recommend cutting example 1 and pasting it after example 4 if you want, that's how I compared them to get those notes.)

extern/test_bmi_cpp/CMakeLists.txt
mattw-nws: Yes... to elaborate, when I added `add_subdirectory(test_bmi_cpp...)` to CMakeLists.txt, it was compiling with `-fvisibility=hidden` for reasons that weren't fully clear... it also wasn't fully clear why it _wasn't_ doing that when building test_bmi_cpp with cmake standalone. Arguably, including that flag and having the desired public symbols explicitly marked (as @PhilMiller suggests) are the best practice, though there's little or nothing in this test module that isn't intended to be visible.

PhilMiller (approved):

PR #600, opened 2023-08-02
Remove native/internal LSTM model

mattw-nws:
Builds on/requires #599

## Additions

- None

## Removals

- LSTM native/internal model, associated tests and data files.

## Changes

- Removals from some CMakeLists.txt files

## Testing

1. Passes all automated tests

## Notes

-

## Todos

-

## Checklist

1.

- [X] Linux


PhilMiller (approved):

hellkite500:
Once we get all these deprecations in place, we should ensure each is represented in a single commit, then document the commit and the deprecation in a changelog.

PhilMiller (approved):

jmframe:
Looks like this removes the C++ LSTM. Have y'all been running this version? If so, were you also running the Python version? Just out of curiosity.

mattw-nws:
@jmframe No, we haven't in a long time, that's part of the reason for the removal--I'm not even sure if it still works. The API around the formulations has been evolving and we kept having to come back to the old non-BMI models and do a lot of retrofitting--so for the time being we're removing them to streamline things, because that API is probably going to need to further evolve quite a bit for a while to deal with multiple discretizations (structured grid models, catchment models, network/routing modules, unstructured grid models...).

I think there's space for non-BMI formulations in the future for various reasons including performance optimization (relevant to LSTM, probably), but the technical debt associated with the upkeep of things that weren't directly related to the capabilities we're being asked to develop was becoming costly. I'd like this to come back in the future after the internal APIs have stabilized more.

mattw-nws:
Oh.. above was WRT the C version. The Python/BMI version is still being kept up by the formulations team and I think they're trying to build some model pieces to produce non-streamflow variables alongside it when it is used, for instance.

jmframe:
Sounds good. That's what I figured.

PR #597, opened 2023-08-02
Removes top-level tshirt models

mattw-nws:
Builds on/requires #595

Removes tshirt, tshirt_c, associated data and test files, and the Bmi_C_Cfe_IT test (which depended on tshirt).

## Additions

- None

## Removals

- tshirt, tshirt_c, associated data and test files, and the Bmi_C_Cfe_IT test (which depended on tshirt).

## Changes

- CMakeLists.txt removals for the legacy components, e.g. the `compare_cfe` target is now gone.

## Testing

1.

## Notes

-

## Todos

- Documentation and example file revisions will wait till the end of the PR chain

## Checklist

- [X] PR has an informative and human-readable title
- [X] Changes are limited to a single goal (no scope creep)
- [X] Code can be automatically merged (no conflicts)
- [X] Code follows project standards (link if applicable)
- [X] Passes all existing automated tests
- [X] Any _change_ in functionality is tested
- [X] New functions are documented (with a description, list of inputs, and expected output)
- [X] Placeholder code is flagged / future todos are captured in comments
- [X] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [X] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

## Testing checklist (automated report can be put here)

1.

### Target Environment support

- [X] Linux

PR #606, opened 2023-08-07
Strictly follow C++14 standard

PhilMiller:
Since we're not testing with the Intel classic `icpc` compiler that we're currently expecting to have to support in production, we should at least ask GCC/Clang to be strict about making sure we're not using language extensions that it might not offer, or that it potentially implements differently.

## Changes

- Change all instances of stack-allocated VLA usage to `std::vector`
- Add `-pedantic-error` to CI build process to prevent future such errors

## Checklist

### DIFF
--- a/.github/actions/ngen-build/action.yaml
+++ b/.github/actions/ngen-build/action.yaml
@@ -166,7 +166,7 @@ runs:
           run: |
             export BOOST_ROOT="$(pwd)/boost_1_72_0"
             export CFLAGS="-fsanitize=address -O1 -g -fno-omit-frame-pointer"
-            export CXXFLAGS="-fsanitize=address -O1 -g -fno-omit-frame-pointer"
+            export CXXFLAGS="-fsanitize=address -O1 -g -fno-omit-frame-pointer -pedantic-errors"
             . .venv/bin/activate
             [ ! -d "$BOOST_ROOT" ] && echo "Error: no Boost root found at $BOOST_ROOT" && exit 1
             cmake -B ${{ inputs.build-dir }} \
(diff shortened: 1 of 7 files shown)


PhilMiller:
I just checked that there are no compiler warnings across the codebase right now. We could actually add `-Werror` on our CI builds!

If we do that, we'd want to specify the OS releases to be used, rather than `ubuntu-latest`/`macos-latest`, so that a new release wouldn't unexpectedly introduce warnings and hence job failures in a PR that doesn't change anything in question.

program--:
In line with this PR, would it make sense to turn off extensions at the build system level? i.e. with `set_target_properties(... PROPERTIES CXX_EXTENSIONS OFF)` on all our CMake targets, or with `set(CMAKE_CXX_EXTENSIONS OFF)` globally?

PhilMiller:
The CMake change might be a good idea.

If I understand correctly, the `-pedantic` compiler flag only affects compiler behavior, while `-std=c++14` vs `-std=gnu++14` would potentially also affect library stuff as well via definition of macros like `_GNU_SOURCE`. Given that we test on both GNU/Linux and macOS, I'm not too worried that we're accidentally dependent on GNU-isms in the system C library.

On the language side, even with `-std=c++14` I think GCC may still accept some C++ constructs that are either first specified in C++17, or are not actually in the standard, but that they don't expect to conflict with later standard text. The `-pedantic` flag gets at that.

PhilMiller:
Bottom line, let's look at the CMake/`-std` settings in another PR

mattw-nws:
> Bottom line, let's look at the CMake/`-std` settings in another PR

Probably so but I think I'm in favor of making comparable changes to CMakeLists.txt, yes. Better for cmake to be telling us the problem while we're developing then find out when you push a PR.

mattw-nws:
And @PhilMiller you've got a conflict now after #597 .

PhilMiller:
> > Bottom line, let's look at the CMake/`-std` settings in another PR
> 
> Probably so but I think I'm in favor of making comparable changes to CMakeLists.txt, yes. Better for cmake to be telling us the problem while we're developing then find out when you push a PR.

I'm consciously starting with the approach here of enforcing these checks in CI rather than on individual developers' systems because the exact nature of warnings and what will or won't trigger them can vary across compiler releases and quirks of library versions and optimization settings. Developers' machines are heterogeneous moving targets, while CI is shared and more stationary.

PR #610, opened 2023-08-15
Fix a bug in output variable name list, replacing (PR#577)

stcui007:
This PR fixing variable name entry issue#535

## Additions

-

## Removals

-

## Changes

include/realizations/catchment/Bmi_Multi_Formulation.hpp
src/realizations/catchment/Bmi_Multi_Formulation.cpp

## Testing

Unit Tests
Run ngen tests

## Notes

-

## Todos

-

## Checklist

- [x ] New functions are documented (with a description, list of inputs, and expected output)

1.

- [x ] Linux

### DIFF
--- a/include/realizations/catchment/Bmi_Multi_Formulation.hpp
+++ b/include/realizations/catchment/Bmi_Multi_Formulation.hpp
@@ -580,6 +580,40 @@ namespace realization {
             }
         }
 
+        /** 
+         * Check that the output variable names in the global bmi_multi are valid names
+         */
+        void check_output_var_names() {
+            // get bmi_multi global output variable names
+            const std::vector<std::string> &output_var_names = get_output_variable_names();
+
+            // store variable names in avaialbleData in the valid output variable name list
+            std::vector<std::string> available_var_names;
+            std::string var_name;
+            for(std::map<std::string,std::shared_ptr<data_access::GenericDataProvider>>::iterator iter = availableData.begin(); iter != availableData.end(); ++iter)
+            {
+                var_name = iter->first;
+                available_var_names.push_back(var_name);
+            }
+
+            // take into account the unit test where both the original and remapped output var names are used in the test codes
+            for (const nested_module_ptr &module: modules) {
+                for (const std::string &out_var_name: module->get_bmi_output_variables()) {
+                    if ((out_var_name == "OUTPUT_VAR_1") || (out_var_name == "OUTPUT_VAR_2") || (out_var_name == "OUTPUT_VAR_3")) {
+                        available_var_names.push_back(out_var_name);
+                    }
+                }
+            }
+
+            // check if an output variable listed in the bmi_multi global is a valid variable
+            for (int i = 0; i < output_var_names.size(); ++i) {
+                auto it = std::find(available_var_names.begin(), available_var_names.end(), output_var_names[i]);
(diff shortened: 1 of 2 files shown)


include/realizations/catchment/Bmi_Multi_Formulation.hpp
```diff
@@ -580,6 +580,40 @@ namespace realization {
...
+            for(std::map<std::string,std::shared_ptr<data_access::GenericDataProvider>>::iterator iter = availableData.begin(); iter != availableData.end(); ++iter)
+            {
+                var_name = iter->first;
+                available_var_names.push_back(var_name);
+            }
+
+            // take into account the unit test where both the original and remapped output var names are used in the test codes
+            for (const nested_module_ptr &module: modules) {
+                for (const std::string &out_var_name: module->get_bmi_output_variables()) {
+                    if ((out_var_name == "OUTPUT_VAR_1") || (out_var_name == "OUTPUT_VAR_2") || (out_var_name == "OUTPUT_VAR_3")) {
```
mattw-nws, line 602: I'm not sure what's going on in the unit test (I'll try to take a closer look today), but putting these "magic strings" in the production code to account for behavior in the unit test isn't the right answer. Either the test is wrong, or the behavior needs to match the test.

mattw-nws (changes requested):
This is indeed cleaner, but it goes back to only fixing the problem for `Bmi_Multi_Formulation` where the problem in #535 can occur in any BMI formulation. This should be possible to do up in `Bmi_Formulation` or at least with only a little duplicate code in `Bmi_Multi_Formulation` and `Bmi_Module_Formulation`

mattw-nws (changes requested):
This is indeed cleaner, but it goes back to only fixing the problem for `Bmi_Multi_Formulation` where the problem in #535 can occur in any BMI formulation. This should be possible to do up in `Bmi_Formulation` or at least with only a little duplicate code in `Bmi_Multi_Formulation` and `Bmi_Module_Formulation`

stcui007:
This is the only place where all the data are accessible. The availableData
variable actually contains all variables in the sub-modules as well as
input forcing, so nothing is left out. If you run test on the example
realization files you provided, they all work as they are supposed to.
Somehow, in Bmi_Formulation or Bmi_Module_formulation, not all data are
accessible at the same time, so a validity comparison could not be made.

On the unit test part, I agree that it does look a bit awkward. But the
unit test code uses both original and remapped output variables in the
codes. Since the original output variables is already remapped, it does not
exist anymore, so the ngen code generate a "throw" by design. It seems to
be trapped.

src/realizations/catchment/Bmi_Multi_Formulation.cpp
```diff
@@ -292,6 +292,11 @@ string Bmi_Multi_Formulation::get_output_line_for_timestep(int timestep, std::st
         throw std::invalid_argument("Only current time step valid when getting multi-module BMI formulation output");
     }
 
+    // check if a requested output variable name is valid, if not, stop the execution
+    if (timestep == 0) {
+        check_output_var_names();
```
mattw-nws, line 297: I think this will work if this function is called at the end of `create_multi_formulation` instead of in the time loop... can you try it there instead?

stcui007:
Sure will do.

mattw-nws (approved):

PR #613, opened 2023-08-17
Tests and cleanup for #610 (`output_variables` fallback behavior)

mattw-nws:

## Additions

- Two tests for invalid entries in `output_variables`

## Removals

- Obsolete fallback condition code in Bmi_Multi_Formulation.cpp

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

1.

- [X] Linux

### DIFF
--- a/src/realizations/catchment/Bmi_Multi_Formulation.cpp
+++ b/src/realizations/catchment/Bmi_Multi_Formulation.cpp
@@ -311,7 +311,7 @@ string Bmi_Multi_Formulation::get_output_line_for_timestep(int timestep, std::st
 
         // Do the rest with a leading comma
         for (int i = 1; i < output_var_names.size(); ++i) {
-            *output_text_stream << "," << get_var_value_as_double(output_var_names[i]);
+            *output_text_stream << delimiter << get_var_value_as_double(output_var_names[i]);
         }
         return output_text_stream->str();
     }
--- a/src/realizations/catchment/Bmi_Multi_Formulation.cpp
+++ b/src/realizations/catchment/Bmi_Multi_Formulation.cpp
@@ -307,7 +307,6 @@ string Bmi_Multi_Formulation::get_output_line_for_timestep(int timestep, std::st
         if (output_var_names.empty()) { return ""; }
 
         // Do the first separately, without the leading comma
-        auto output_data_provider_iter = availableData.find(output_var_names[0]);
         *output_text_stream << get_var_value_as_double(output_var_names[0]);
 
         // Do the rest with a leading comma
--- a/src/realizations/catchment/Bmi_Multi_Formulation.cpp
+++ b/src/realizations/catchment/Bmi_Multi_Formulation.cpp
@@ -306,23 +306,15 @@ string Bmi_Multi_Formulation::get_output_line_for_timestep(int timestep, std::st
         // This almost certainly should never happen, but just to be safe ...
         if (output_var_names.empty()) { return ""; }
 
-        try {
-            // Do the first separately, without the leading comma
-            auto output_data_provider_iter = availableData.find(output_var_names[0]);
-            *output_text_stream << get_var_value_as_double(output_var_names[0]);
+        // Do the first separately, without the leading comma
+        auto output_data_provider_iter = availableData.find(output_var_names[0]);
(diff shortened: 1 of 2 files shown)


mattw-nws:
As a side note, the test logic in Bmi_Multi_Formulation_Test.cpp is probably overengineered at this point... we need tests to make sure the test code produces what it's supposed to. Seems like there should be an easier way.

stcui007:
I only have some rather trivial comments. On line 486 of Bmi_Multi_Formulation_Test.cpp, the comment line, the number of example scenarios should now be 6?

In the same file, line 635-636, it is probably better to remove one of the empty lines.

stcui007 (approved):

src/realizations/catchment/Bmi_Multi_Formulation.cpp
```diff
@@ -306,23 +306,15 @@ string Bmi_Multi_Formulation::get_output_line_for_timestep(int timestep, std::st
...
-            }
-            return output_text_stream->str();
-        }
-        catch (const std::exception &e) {
-            std::cerr << "WARN: " << e.what()
-                      << "; reverting to default behavior for multi-BMI formulation type (using last module)";
-            output_text_stream->str(std::string()); // ... clear any output contents being staged ...
-            is_out_vars_from_last_mod = true;          // ... revert to default behavior (just use last nested module)
+        // Do the first separately, without the leading comma
+        auto output_data_provider_iter = availableData.find(output_var_names[0]);
```
PhilMiller, line 310: This exists in the original code, but `output_data_provider_iter` is not actually tested or used through the rest of this function. That makes me concerned that the overall logic is (still) flawed.

stcui007: The ```output_data_provider_iter``` was actually used in ```check_output_var_names()``` function which was originally called from within this function. After unit test, the call to ```check_output_var_names()``` was moved to line 142 which is probably a more appropriate place.

mattw-nws: The `.find()` may have been left there to trigger the catch block that was removed... but in any case it's leftover code from 1a29230 that did the digging for each value before `get_var_value_as_double` existed, it appears. I'll drop this real quick since it's dead code and also not cost-free.

src/realizations/catchment/Bmi_Multi_Formulation.cpp
```diff
@@ -306,23 +306,14 @@ string Bmi_Multi_Formulation::get_output_line_for_timestep(int timestep, std::st
...
-            std::cerr << "WARN: " << e.what()
-                      << "; reverting to default behavior for multi-BMI formulation type (using last module)";
-            output_text_stream->str(std::string()); // ... clear any output contents being staged ...
-            is_out_vars_from_last_mod = true;          // ... revert to default behavior (just use last nested module)
+        // Do the first separately, without the leading comma
+        *output_text_stream << get_var_value_as_double(output_var_names[0]);
+
+        // Do the rest with a leading comma
+        for (int i = 1; i < output_var_names.size(); ++i) {
+            *output_text_stream << "," << get_var_value_as_double(output_var_names[i]);
```
PhilMiller, line 314:
```suggestion
            *output_text_stream << delimiter << get_var_value_as_double(output_var_names[i]);
```
I just noticed an apparent bug/oversight in this function - the `delimiter` argument is just entirely unused.

At a broader level, it makes *no* sense to me that the BMI formulation class is doing string formatting of the variables being output, rather than just returning their values to be passed to an output routine.

PhilMiller (approved):
Looks fine to me.

stcui007 (approved):
The new change makes the codes in the function more generalized. Looks good to me.

src/realizations/catchment/Bmi_Multi_Formulation.cpp
mattw-nws: Yeah I am inferring that this goes back to the earliest prototypes. The mdframe PR #533 will help with this... however... we actually become memory-bound when running large simulations so storing the whole result in memory and dumping it out at the end will not be feasible, we'll have do develop some kind of streaming output system... which this at least accomplishes. That is probably why we're still doing this.

PR #615, opened 2023-08-18
Multi level hydrofabric

donaldwj:

## Additions

- Support for multiple processing levels
- Support for separate models runtimes for each level
-

## Removals

- Old time loop has been replaced
-

## Changes

- Added Layer Objects to Ngen driver
- Layer objects now manage model realizations

## Testing

Manual Test of 5 catchment data set

## Notes

-

## Todos

-

## Checklist

1.

- [ ] Linux


include/core/Layer.hpp
```diff
@@ -0,0 +1,141 @@
...
+                    response *= (catchment_data->get_feature(id)->get_property("areasqkm").as_real_number() * 1000000);
+                }catch(std::invalid_argument &e)
+                {
+                    response *= (catchment_data->get_feature(id)->get_property("area_sqkm").as_real_number() * 1000000);
+                }
+                //TODO put this somewhere else as well, for now, an implicit assumption is that a modules get_response returns
+                //m/timestep
+                //since we are operating on a 1 hour (3600s) dt, we need to scale the output appropriately
+                //so no response is m^2/hr...m^2/hr * 1hr/3600s = m^3/hr
+                response /= 3600.0;
```
mattw-nws, line 110: So, there was no great place to have a timestep variable before... but now we have a Simulation_Time object with an `output_interval`... even if we assume that the output is m/h (not m/timestep!), shouldn't this response be divided by the number of timesteps per hour if the timestep is not = 1?

donaldwj: this was fixed in my branch

include/core/Layer.hpp
```diff
@@ -0,0 +1,141 @@
...
+            if(output_time_index%100 == 0) std::cout<<"Running timestep " << output_time_index <<std::endl;
+            std::string current_timestamp = simulation_time.get_timestamp(output_time_index);
+            for(const auto& id : processing_units) 
+            {
+                int sub_time = output_time_index;
+                //std::cout<<"Running cat "<<id<<std::endl;
+                auto r = features.catchment_at(id);
+                //TODO redesign to avoid this cast
+                auto r_c = std::dynamic_pointer_cast<realization::Catchment_Formulation>(r);
+                double response = r_c->get_response(output_time_index, description.time_step);
```
mattw-nws, line 95: will `description.time_step` and `simulation_time.get_output_interval_seconds()` ever differ? Are they redundant or is there a current/potential future semantic difference?

mattw-nws: Actually since it looks like `description` has a `time_step_units`, it's likely that `description.time_step` should not be used here and `simulation_time.get_output_interval_seconds()` should be used instead, or that `description.time_step` should be put through a unit conversion first (probably the former).

include/core/Layer.hpp
```diff
@@ -20,9 +20,9 @@ namespace ngen
         public:
 
         #ifdef NGEN_MPI_ACTIVE
-            typedef hy_features::HY_Features_MPI feature_type;
+            using feature_type = hy_features::HY_Features_MPI;
         #else
-            typedef hy_features::HY_Features feature_type;
+            typedef feature_type = hy_features::HY_Features;
```
mattw-nws: Trying to work out the rebase, and when I get to this commit (about), this fails on GCC8, saying `feature_type` does not define a type... when I switch this to `using` instead of `typedef`, it seems to compile. Is there a particular reason the above is `using` and this is `typedef`? Or is it just an oversight? If the latter I can amend it in that commit. If the difference is deliberate, we can figure out what's going on after the rebase.

.gitignore
```diff
@@ -97,7 +97,6 @@ CMakeCache.txt
 /cmake-build-debug
 /cmake_build
 /build
-/cmake_build*
```
PhilMiller, line 100: PR should leave `.gitignore` alone - it's out of scope.

include/core/HY_Features.hpp
```diff
@@ -59,24 +60,14 @@ namespace hy_features {
         /**
          * @brief Construct a new HY_Features object from a Network and a set of formulations.
          * 
-         * Constructs the HY_Catchment objects for each catchment feature in the network, and attaches tha formaulation
+         * Constructs the HY_Catchment objects for each catchment connecting them with the provided link_key attaches tha formaulation
```
PhilMiller, line 63: The grammar is the revised comment doesn't make sense anymore.

Also,
```suggestion
         * Constructs the HY_Catchment objects for each catchment connecting them with the provided link_key attaches the formulation
```

include/core/Layer.hpp
```diff
@@ -0,0 +1,147 @@
...
+                }
+                double response_m_s = response * (area * 1000000);
+                //TODO put this somewhere else as well, for now, an implicit assumption is that a module's get_response returns
+                //m/timestep
+                //since we are operating on a 1 hour (3600s) dt, we need to scale the output appropriately
+                //so no response is m^2/hr...m^2/hr * 1hr/3600s = m^3/hr
+                double response_m_h = response_m_s / 3600.0;
+                //update the nexus with this flow
+                for(auto& nexus : features.destination_nexuses(id)) {
+                    //TODO in a DENDRIDIC network, only one destination nexus per catchment
```
PhilMiller, line 116:
```suggestion
                    //TODO in a DENDRITIC network, only one destination nexus per catchment
```

include/core/Layer.hpp
```diff
@@ -0,0 +1,147 @@
...
+        virtual ~Layer() {}
+
+        /***
+         * @brief Return the next timestep that will be processed by this layer in epoc time units
+        */
+        time_t next_timestep_epoch_time() { return simulation_time.next_timestep_epoch_time(); }
+
+
+        /***
+         * @brief Return the last timesteo that was processed by this layer in epoc time units
```
PhilMiller, line 52:
```suggestion
         * @brief Return the next timestep that will be processed by this layer in epoch time units
        */
        time_t next_timestep_epoch_time() { return simulation_time.next_timestep_epoch_time(); }


        /***
         * @brief Return the last timestep that was processed by this layer in epoch time units
```

src/NGen.cpp
```diff
@@ -333,74 +337,105 @@ int main(int argc, char *argv[]) {
...
+    for( int count = 0; count < num_times; count++) 
+    {
+      // The Inner loop will advance all layers unless doing so will break one of two constraints
+      // 1) A layer may not proceed ahead of the master simulation object's current time
+      // 2) A layer may not proceed ahead of any layer that is computed before it
+      // The do while loop ensures that all layers are tested at least once while allowing 
+      // layers with small time steps to be updated more than once
+      // If a layer with a large time step is after a layer with a small time step the
+      // layer with the large time step will wait for multiple timesteps from the preceeding
+      // layer.
```
PhilMiller, line 400: Since we're not (currently) doing temporal interpolation, should we be asserting that the various time step increments are all (close enough to) integer ratios, so that we're never left with a layer/model at an unaligned time?

mattw-nws:
@donaldwj Do you have any uncommitted/unpushed updates to this PR? I'm going to rebase on master again to make sure things work and clean up a few of the minor issues.

src/core/HY_Features.cpp
```diff
@@ -3,59 +3,67 @@
...
+      formulation->set_output_stream(formulations->get_output_root() + feat_id + ".csv"); // MERGE bd63e5b
+      // TODO: add command line or config option to have this be omitted
+      //FIXME why isn't default param working here??? get_output_header_line() fails.
+      formulation->write_output("Time Step,""Time,"+formulation->get_output_header_line(",")+"\n");
+      //Find upstream nexus ids
+      origins = network.get_origination_ids(feat_id);
+
+      // get the catchment level from the hydro fabric
+      const auto& cat_json_node = fabric->get_feature(feat_id);
+      long lv = cat_json_node->has_key("level") ? cat_json_node->get_property("level").as_natural_number() : 0;
```
mattw-nws, line 43: I was going through and unifying "level" to "layer" for consistency... but this appears it would actually have a functional impact, because the provided /data/catchment_data_multilayer.geojson file actually _does_ have `"layer"` keys instead of `"level"`... of course the test code was previously looking for a different example file which was not committed, so I don't know what was in that one.  @donaldwj can you look at this please? I will test some with this but am not as confident I'll figure out the impact.

src/core/HY_Features_MPI.cpp
```diff
@@ -34,17 +34,28 @@ HY_Features_MPI::HY_Features_MPI( PartitionData partition_data, geojson::GeoJSON
...
           //Find and prepare formulation
           auto formulation = formulations->get_formulation(feat_id);
           formulation->set_output_stream(formulations->get_output_root() + feat_id + ".csv");
           // TODO: add command line or config option to have this be omitted
           //FIXME why isn't default param working here??? get_output_header_line() fails.
           formulation->write_output("Time Step,""Time,"+formulation->get_output_header_line(",")+"\n");
+          
+          // get the catchment level from the hydro fabric
+          const auto& cat_json_node = linked_hydro_fabric->get_feature(feat_id);
+          long lv = cat_json_node->has_key("level") ? cat_json_node->get_property("level").as_natural_number() : 0;
```
mattw-nws, line 48: Ditto for comment in non-MPI version of this file, @donaldwj

src/core/network.cpp
```diff
@@ -43,6 +44,20 @@ Network::Network( geojson::GeoJSON fabric ){
     else{
       v1 = this->descriptor_map[ feature_id];
     }
+
+    if ( this->level_map.find(feature_id) == this->level_map.end() )
+    {
+      if ( feature->has_property("level") )
+      {
+        const auto& prop = feature->get_property("level");
```
mattw-nws, line 52: This again doesn't match the data/catchment_data_multilayer.geojson file @donaldwj

src/core/HY_Features.cpp
mattw-nws: After changing this, it _also_ passes the included test with the example file that has `"layer"`... for better and worse.

hellkite500 (approved):
Going to bite the bullet and merge this, we can come back and address various aspects as needed.

PR #617, opened 2023-08-22
Formulation `record_duration()` derived from last t_delta

mattw-nws:
Attempts to actually implement `record_duration()` in Bmi_Module_Formulation and Bmi_Multi_Formulation by capturing the last t_delta passed to get_response.

## Additions

- Impelmentation of `record_duration()` in Bmi_Module_Formulation and Bmi_Multi_Formulation

## Removals

- Heuristics in Bmi_Multi_Formulation to find a `record_duration()` from one of its nested `DataProvider`s

## Changes

-

## Testing

1.

## Notes

-

## Todos

- Test results in Bmi_Multi_Formulation_Test are confusing (concerning?) as it looks like perhaps temporal interpolation happens differently between those two scenarios. Might be a difference deriving from the formulation somehow though, needs more research.

## Checklist

1.

- [ ] Linux

### DIFF
--- a/test/realizations/catchments/Bmi_Cpp_Formulation_Test.cpp
+++ b/test/realizations/catchments/Bmi_Cpp_Formulation_Test.cpp
@@ -272,8 +272,41 @@ TEST_F(Bmi_Cpp_Formulation_Test, GetResponse_0_b) {
     double response;
     for (int i = 0; i < 39; i++) {
         response = formulation.get_response(i, 3600);
+        //std::cout << "i: " << i << "  response: " << response << std::endl;
     }
     double expected = 2.7809780039160068e-08;
+    ASSERT_EQ(expected, response);
+}
+
+/**
+ * Test with non-3600 timestep, smoke test and ensure record_duration reflects valid value.
+ */
+TEST_F(Bmi_Cpp_Formulation_Test, GetResponse_0_c) {
+    int ex_index = 0;
+
+    Bmi_Cpp_Formulation formulation(catchment_ids[ex_index], std::make_unique<CsvPerFeatureForcingProvider>(*forcing_params_examples[ex_index]), utils::StreamHandler());
+    formulation.create_formulation(config_prop_ptree[ex_index]);
+
+    formulation.get_response(0, 1800);
+    ASSERT_EQ(formulation.record_duration(), 1800);
+}
+
+/**
+ * Test with non-3600 timestep, smoke test and ensure record_duration reflects valid value.
+ */
+//TODO: Understand why this is? This is just a test model, but the results are non-intuitive.
+TEST_F(Bmi_Cpp_Formulation_Test, GetResponse_0_d) {
+    int ex_index = 0;
+
+    Bmi_Cpp_Formulation formulation(catchment_ids[ex_index], std::make_unique<CsvPerFeatureForcingProvider>(*forcing_params_examples[ex_index]), utils::StreamHandler());
+    formulation.create_formulation(config_prop_ptree[ex_index]);
+
+    double response;
+    for (int i = 0; i < 78; i++) {
+        response = formulation.get_response(i, 1800);
+        //std::cout << "i: " << i << "  response: " << response << std::endl;
+    }
+    double expected = 1.3904890019580034e-08; // Half of test GetResponse_0_b, which is sensible.
     ASSERT_EQ(expected, response);
 }
 
--- a/test/realizations/catchments/Bmi_Multi_Formulation_Test.cpp
+++ b/test/realizations/catchments/Bmi_Multi_Formulation_Test.cpp
@@ -643,8 +643,41 @@ TEST_F(Bmi_Multi_Formulation_Test, GetResponse_0_b) {
     double response;
     for (int i = 0; i < 39; i++) {
         response = formulation.get_response(i, 3600);
+        //std::cout << "i: " << i << "  response: " << response << std::endl;
     }
     double expected = 2.7809780039160068e-08;
+    ASSERT_EQ(expected, response);
+}
+
+/**
+ * Test with non-3600 timestep, smoke test and ensure record_duration reflects valid value.
+ */
+TEST_F(Bmi_Multi_Formulation_Test, GetResponse_0_c) {
+    int ex_index = 0;
+
+    Bmi_Multi_Formulation formulation(catchment_ids[ex_index], std::make_unique<CsvPerFeatureForcingProvider>(*forcing_params_examples[ex_index]), utils::StreamHandler());
+    formulation.create_formulation(config_prop_ptree[ex_index]);
+
+    double response = formulation.get_response(0, 1800);
+    ASSERT_EQ(response, 00);
+    ASSERT_EQ(formulation.record_duration(), 1800);
+}
+
+/**
+ * Test with non-3600 timestep in example 0 after several iterations.
+ */
+TEST_F(Bmi_Multi_Formulation_Test, GetResponse_0_d) {
+    int ex_index = 0;
+
+    Bmi_Multi_Formulation formulation(catchment_ids[ex_index], std::make_unique<CsvPerFeatureForcingProvider>(*forcing_params_examples[ex_index]), utils::StreamHandler());
+    formulation.create_formulation(config_prop_ptree[ex_index]);
+
+    double response;
+    for (int i = 0; i < 78; i++) {
+        response = formulation.get_response(i, 1800);
+        //std::cout << "i: " << i << "  response: " << response << std::endl;
+    }
+    double expected = 1.3904890019580034e-08;
     ASSERT_EQ(expected, response);
 }
 


PhilMiller:
The code seems reasonable as an implementation of what it says. The whole idea feels really questionable to me, though. Any given source of data should actually be able to tell what interval that data is valid for or (for rates) extends over.

BMI modules have `get_time_step` and `get_time_units`, but as I understand it, we're trying to account for the engine multi-stepping a module?

mattw-nws:
> BMI modules have get_time_step and get_time_units, but as I understand it, we're trying to account for the engine multi-stepping a module?

You've about got it. Multi-stepping or fractional-stepping, really. So, many models may just choose an arbitrary value if we used get_time_step, because they can model based on any advance of time... others may have to run a specific timestep (perhaps even a variable one) to maintain stability... for a forcing module publishing with BMI, the native timestep is very much _not_ arbitrary. Right now we require modules to support `update_until` sensibly, which means for non-rate quantities (e.g. water output) they should compute the total since the last `t` passed to `update_until(t)`. For rates, or something like temperature, it could be an instantaneous value. Whether that instantaneous value is interpolated in any way if the `delta_t` thresholds don't match some underlying "native" timestep is largely up to the model--and there's no way in the interface to communicate that either. This opens a number of questions if we start thinking hard about things. But given the uncertainty inherent in what we're modeling, it probably doesn't likely matter much.

More generally in response to your comment.... yeah I was wanting to have some discussion about what the formulation's timestep/record_duration should represent and what it should be used for. Because I'm not sure we know yet.

program--:
Similar to what @PhilMiller said, I'm wondering whether the formulation should be handling time management whatsoever... My reasoning is that, at least in my head, a `simulation` and a `formulation` should be different, such that a `formulation` is stateless, and is almost like a "recipe" that could be used in any given number of `simulation`s (i.e. formulations should be composed within a simulation). The current design seems to share the responsibility of time management between `Simulation_Time`, a `Formulation`, and any surrounding executing code.

In an ideal situation, I'd think the user would give a temporal domain, and `Simulation_Time` (or some other simulation object) would be responsible for orchestrating models/formulations over that temporal domain, taking into account different model strides. (for example: step along the domain at the shortest stride of all the formulations)

> Hopefully that makes sense, and hopefully it is relevant here in this discussion.

mattw-nws:
> My reasoning is that, at least in my head, a `simulation` and a `formulation` should be different, such that a `formulation` is stateless, and is almost like a "recipe" that could be used in any given number of `simulation`s ...

@program-- It's relevant. These objects might be poorly named, or that might be part of what's going on here. The names are "formulation"s, but they descend from `HY_CatchmentRealization`... I think formulation is the domain science term and "realization" comes from HY_Features... so these are like _instances_ of a formulation, in practice. @hellkite500 might be able to speak a little more on what "realization" means as I'm still fuzzy on the abstract distinctions. But instances of these objects are definitely instances of a formulation, not just a formulation recipe.... there's an instance tied to a specific catchment, and multiple instances (sometimes of the same formulation recipe) exist--one for each catchment--and they each are composed of an instance of a BMI model that definitely has state--including some notion of the current time. Does that affect your thinking any?

program--:
> > My reasoning is that, at least in my head, a `simulation` and a `formulation` should be different, such that a `formulation` is stateless, and is almost like a "recipe" that could be used in any given number of `simulation`s ...
> 
> @program-- It's relevant. These objects might be poorly named, or that might be part of what's going on here. The names are "formulation"s, but they descend from `HY_CatchmentRealization`... I think formulation is the domain science term and "realization" comes from HY_Features... so these are like _instances_ of a formulation, in practice. @hellkite500 might be able to speak a little more on what "realization" means as I'm still fuzzy on the abstract distinctions. But instances of these objects are definitely instances of a formulation, not just a formulation recipe.... there's an instance tied to a specific catchment, and multiple instances (sometimes of the same formulation recipe) exist--one for each catchment--and they each are composed of an instance of a BMI model that definitely has state--including some notion of the current time. Does that affect your thinking any?

@mattw-nws Yeah that does a bit. I think the abstractions we have at play (or at least should consider/differentiate) are these:
- **Realizations**, which are the topological representations of a catchment (or nexuses, but typically these will be points anyways)
- **Models**, which provide a computational backend
- **Formulations**, which, as I understand it, is a paired _instance_ of a realization and model
- **Simulations**, which are the temporal domain that a given formulation runs in.

> Again, this is my interpretation of them, so how I'm defining them might not be fully correct

I think what I should clarify from my comment is that (I think) we *should* be considering a formulation separate from a simulation (conceptually at least, though I think this might required knowing implementation details of the model). I think currently a formulation is the top-level abstraction that handles managing a realization + model + simulation -- I'm suggesting that a simulation manages formulations instead.

Separately, I think this also concerns discussion about how formulations and realizations are coupled, since it seems like we've modeled formulations as *being* realizations, rather than composing realizations into a formulation. There's probably an argument for that case though, since for example a catchment area realization "could" be modeled as a formulation via an "area" model, but I don't know if it's something we should consider generally. It might just be semantics at that point though... 🤷🏾 haha

Despite my comments, I think for our current needs the implementation of this PR is reasonable though. :smile:

> Also I just realized I say "I think" a lot... haha

PhilMiller:
@program-- I'd like to agree with you in principle about the delineation of responsibilities between Formulation and Simulation, but in practice, that's never going to happen. Every model we're going to work with is going to have its own internal representation of its time-dependent state.

In an ideal world, maybe the ngen framework engine would own all of the state variables, and pass the previous time's values to purely functional model evaluators to get back the next time's values, but that's just not a realistic view of how these things gets designed, nor how BMI is conceived to work.

program--:
@PhilMiller I think you misunderstand my suggestion (or, more likely, I didn't clarify it correctly 😛), but this idea doesn't concern BMI or the underlying models really, apart from acknowledging their state.

I'm suggesting **inverting** the class relationship/composition between `Simulation_Time` and the `Formulation`s. I think, in our case, `Simulation_Time` is composed within `Formulation_Manager`, so specifically I am saying that some `simulation` object would take ownership of (1) a temporal domain (i.e. X date to Y date), and (2) would derive a stride from the formulation with the most granular model (i.e. which model has the smallest non-negative `get_time_step()` value) -- the caveat here is that obviously this can get muddled pretty quickly depending on the units of time a model is stepping (but UDUNITS can aid in that).

> Thinking about this a bit more, I think this is maybe kind of what `Formulation_Manager` was intended to do, but it does much more than that.

This then gives the relationship that a `simulation` handles orchestrating time steps from the given domain to the models within its `Formulation`s. This also means that `simulation` does not care how the underlying models are implemented, just that a time step could be passed to a `Formulation`, and then value(s) are returned.

> In other words, given the domain and a simulation with *N* formulations, we can ask the simulation:
>
> > what's the output of all *N* formulations at `2023-08-25 01:00:00`, `step 6`, etc.?
>
> and the simulation would handle passing that step to the formulations.
>
> For another example similar to what I'm describing, see MATLAB's timetable synchronization function: https://www.mathworks.com/help/matlab/ref/timetable.synchronize.html

In the case of `Formulation`, it's basically another layer of abstraction over BMI, that is stateful in knowing the last time step a model computed at, and for some reason also handles outputting to disk/strings. I'm sure there's a way to design a stateless interface over the underlying stateful models. To be exact, I'm not suggesting that `ngen` would be able to take ownership of the models' timing implementations, just that we work on top of them to synchronize how a model steps versus the domain we are given.

I'm not stating that these constraints are necessarily exact or correct, but I'm trying to highlight that maybe the relationship between our temporal domain/time stepping/etc and what we consider a formulation could be represented better.

mattw-nws:
> In an ideal world, maybe the ngen framework engine would own all of the state variables, and pass the previous time's values to purely functional model evaluators to get back the next time's values, but that's just not a realistic view of how these things gets designed, nor how BMI is conceived to work.

🤔 Interesting concept... stateless functional models... that would probably be super-efficient.... you could program the models onto a FPGA accelerator! (Okay now I'm getting totally ahead of myself.) You could _kind of_ do that with BMI but it would probably be decidedly _not_ efficient... the model could make all of its state variables accessible and expect them to get passed back and forth (set_value and get_value) for every timestep or something...

mattw-nws:
> I'm suggesting **inverting** the class relationship/composition between `Simulation_Time` and the `Formulation`s. I think, in our case, `Simulation_Time` is composed within `Formulation_Manager`, so specifically I am saying that some `simulation` object would take ownership of (1) ... and (2) ...

Have you had a look at the [Layer object in the Multi-Layer PR?](https://github.com/NOAA-OWP/ngen/pull/615/files#diff-f9b4969217a3cfd61c03320c4bff10fbe92fddc925bb5c1fd0ef17b7f204422d) This could be a step in the direction you're suggesting. I haven't internalized it all myself and I think the `Formulation_Manager` still gets a `Simulation_Time` object representative of the conglomerate/composite simulation, but the layer gets composed with a Simulation_Time representing what's going on in that layer and takes over incrementing time in the set of formulations in that layer.

I think `Formulation_Manager` is a little poorly named too, it's more like "Simulation_Manager" (for the whole composite simulation run). Overall I'm not sure what the right answer is, and as I mentioned I'm still fuzzy on the abstract distinction between realizations and [other concept]. But Formulations originally weren't exclusively BMI, so there were BMI formulations which were a subclass of formulations and then non-BMI formulations, which shared a similar API (the reason for your comment about the formulation being another layer of abstraction over BMI)... there's also the Adapter set of classes, but those are a little more obvious, making non-C++ BMI formulations look like C++ BMI formulations.

program--:
> Have you had a look at the [Layer object in the Multi-Layer PR?](https://github.com/NOAA-OWP/ngen/pull/615/files#diff-f9b4969217a3cfd61c03320c4bff10fbe92fddc925bb5c1fd0ef17b7f204422d) This could be a step in the direction you're suggesting...

Yeah, just this morning I happened to take a look :smile: I agree that a `Layer` is closer to what I'm suggesting; thanks for pointing that out!

> I think Formulation_Manager is a little poorly named too, it's more like "Simulation_Manager" (for the whole composite simulation run)...

I also agree that `Formulation_Manager` may be poorly named -- rather I think we may just need to better specify its purpose/responsibilities, but that may be a separate discussion

mattw-nws:
Interestingly, I just noticed that these exist... which I just re-created in this PR, though at a different level in the hierarchy...

https://github.com/NOAA-OWP/ngen/blob/097d34b6ca60aac658d68a63bfe6a0f3f9cb106b/include/realizations/catchment/Bmi_Module_Formulation.hpp#L961-L964

Both are unused other than being initialized to zero... :-\

Issue #504, opened 2023-03-28
Odd forcing file format failure when using `SWDOWN`+`LWDOWN`


mattw-nws:
## Current behavior

Using one of the older-style forcing files, with a header like this:
```
Time,RAINRATE,T2D,Q2D,U2D,V2D,PSFC,SWDOWN,LWDOWN
2015-12-01 00:00:00,0,285.8000183,0.0093,-2.299999952,0.100000001,100310,0,361.3000183
```
blows up, with some kind of time indexing issue...The error is
```
terminate called after throwing an instance of 'std::out_of_range'
  what():  Forcing had bad init_time 0 for value request
```
It seems forcing specific. I get a similar error if I use a NetCDF forcing built from the same file.

This is the weird part: If I change that header to either of these forms, everything works:
```
Time,RAINRATE,T2D,Q2D,U2D,V2D,PSFC,DSWRF_surface,LWDOWN
2015-12-01 00:00:00,0,285.8000183,0.0093,-2.299999952,0.100000001,100310,0,361.3000183
```
or
```
Time,RAINRATE,T2D,Q2D,U2D,V2D,PSFC,SWDOWN,DLWRF_surface
2015-12-01 00:00:00,0,285.8000183,0.0093,-2.299999952,0.100000001,100310,0,361.3000183
```
one or the other of SW rad or LW rad has to use the AORC-style names... but it doesn't matter which!

This seems like something that could be related to the builtin `et_calc` but I could not see any evidence that this was being used... the formulation being used was tested with the PET module and with Noah's `ETRAN` output being fed to CFE's ET input.

Notably, removing CFE from the formulation made the problem go away... so that is related, but CFE itself is not likely to be the cause.

## Expected behavior

These headers should work for both CSV and NetCDF format.

## Steps to replicate behavior (include URLs)

1. Use the data/gauge_01073000 dataset in PR #502 , and change the headers to the examples above.
2. Create a NetCDF version of the forcing with those headers using the Python script in /utilities and the same behavior occurs.

mattw-nws:
@stcui007 Since #599 and other changes, this may have changed. Can you try to replicate this bug, and if it still happens, try to see what might be causing it?

stcui007:
I'll look into it.

PR #624, opened 2023-08-31
Fix using the forcing file header error.

stcui007:
The details are described in Issue #504. Briefly, when a forcing file with the following format:
```Time,RAINRATE,T2D,Q2D,U2D,V2D,PSFC,SWDOWN,LWDOWN```
```2015-12-01 00:00:00,0,285.8000183,0.0093,-2.299999952,0.100000001,100310,0,361.3000183```
is used, ngen throw a runtime " out of range" error.
The usually used header is as follows.
```time,APCP_surface,DLWRF_surface,DSWRF_surface,PRES_surface,SPFH_2maboveground,TMP_2maboveground,UGRD_10maboveground,VGRD_10maboveground,precip_rate```
```2015-12-01 00:00:00,0.0,360.8000183105469,0.0,100220.0,0.009399999864399433,285.8999938964844,-2.299999952316284,0.10000000149011612,0.0```

## Additions

-

## Removals

-

## Changes

include/realizations/catchment/Bmi_Multi_Formulation.hpp

## Testing

Run ngen test

## Notes
It is found the error was caused by setting ```init_time``` to ```0``` in ```CsvPerFeatureForcingProvider.hpp```. This was ultimately traced to ```get_data_start_time()``` which call ```get_variable_time_begin(const std::string &variable_name)``` in ```Bmi_Multi_Formulation.hpp```. This function uses ```availableData``` to extract the start time. If ```var_name``` in ```availableData``` is not in ```forcing``` data, then it returns ```0``` as start time. The solution makes sure if a ```var_name``` chosen from ```availableData``` is not a forcing variable name, we use the first forcing variable name as ```var_name``` to choose start time, which is always a valid time. We could actually use any of the forcing variable names to choose the start time, here we just uses the first one for convenience.

Also note that the fix works for both CSV and netCDF data.

## Todos

-

## Checklist

- [ x] Code can be automatically merged (no conflicts)
- [ x] Code follows project standards (link if applicable)
- [x ] Passes all existing automated tests
- [x ] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

1.

- [x ] Linux

### DIFF
--- a/include/realizations/catchment/Bmi_Multi_Formulation.hpp
+++ b/include/realizations/catchment/Bmi_Multi_Formulation.hpp
@@ -245,6 +245,10 @@ namespace realization {
                 for(std::map<std::string,std::shared_ptr<data_access::GenericDataProvider>>::iterator iter = availableData.begin(); iter != availableData.end(); ++iter)
                 {
                     var_name = iter->first;
+                    auto it = std::find(forcing->get_available_variable_names().begin(), forcing->get_available_variable_names().end(), var_name);
+                    if (it == forcing->get_available_variable_names().end()) {
+                        var_name = forcing->get_available_variable_names()[0];
+                    }
                     //TODO: Find a probably more performant way than trial and exception here.
                     try {
                         time_t rv = availableData[var_name]->get_data_start_time();
--- a/include/realizations/catchment/Bmi_Multi_Formulation.hpp
+++ b/include/realizations/catchment/Bmi_Multi_Formulation.hpp
@@ -239,26 +239,10 @@ namespace realization {
 
         time_t get_variable_time_begin(const std::string &variable_name) {
             std::string var_name = variable_name;
+            // when unspecified, assume all data is available for the same range.
+            // If no var_name, use forcing ...
             if(var_name == "*" || var_name == ""){
-                // when unspecified, assume all data is available for the same range.
-                // Find one that successfully returns...
-                for(std::map<std::string,std::shared_ptr<data_access::GenericDataProvider>>::iterator iter = availableData.begin(); iter != availableData.end(); ++iter)
-                {
-                    var_name = iter->first;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by PhilMiller** [comment]
Please fill in a description of the problem and its resolution in the PR notes, and correct the checklist formatting.

File: include/realizations/catchment/Bmi_Multi_Formulation.hpp [hunk 1]
```diff
@@ -245,6 +245,10 @@ namespace realization {
                 for(std::map<std::string,std::shared_ptr<data_access::GenericDataProvider>>::iterator iter = availableData.begin(); iter != availableData.end(); ++iter)
                 {
                     var_name = iter->first;
+                    auto it = std::find(forcing->get_available_variable_names().begin(), forcing->get_available_variable_names().end(), var_name);
+                    if (it == forcing->get_available_variable_names().end()) {
+                        var_name = forcing->get_available_variable_names()[0];
```
mattw-nws, line 250: The whole loop now boils down to this line. Just eliminate the `for` loop and set `var_name` to this... even the `try` block below is probably not useful now, because (I'm pretty sure) if `availableData[var_name]->get_data_start_time();` fails on a forcing var_name, something has gone wrong (in the current code, at least).

Actually... I think the whole use of a variable name is redundant in this code path (where `var_name == "*" || var_name = ""`... I think this whole section can be reduced to:

```
// Find one that successfully returns...
if(var_name == "*" || var_name == ""){
    return forcing->get_data_start_time();
}
// If not found ...
```

Can you try that?

mattw-nws: Actually the comment above that `if` is now not really correct either, probably just make it `// If no var_name, use forcing`... if the above works.

stcui007: Just tried. It seems to work.

mattw-nws (approved):

PR #625, opened 2023-09-01
GeoPackage `divide_id` attribute hotfix

mattw-nws:
HF pre-release files actually use a different schema and different ID attribute for divides, this PR makes that version work properly for GPKG files. The new format in geojson representation is _not_ fixed. This may break older GPKG formats, but tries to leave them working--no guarantees. Going forward GPKG format is only intended to support the latest HF formats.

## Additions

- Can now pass an `id_col` when reading features table from GPKG, some probing tries to guess which to pass for the divides layer.

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

- Probably remove support for the never-released format, do something with geojson support (update for new format/both formats, or deprecate?)

## Checklist

1.

- [X] Linux


src/geopackage/read.cpp
```diff
@@ -118,8 +135,12 @@ std::shared_ptr<geojson::FeatureCollection> geopackage::read(
         max_y = bbox[3] > max_y ? bbox[3] : max_y;
     }
 
-    return std::make_shared<geojson::FeatureCollection>(
+    auto fc = std::make_shared<geojson::FeatureCollection>(
         std::move(features),
         std::vector<double>({min_x, min_y, max_x, max_y})
     );
+
+    fc->update_ids();
```
PhilMiller, line 143: This is the only line here I'm curious about, since it's a deeper change in behavior.

That function doesn't have any sort of docstring, and I can't immediately intuit what the cross-reference data structure it's inserting into is for. I see this was added in a commit describing it as a bug fix, but there's not enough detail for me to understand it. Maybe an occasion to document it?

PhilMiller:
Code otherwise looks reasonable to me.

src/geopackage/read.cpp
mattw-nws: Oooh... I will have to take a look and/or ask @program-- to comment... he gave me that part of the PR in a patch. I believe it's something just missed from the geojson analogue though.

program--: I can't say exactly the purpose, but it seems like the `feature_by_id` member was used as a way to try and keep O(1) access to features and also keep them in a vector? I'm not sure _why_, but the reason this was needed in this PR was that something changed that caused a different `geojson::FeatureCollection` constructor be called, and not all of the constructors place the given features into the `feature_by_id` map.

PhilMiller: Ok, I'll take that as given for now. Maybe we'll rethink these data structures down the line, or at least be able to stop having to fiddle with how they're implemented so much.

src/geopackage/read.cpp
```diff
@@ -44,6 +44,22 @@ std::shared_ptr<geojson::FeatureCollection> geopackage::read(
         throw std::runtime_error(errmsg);
     }
 
+    // Introspect if the layer is divides to ses which ID field is in use
```
hellkite500, line 47:
```suggestion
    // Introspect if the layer is divides to see which ID field is in use
```

hellkite500 (dismissed):

mattw-nws:
I finally validated that the updated sample dataset included here does work with routing, so this is ready to go. I noticed that I omitted some deletions however. @PhilMiller or @hellkite500  can you re-approve? Thanks.

PhilMiller (approved):

PR #561, opened 2023-07-05
Shift UDUNITS to solely external installation

PhilMiller:
UDUNITS is available in every package repository we care about. If users want a specific version, they can still build it themselves and point CMake at it. Otherwise, there's no good reason for us to treat it specially by bundling code to download and build it.

## Changes

- Document that udunits should be installed from a package repository
- Remove code to download and install it ourselves
- Update the CI scripts to install it before configuring and building ngen

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows project standards (link if applicable)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

PR #631, opened 2023-09-06
Initial CMake Refactoring

### LINKED ISSUE #565: Two different FindNetCDF.cmake
We have two different CMake modules for finding the NetCDF4 library
* `cmake/FindNetCDF.cmake`, identical to https://github.com/Kitware/VTK/blob/master/CMake/FindNetCDF.cmake
* `cmake/modules/FindNetCDF.cmake`, maybe derived from https://github.com/jedbrown/cmake-modules/blob/master/FindNetCDF.cmake

At best, only one of them ever gets used in any way, and the other is just dead code. At worst, like in #362, they pick up different installations and lead to undefined behavior in the built code.

1.

### LINKED ISSUE #581: test/CMakeLists.txt defines a function `add_test` that it not the CMake-provided `add_test`
The very generically-named function `add_test` [we define](https://github.com/NOAA-OWP/ngen/blob/master/test/CMakeLists.txt#L44) collides with the common function of the same name [provided by CMake](https://cmake.org/cmake/help/latest/command/add_test.html)

At the very least, we should rename our local thing.

### LINKED ISSUE #585: CMake Policy CMP0074 set after some `find_package` calls
https://cmake.org/cmake/help/latest/policy/CMP0074.html
https://github.com/NOAA-OWP/ngen/blob/b88ed69a5f2514b6bf02ddd6155308e7057f457c/CMakeLists.txt#L177

Comes in the middle of our `CMakeLists.txt`, after some calls to `find_package` but before others. Should we move that up before all of them, so that we get consistent behavior?

## Current behavior

Some calls to `find_package` will not consider `PackageName_ROOT` variables out of concern that they may not mean what's expected, while others will once the policy is set.

## Expected behavior

All calls to `find_package` behave consistently.

### LINKED ISSUE #629: Initial CMake handling refactor
After a brief conversation, a few issues with CMake throughout the project have been identified:

1. We should bump the CMake minimum version to 3.17. The primary benefit of this is getting better FindPython support, along with better defaulted CMake policies.
2. `cmake_minimum_required` is called throughout subdirectory CMakeLists. We should remove these to ensure that CMake policies aren't being changed when adding a subdirectory to the root CMakeLists.
3. Source files for some sublibraries are being globbed using a system command, see: https://github.com/NOAA-OWP/ngen/blob/master/cmake/dynamic_sourced_library.cmake. In this case, best practice is to manually list out source files, but a good middle ground is to use `file(GLOB_RECURSE ...)` with the `CONFIGURE_DEPENDS` flag. See [here](https://cmake.org/cmake/help/v3.17/command/file.html#filesystem) for more details on the flag.
4. Some external dependencies (NetCDF, iso_c_fortran_bmi) are added to the project, but do not have targets associated to them. Ideally, we should create `IMPORTED` targets encapsulating these, in accordance with best practices.
5. CMake configure options are not consistent (i.e. `MPI_ACTIVE` vs `NGEN_ACTIVATE_PYTHON`). I propose refactoring the names to prefix with `NGEN_WITH_*`, such that the given example options would be named `NGEN_WITH_MPI` and `NGEN_WITH_PYTHON` -- or `NGEN_ACTIVATE_*` since some options are already formatted this way. This is a lower priority, but can be done without affecting the preprocessor definitions in use.
6. #585
7. #581 - Probably should handle shadowing of the builtin `add_test` function. Maybe just rename to `ngen_add_test`.
8. #565 - This can be dealt with within (4).

program--:
This PR resolves #629, resolves #585, resolves #581, resolves #586, and resolves #565.

## Changes

- Reorganizes the root CMakeLists.txt into clearly defined sections, in the order of:
  1. Policies
  2. Project Variables
  3. Options
  4. Dependencies
  5. Project Targets
  6. Build Summary
- Flattens the `cmake/` directory and sets it as the module dir, since there is no need for `cmake/modules`.
- Updates `ngen` CMake configure options to have consistent naming:
  - `NETCDF_ACTIVE -> NGEN_WITH_NETCDF`
  - `MPI_ACTIVE -> NGEN_WITH_MPI`
  - `UDUNITS_ACTIVE -> NGEN_WITH_UDUNITS`
  - ... etc. See the third section of the root CMakeLists for the rest
  > **Note**
  > This does not currently affect preprocessor definitions.
  
  > **Note**
  > Previous configure options/env vars will be supported, but will
  > notify at configure-time that they are deprecated. Users should switch
  > to using new option names.

- Removes all calls to `cmake_minimum_required` outside of the root CMakeLists
- Adds imported target for NetCDF
- Adds imported target for iso_c_fortran_bmi
- Update ngen-defined `add_test` to `ngen_add_test` and use CMake argument parsing rather than explicit numbering to define test sources/linking/construction dependencies.
- Fixes target linking, such that new targets should only link to direct dependencies, and not transitive dependencies. (i.e. test targets linking directly to NetCDF when NGen components expose NetCDF linking)
- Updates `dynamic_sourced_cxx_library` to use `file(GLOB_RECURSE ... CONFIGURE_DEPENDS ...)` instead of calling a system command.
- Adds `deprecated_option` function, which allows for compatibility between new and old options.

## Notes

I removed the `FindNetCDF.cmake` module in `cmake/` rather than the one in `cmake/modules/` since it was more up to date (or at least had more recent commits...) -- however, I should note the one in the root `cmake/` directory created NetCDF targets. I'm not 100% on the reason for both, but I'm assuming here the root one was not working correctly.

## TODO

- [x] Create imported target for `iso_c_fortran_bmi` library when `NGEN_WITH_FORTRAN` is enabled.
- [ ] Update relevant documentation.

  > **Important**
  > Maybe this should be updated in the github wiki?
  > We are a bit fragmented as far as documentation goes.

## Checklist


src/core/CMakeLists.txt
```diff
@@ -60,6 +58,10 @@ else ()
             )
 endif ()
 
+if(NGEN_WITH_MPI)
+    target_link_libraries(core PUBLIC MPI::MPI_C MPI::MPI_CXX)
```
PhilMiller, line 62: Does this need a nested check for Fortran as well, to pull in the MPI Fortran bindings?

cmake/FindUDUNITS2.cmake
```diff
@@ -0,0 +1,27 @@
...
+mark_as_advanced(UDUNITS2_INCLUDE)
+
+find_library(UDUNITS2_LIBRARY
+    NAMES
+        libudunits2.so
+        libudunits2.dylib
+)
+mark_as_advanced(UDUNITS2_LIBRARY)
+
+if(DEFINED UDUNITS2_INCLUDE)
```
PhilMiller, line 15: Should this be 
```suggestion
if(DEFINED UDUNITS2_INCLUDE AND DEFINED UDUNITS2_LIBRARY)
```

CMakeLists.txt
```diff
@@ -1,332 +1,387 @@
...
+
+# Options =====================================================================
+include(CMakeDependentOption)
+
+option(NGEN_WITH_MPI         "Build with MPI support"             OFF)
+option(NGEN_WITH_NETCDF      "Build with NetCDF support"          ON)
+option(NGEN_WITH_SQLITE      "Build with SQLite3 support"         OFF)
+option(NGEN_WITH_UDUNITS     "Build with UDUNITS2 support"        ON)
+option(NGEN_WITH_BMI_FORTRAN "Build with Fortran BMI support"     OFF)
+option(NGEN_WITH_BMI_C       "Build with C BMI support"           OFF)
```
PhilMiller, line 44:
```suggestion
option(NGEN_WITH_BMI_C       "Build with C BMI support"           ON)
```
There's no practical reason I can see for not always building the C library support. No extra compiler needed, no huge added code, etc.

CMakeLists.txt
```diff
@@ -1,332 +1,387 @@
...
-if ("${CMAKE_SYSTEM_NAME}" STREQUAL "Darwin")
+# FIXME: What is this used for? - Justin
+if("${CMAKE_SYSTEM_NAME}" STREQUAL "Darwin")
     set(SHARED_LIB_EXTENSION "dylib")
 elseif(WIN32)
-    error( "Windows platforms are not currently supported" )
+    error("Windows platforms are not currently supported")
 else()
     set(SHARED_LIB_EXTENSION "so")
 endif()
```
PhilMiller, line 115: This entire block is unused now - it was only used for the bundled UDUNITS2 build, which was eliminated in #561.

It maybe should inform the code in https://github.com/NOAA-OWP/ngen/blob/master/include/realizations/catchment/AbstractCLibBmiAdapter.hpp#L80 but it doesn't currently.

CMakeLists.txt
```diff
@@ -1,332 +1,387 @@
...
 if(UDUNITS_QUIET)
-add_compile_definitions(UDUNITS_QUIET)
+    add_compile_definitions(UDUNITS_QUIET)
 endif()
 
+# FIXME: I don't think this is needed now? - Justin
 if(ET_QUIET)
-add_compile_definitions(ET_QUIET)
+    add_compile_definitions(ET_QUIET)
 endif()
```
PhilMiller, line 231: The internal ET code has been deleted, so yeah, this can be cleaned out.

cmake/DeprecatedOptions.cmake
```diff
@@ -0,0 +1,28 @@
...
+
+Adds a CMake option that will output a deprecation message on use.
+
+The `NEW` variable is the option that should be used when building.
+
+The `OLD` variable is the deprecated option that corresponds to a new option.
+
+The `INCLUDE_ENV` flag also checks the environment variables for a variable named `OLD`, i.e. ENV{<OLD variable>}.
+#]==]
+macro(deprecated_option)
```
program--, line 12: This should be a function, not a macro

CMakeLists.txt
```diff
@@ -1,332 +1,387 @@
...
-endif()
+# -----------------------------------------------------------------------------
 
-add_subdirectory("src/core")
-add_dependencies(core libudunits2)
-add_subdirectory("src/geojson")
+# Natively support BMI C++ modules and pre-compile in the test_bmi_cpp mock/example.
+set(TEST_BMI_CPP_DIR ${NGEN_EXT_DIR}/test_bmi_cpp)
+git_update_submodule(${NGEN_EXT_DIR}/bmi-cxx)
+add_subdirectory(${TEST_BMI_CPP_DIR} ${TEST_BMI_CPP_DIR}/cmake_build) # FIXME: Why is this here? - Justin
```
PhilMiller, line 238: Does the test get built if this isn't here? I would think not

program--: Right, I think I put this comment because I intended to move into the `test/CMakeLists.txt`, since it doesn't make sense to build it unless tests are being built -- if that makes sense

program--: If tests are enabled, it may make sense to have flags (that are dependent on tests being enabled) for building the various test BMI models in this same fashion. i.e. if we want to build the fortran test bmi model when configuring for tests, then we could pass `-DNGEN_BUILD_FORTRAN_TEST_BMI` or something similar

PhilMiller: If it's easy to square away, go ahead and do it in this PR. If it would take more than a few minutes of additional effort to figure out, let's just leave it be for now.

program--: Good call, I think for the sake of this PR, I'll move this code block into the test CMakeLists as is, and then handle those extra options in a future PR.

src/core/CMakeLists.txt
program--: This linking is for `HY_Features_MPI`, so I don't think so

CMakeLists.txt
program--: I've changed the name of the variable to `NGEN_SHARED_LIB_EXTENSION` and added it as a compile definition. No changes to the dynamic loader, but it's setup for that now, which I think should be useful.

CMakeLists.txt
```diff
@@ -1,332 +1,374 @@
...
+    add_compile_definitions(NGEN_WITH_UDUNITS)
+
+    # Create UDUNITS2 target
+    # Note: GLOBAL is required here in order to extend the scope of the target.
+    #       see: https://stackoverflow.com/a/46491758/6891484
+    add_library(libudunits2 SHARED IMPORTED GLOBAL)
+    set_target_properties(libudunits2 PROPERTIES IMPORTED_LOCATION "${UDUNITS2_LIBRARY}")
+    target_include_directories(libudunits2 INTERFACE "${UDUNITS2_INCLUDE}")
+        
+    get_target_property(_SUMMARY_UDUNITS_INCLUDE libudunits2 INTERFACE_INCLUDE_DIRECTORIES)
```
PhilMiller, line 148: If there's any other changes to be made, it would be kinda nice if this were all encapsulated in `FindUDUNITS2.cmake`. It would also make that more share-able with other projects as a result.

CMakeLists.txt
```diff
@@ -1,332 +1,374 @@
...
+
+# Define project version and use via generated config header
 project(ngen VERSION 0.1.0)
-configure_file(include/NGenConfig.h.in include/NGenConfig.h)
 
-if("${PROJECT_NAME}" STREQUAL "${CMAKE_PROJECT_NAME}")
-    option(PACKAGE_TESTS "Build automated tests" ON)
+configure_file("${NGEN_INC_DIR}/NGenConfig.h.in" "${NGEN_INC_DIR}/NGenConfig.h")
+
+add_executable(ngen "${NGEN_SRC_DIR}/NGen.cpp")
```
PhilMiller, line 80: It would vaguely make sense to move this and everything that touches the target `ngen` down to `src/CMakeLists.txt`, but I don't see that as essential

CMakeLists.txt
program--: Yeah, I think part of this was initially in there, but I moved it out for some reason (couldn't tell you why haha). I see no problems with moving this and the NetCDF target creation into their respective modules. I'll make a short one for the iso_c_bmi_fortran target too, since then it can just be a `find_package` call.

test/CMakeLists.txt
```diff
@@ -1,300 +1,320 @@
...
-include_directories(${gtest_SOURCE_DIR}/include ${gtest_SOURCE_DIR})
-include_directories(${gmock_SOURCE_DIR}/include ${gmock_SOURCE_DIR})
-include_directories(${PROJ_ROOT_INCLUDE_DIR})
+add_compile_definitions(NGEN_BMI_CPP_LIB_TESTS_ACTIVE)
 
-if (NGEN_ACTIVATE_PYTHON)
-    include_directories(${PROJECT_SOURCE_DIR}/extern/pybind11/include)
+# Natively support BMI C++ modules and pre-compile in the test_bmi_cpp mock/example.
+set(TEST_BMI_CPP_DIR ${NGEN_EXT_DIR}/test_bmi_cpp)
+git_update_submodule(${NGEN_EXT_DIR}/bmi-cxx)
```
PhilMiller, line 6: Should `bmi_cxx` be referenced directly from the top-level CMakeLists.txt?

PhilMiller (dismissed):

PhilMiller:
I'm really happy with this. It's a big enough change that it's maybe worth having others look over it, too.

Regardless, everyone on the team needs to be aware of its impact

mattw-nws:
> I'm not 100% on the reason for both...
I think I added the 2nd one (the one kept?) not realizing that there was already one not in `modules`. Plain accident.

.github/actions/ngen-build/action.yaml
```diff
@@ -176,6 +180,7 @@ runs:
               -DBMI_FORTRAN_ACTIVE:BOOL=${{ inputs.bmi_fortran }} \
               -DNGEN_ACTIVATE_ROUTING:BOOL=${{ inputs.use_troute }} \
               -DNETCDF_ACTIVE:BOOL=${{ inputs.use_netcdf }} \
+              -DNGEN_WITH_SQLITE:BOOL=${{ inputs.use_sqlite }} \
               -DMPI_ACTIVE:BOOL=${{ inputs.use_mpi }} -S .
```
mattw-nws, line 184: Wouldn't it be in scope to update all of these? Any particular reason why not?

program--: The primary reason was to ensure that this PR was entirely backwards compatible -- and ensuring that the deprecated options work as expected (even though they will be removed eventually). This then gives us leeway in updating other portions affected by this PR (i.e. documentation, preprocessor defs, etc.)

CMakeLists.txt
```diff
@@ -1,332 +1,335 @@
...
+# find_package() uses <PackageName>_ROOT variables.
+# https://cmake.org/cmake/help/latest/policy/CMP0074.html
+cmake_policy(SET CMP0074 NEW)
 
-if(MPI_ACTIVE)
-   set (CMAKE_CXX_COMPILER "mpicxx")
+# find_package() uses upper-case <PACKAGENAME>_ROOT variables.
+# https://cmake.org/cmake/help/latest/policy/CMP0144.html
+if(POLICY CMP0144)
+    cmake_policy(SET CMP0144 NEW)
```
mattw-nws, line 19: For other reviewers... that is, it uses them _also_... So this makes either `NetCDF_ROOT` _or_ `NETCDF_ROOT` work.

CMakeLists.txt
```diff
@@ -1,332 +1,335 @@
...
 
+# -----------------------------------------------------------------------------
+# If MPI support is enabled, set the C++ compiler to "mpicxx"
+if(NGEN_WITH_MPI)
+    find_package(MPI REQUIRED)
+    add_compile_definitions(NGEN_MPI_ACTIVE)
+endif()
+
+# -----------------------------------------------------------------------------
+# FIXME: Is this needed for icpx? or even icpp? This only enables support for SYCL - Justin
```
mattw-nws, line 100: I added this when trying to get the build working on Hera... I don't remember exactly the symptoms, but I was having a hard time getting everything to use icpx, and adding this fixed it. This was a scenario where you had to load up the GCC9 module to get its stdlib, and then the OneAPI module to get icpx... but certain things (gfortran/mpif90?) kept going back to using GCC toolchain, and when I added this that problem stopped happening.... if I remember correctly (that was close to a year ago). It could be that something else is the "right" solution and this fixed it by accident, but I'm not sure what the right answer is then.

program--: Hmm maybe the IntelDPCPPConfig.cmake file is different on Hera -- the one I have doesn't modify anything besides SYCL variables. I also have to explicitly set `INTEL_DPCPP` for this branch to execute, even with `IntelLLVM` as my toolchain... so maybe it's an environment requirement. In any case, I don't think there's an issue with keeping this section of code in, so I'll modify the comment to provide context and leave it as is 😄

mattw-nws: I'll add (for posterity) that there is probably a smarter way of figuring out whether it ought to be turned on (i.e. turn on if CMAKE_CXX_COMPILER is icpx (which is non-trivial if it's `mpicxx`) or something) but I just kludged in a -D solution.

test/CMakeLists.txt
mattw-nws: The thinking here was actually that we should stop using `include/bmi.hpp`--I think it is identical--and point to the upstream version...sort of as a matter of principle. Dogfooding, in a certain way. We are adamant that formulations must use an identical header to "ours" (for fairly obvious reasons of compatability)... what we really mean is that conformance to the official BMI header is required, both for semantic and technical reasons. So, do what we say and use a copy that can't be modified (practically)... and encourage the formulations to do the same, in fact.

mattw-nws (dismissed):
Consideration for follow-on might be to go ahead and `add_subdirectory` (plus build directory) for `iso_c_fortran_bmi`, similar to what is done for `test_bmi_cpp`, since the former is essentially a requirement when `NGEN_WITH_BMI_FORTRAN` is ON--this would guarantee that it gets built. Right now if the user enables fortran they're required to manually build iso_c_fortran_bmi.

CMakeLists.txt
PhilMiller: "If CMAKE_CXX_COMPILER is icpx" would be a direct test of `if (CMAKE_CXX_COMPILER_ID STREQ IntelLLVM)`

This block is sketchy, but not really a big deal right now.

For context, the Intel compilers (and Nvidia, not that it matters for us) don't ship their own implementation of the standard library or the compiler-ABI-defining headers. Rather, they defer to a reference version of GCC either based on `PATH` or a command-line argument setting the reference explicitly to get their `libstdc++`. That's why they need the GCC9 module (for instance) loaded - it has to be a version that supports the language standard icpx is compiling for, but crucially, not one whose headers rely on language version support beyond what icpc implements. It's a really delicate dance, with poorly documented compatibility.

PhilMiller:
I was seeing a local build failure with this branch that doesn't occur on `master`:

```
cd \
/Users/phil/Code/noaa/ngen/build_bmi_conformance/test \
&& \
/Library/Developer/CommandLineTools/usr/bin/c++ \
-DNETCDF_ACTIVE \
-DNGEN_BMI_CPP_LIB_TESTS_ACTIVE \
-DNGEN_BMI_C_LIB_ACTIVE \
-DNGEN_BMI_C_LIB_TESTS_ACTIVE \
-DNGEN_SHARED_LIB_EXTENSION \
-DNGEN_WITH_SQLITE3 \
-DNGEN_WITH_UDUNITS \
-I/Users/phil/Code/noaa/ngen/include \
-I/Users/phil/Code/noaa/ngen/include/simulation_time \
-I/Users/phil/Code/noaa/ngen/include/core \
-I/Users/phil/Code/noaa/ngen/include/core/catchment \
-I/Users/phil/Code/noaa/ngen/include/core/nexus \
-I/Users/phil/Code/noaa/ngen/include/core/hydrolocation \
-I/Users/phil/Code/noaa/ngen/include/core/mediator \
-I/Users/phil/Code/noaa/ngen/include/realizations/catchment \
-I/Users/phil/Code/noaa/ngen/include/geojson \
-I/Users/phil/Code/noaa/ngen/include/forcing \
-I/Users/phil/Code/noaa/ngen/include/utilities \
-I/Users/phil/Code/noaa/ngen/extern/pybind11/include \
-isystem \
/opt/homebrew/include \
-isystem \
/Users/phil/Code/noaa/ngen/test/googletest/googletest/include \
-isystem \
/Users/phil/Code/noaa/ngen/test/googletest/googletest \
-std=gnu++14 \
-arch \
arm64 \
-isysroot \
/Library/Developer/CommandLineTools/SDKs/MacOSX13.1.sdk \
-mmacosx-version-min=12.6 \
-fvisibility=hidden \
-MD \
-MT \
test/CMakeFiles/test_partition.dir/utils/Partition_Test.cpp.o \
-MF \
CMakeFiles/test_partition.dir/utils/Partition_Test.cpp.o.d \
-o \
CMakeFiles/test_partition.dir/utils/Partition_Test.cpp.o \
-c \
/Users/phil/Code/noaa/ngen/test/utils/Partition_Test.cpp


/Users/phil/Code/noaa/ngen/test/utils/Partition_Test.cpp:2:10: fatal error: 'gmock/gmock.h' file not found
#include "gmock/gmock.h"
         ^~~~~~~~~~~~~~~
1 error generated.
```

I just pushed a (rebase and) fix

PhilMiller (approved):

PhilMiller (changes requested):
Matt and I are seeing various issues with this on the UCS systems. We'll try to get that worked out today, but merging should be held until then.

cmake/FindNetCDF.cmake
```diff
@@ -1,131 +1,145 @@
...
-    add_library(NetCDF::NetCDF UNKNOWN IMPORTED)
-    set_target_properties(NetCDF::NetCDF PROPERTIES
-      IMPORTED_LOCATION "${NetCDF_LIBRARY}"
-      INTERFACE_INCLUDE_DIRECTORIES "${NetCDF_INCLUDE_DIR}")
-  endif ()
-endif ()
+if(NETCDF_FOUND)
+  # Create NetCDF target
+  add_library(NetCDF SHARED IMPORTED GLOBAL)
+  target_include_directories(NetCDF INTERFACE "${NETCDF_INCLUDE_DIR}" "${NETCDF_CXX_INCLUDE_DIR}")
```
PhilMiller, line 140:
```suggestion
  target_include_directories(NetCDF INTERFACE "${NETCDF_INCLUDE_DIR}" "${NETCDF_CXX_INCLUDE_DIR}")
  target_link_libraries(NetCDF INTERFACE "${NETCDF_LIBRARY}" "${NETCDF_CXX_LIBRARY}")
```

PhilMiller: This seems to address the NetCDF linking issue I was seeing on the UCS system

program--: interesting... It seems like it should be redundant with the lines after it, but if it resolved that then 🤷🏾

PhilMiller: I was seeing it link against one of the libraries, but not both.

PhilMiller (approved):

PR #643, opened 2023-09-19
Implement a logging function

stcui007:
This PR creates an initial logging solution that mimics Python's logging functionality.

## Additions

include/utilities/logging_utils.h
This header file contains the following logging functions as described in Issue#574:
void debug(char *)
void info(char *)
void warning(char *)
void error(char *)
void critical(char *)
that handle different types of output.

Also added a unit test code:
test/utils/logging_Test.cpp

## Removals

-

## Changes

As an illustration, ```Bmi_Module_Formulation.hpp``` is modified with some codes to use ```logging_utils.h``` to output various messages

A few CMakelist.txt files related to the logging functionalities and unit test have minor changes.

## Testing

Run ngen test
Unit tests

## Notes

-

## Todos

-

## Checklist

- [x ] Passes all existing automated tests
- [x ] New functions are documented (with a description, list of inputs, and expected output)
- [x ] Placeholder code is flagged / future todos are captured in comments
- [x ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [x ] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

1.

- [x ] Linux

### DIFF
--- a/include/utilities/logging_utils.h
+++ b/include/utilities/logging_utils.h
@@ -1,0 +1,96 @@
+#ifndef NGEN_LOGGING_UTILS_H
+#define NGEN_LOGGING_UTILS_H
+
+#include <iostream>
+#include <cstring>
+#include <string>
+
+#define MAX_STRING_SIZE 2048
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+namespace logging {
+    /**
+     * Send debug output to std::cerr
+     * @param msg The variable carries the humanly readable debug text info.
+     */
+    inline void debug(char* msg)
+    {
+        #ifndef NGEN_QUIET
+            std::string msg_str = std::string(msg);
+            if (msg_str.size() > MAX_STRING_SIZE) {
+                std::cerr << "DEBUG: the message string exceeds the maximum length limit" << std::endl;
+            }
+
+            std::cerr<<"DEBUG: " << msg_str << std::endl;
+        #endif
+    }
+
+    /**
+     * Send info content to std::cerr
+     * @param msg The variable carries the humanly readable info text.
+     */
+    inline void info(char* msg)
+    {
+        #ifndef NGEN_QUIET
+            std::string msg_str = std::string(msg);
+            if (msg_str.size() > MAX_STRING_SIZE) {
+                std::cerr << "INFO: the message string exceeds the maximum length limit" << std::endl;
+            }
+
+            std::cerr<<"INFO: " <<std::string(msg) << std::endl;
+        #endif
+    }
+
+    /**
+     * Send warning to std::cerr
+     * @param msg The variable carries the humanly readable warning message.
+     */
+    inline void warning(char* msg)
+    {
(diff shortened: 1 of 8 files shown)


include/utilities/logging_utils.h
```diff
@@ -0,0 +1,96 @@
...
+namespace logging {
+    /**
+     * Send debug output to std::cerr
+     * @param msg The variable carries the humanly readable debug text info.
+     */
+    inline void debug(char* msg)
+    {
+        #ifndef NGEN_QUIET
+            std::string msg_str = std::string(msg);
+            if (msg_str.size() > MAX_STRING_SIZE) {
```
mattw-nws, line 23: The purpose of `MAX_STRING_SIZE` is not to prevent clients from giving us messages larger than we want to read. Rather, it is to give a bound to C implementations that may need to allocate specific buffer sizes, and perhaps to specify a point at which reading should be stopped if the string is not properly null-terminated. If this latter is the case, std::string::size is probably not going to work anyway.

For a C++ implementation as we have here, what probably makes the most sense is to ignore it. If we get a string that's not properly terminated, we probably can't know what to do with it anyway. Just pass the input to std::cerr... someone else might disagree, the most we could do is check the C-string length with something like `strnlen`, but this has the same problem as std::string::size. I would probably just ignore the `MAX_STRING_SIZE` in this implementation.

This simplifies these functions significantly, allowing you to get rid of multiple variables and copies of the string.

mattw-nws (changes requested):

stcui007:
These are all reasonable suggestions. I used "CRITICALl" here just for
illustration as only "CRITICAL" and "ERROR" are not "QUIET" so I can see
the output actually works. I did test all other functions though. Future
version will make appropriate changes.

include/utilities/logging_utils.h
```diff
@@ -0,0 +1,96 @@
...
+
+        std::cerr<<"CRITICAL: " <<std::string(msg) << std::endl;
+    }
+}
+
+#ifdef     __cplusplus
+}
+#endif
+
+#endif //NGEN_LOGGING_UTILS_H
```
mattw-nws, line 96: Additionally, really any new functionality added should have some unit tests. This is pretty simple so the unit test doesn't have to be huge, but something that puts a message through each function and makes sure it comes out with the expected text would be good. You can look at this approach from one of @program-- 's old commits for an example of how to do that:
https://github.com/program--/ngen/blob/4a1d137bff0ff3324cfb2239ac1eaec72b66705d/test/forcing/CsvPerFeatureForcingProvider_Test.cpp#L202-L216

mattw-nws (dismissed):

include/utilities/logging_utils.h
stcui007: Agree. this should be done.

hellkite500 (approved):

mattw-nws (commented):
[Resetting my review]

PR #670, opened 2023-11-06
Minor GPKG/SQLite Reorganization

### LINKED ISSUE #567: PR #522 Compiler Optimization Issue
This issue is meant to track an ongoing issue in PR #522 regarding optimizations in GCC and Clang on boost::geometry SRS transformation objects. Clang/LLVM is being used primarily to debug this issue.

## Current Behavior
When compiling `wkb.cpp` with `-O0`, compilation is successful.

However, when compiling with `-O1` or higher, compilation fails. Verbose output from clang shows that recursive calls to `llvm::ScalarEvolution::getRangeRef` are causing a stack overflow.

By increasing the process stack size (using `ulimit -s 15036`), we can successfully compile with `-O1` and higher. This, however, doesn't alleviate the issue of the recursive calls during optimization.

### Bisect Checking
To figure out which optimization is causing the overflow, we use the LLVM option `-opt-bisect-limit=N` to find both the *last* successful compilation, and the *first* failed compilation. In my case, this was passes 97 and 98. Outputting the results of these passes and using `diff` returned:

```diff
98c98
< BISECT: running pass (98) SROAPass on _ZN5boost8geometry3srs4dpar10parametersIdE3addINS2_10value_projEEERS4_T_
---
> BISECT: NOT running pass (98) SROAPass on _ZN5boost8geometry3srs4dpar10parametersIdE3addINS2_10value_projEEERS4_T_
```
This (I think) tells us the [SROA Optimization](https://llvm.org/doxygen/classllvm_1_1SROAPass.html) is causing the error.

> I believe this is the SROA optimization applied to `boost::geometry::srs::dpar::parameters::add(value_proj)`, which is used twice in the projection code for defining `epsg5070` and `epsg3857`, though I'm not sure.

### Behavior in GCC vs Clang
Currently, GCC does not segfault regardless of having a default stack size. However, compilation is _very_ slow (upwards of 30+ minutes). I'm guessing that the same or similar optimization is causing the issue, since outputting and inspecting the optimization call graph shows optimizing hanging up at the `boost::geometry::srs::dpar::parameters` explicit constructor, which calls `::add(value_proj)` in our case.

## System Information
I am using the following versions to debug this:
- gcc version 13.1.1 20230429
- clang version 15.0.7
- Intel(R) oneAPI DPC++/C++ Compiler 2023.1.0 (2023.1.0.20230320)

With Linux kernel: `6.4.2-arch1-1`

The LLVM IR output for this issue is attached here: [wkb.ll.tar.gz](https://github.com/NOAA-OWP/ngen/files/12032341/wkb.ll.tar.gz). I am running the following on it to verify optimization issues:

```
opt -O2 -S -verify-each
```

## Potentially Related Issues
```
https://github.com/llvm/llvm-project/issues/49579
```

program--:
This PR aims to solve a linkage error seen when enabling tests by simplifying the sqlite wrapper public interface and aims to reorganize the SQLite wrapper into a cleaner structure. Additionally, all GPKG and SQLite functionality is now enclosed inside the `ngen` namespace (i.e. `ngen::sqlite` and `ngen::geopackage`).

Previous error:
```
../src/geopackage/libgeopackage.a(read.cpp.o): In function `geopackage::read(std::string const&, std::string const&, std::vector<std::string, std::allocator<std::string> > const&)':
/home/shengting.cui/Xia_huc01/files_huc01/ngen/src/geopackage/read.cpp:98: undefined reference to `geopackage::sqlite_iter geopackage::sqlite::query<std::string>(std::string const&, std::string const&)'
collect2: error: ld returned 1 exit status
gmake[2]: *** [test/CMakeFiles/test_geopackage.dir/build.make:139: test/test_geopackage] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:2003: test/CMakeFiles/test_geopackage.dir/all] Error 2
```

This error most likely resulted from my accidental implementation of the templated `query` function inside a source file, rather than directly in the header, i.e.:
https://github.com/NOAA-OWP/ngen/blob/7d91513c94c4d93fe8183e31acd66c0ba477af7c/include/geopackage/NGen_SQLite.hpp#L224-L225

https://github.com/NOAA-OWP/ngen/blob/7d91513c94c4d93fe8183e31acd66c0ba477af7c/src/geopackage/sqlite/database.cpp#L88-L108

## Changes

- Namespaces
- `sqlite` -> `ngen::sqlite::database`
- `sqlite_iter` -> `ngen::sqlite::database::iterator`
- `sqlite_deleter` -> `ngen::sqlite::database::deleter`
- `database::query` now has only two overloads. One is `query(std::string, span<std::string>)`, with the second argument defaulting to an empty span. This is for code deduplication. The second overload is a templated function that takes a parameter pack of strings, and constructs the span in place. This overload is only enabled when all parameter pack arguments are convertible to `std::string`.
- Renamed header files to be more consistent with the source files. I opted to keep the `ngen::geopackage` implementations separate to help with incremental compilation, and to aid future debugging of the optimization issue #567.
- Separate projection header and functions into `ngen::srs::epsg` and new header/source `proj.{hpp/cpp}`.
- Resolve #567 by defining a static table of only the projections we need, rather than searching against all defined projections in boost.geometry.

## Testing

All unit tests passed locally using GCC 13.1.1, Clang 16.0.6, and IntelLLVM 2023.

## Checklist

- [X] Linux
- [X] macOS


PhilMiller:
Could you please note what the linkage error was in the PR description?

hellkite500 (approved):

PhilMiller:
Because diffing is not going to line things up nicely, were there any substantive changes in the implementation of `sqlite_iter/iterator` or `sqlite/database`? I can review the code de novo, but it would be helpful to give extra attention to bits that aren't just renamed versions of exactly the same functionality.

program--:
> Because diffing is not going to line things up nicely, were there any substantive changes in the implementation of `sqlite_iter/iterator` or `sqlite/database`? I can review the code de novo, but it would be helpful to give extra attention to bits that aren't just renamed versions of exactly the same functionality.

No, the only changes made were that I renamed member variables of the iterator and database classes, but the member function logic is the same. I did rename 2 member functions as well, `database::has_table` is `database::contains`, and `database::column_index` became `database::find`.

PhilMiller (approved):

program--:
Moving back to draft because of compilation issues... most likely #567

hellkite500 (approved):

PR #673, opened 2023-11-09
Sqlite null fix

hellkite500:
Fix handling of null/nan enteries in hydrofabric.

## Changes

- Allow sqlite iterator to read column types per iteratorion
- Check for string `"null"` in alt_ids when referencing feature by alternative id property

## Testing

1. Tested locally, and via local use of the partition generator to generate partitions of huc01 hydrofabric.

## Checklist

- [x] Linux
- [x] MacOS

### DIFF
--- a/src/geopackage/sqlite/iterator.cpp
+++ b/src/geopackage/sqlite/iterator.cpp
@@ -76,7 +76,13 @@ sqlite_iter& sqlite_iter::next()
         const int returncode = sqlite3_step(this->ptr());
         if (returncode == SQLITE_DONE) {
             this->iteration_finished = true;
-        } else if (returncode != SQLITE_ROW) {
+        }
+        else if( returncode == SQLITE_ROW){
+            //Update column dtypes for the next row
+            for (int i = 0; i < this->column_count && i < column_types.size(); i++) {
+                this->column_types[i] = (sqlite3_column_type(this->ptr(), i));
+            }
+        } else {
             throw sqlite_error("sqlite3_step", returncode);
         }
         this->iteration_step++;
--- a/src/geopackage/sqlite/iterator.cpp
+++ b/src/geopackage/sqlite/iterator.cpp
@@ -78,7 +78,18 @@ sqlite_iter& sqlite_iter::next()
             this->iteration_finished = true;
         }
         else if( returncode == SQLITE_ROW){
-            //Update column dtypes for the next row
+            // Update column dtypes for the next row
+            // There are a couple ways to do this, each with some nuance:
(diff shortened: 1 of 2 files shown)


src/geopackage/sqlite/iterator.cpp
```diff
@@ -76,7 +76,13 @@ sqlite_iter& sqlite_iter::next()
         const int returncode = sqlite3_step(this->ptr());
         if (returncode == SQLITE_DONE) {
             this->iteration_finished = true;
-        } else if (returncode != SQLITE_ROW) {
+        }
+        else if( returncode == SQLITE_ROW){
+            //Update column dtypes for the next row
+            for (int i = 0; i < this->column_count && i < column_types.size(); i++) {
+                this->column_types[i] = (sqlite3_column_type(this->ptr(), i));
```
PhilMiller, line 83: Does this need to be done for every row returned by the iterator, or just once for the first row we get? Why can/would column types differ from row to row?

PhilMiller (commented):
You've got a typo in your second commit message - 'alt_it' vs 'alt_id', that would make this commit harder to find by searching later

src/geopackage/sqlite/iterator.cpp
hellkite500: If the value in a column is `NaN`, then `sqlite3_column_type` will report that as a `5` or NULL type.  So if the first row contains `NaN` but other rows have valid data, then all types are reported as `NULL`.  (This is how we discovered this issue...)

Re-checking for each row ensures we get a non-null type if there is valid data in the column in that row.

hellkite500:
> You've got a typo in your second commit message - 'alt_it' vs 'alt_id', that would make this commit harder to find by searching later

Third time's a charm, should be fixed now...

src/geopackage/sqlite/iterator.cpp
program--: For additional context, `sqlite3` has table definitions that include a schema (i.e. `PRAGMA table_info(...);`), however, when interacting with rows, sqlite does **not** refer to the types in the table definition, but instead rely on row-local affinity (or something of that nature) to determine types -- so columns do not have a type associated wholly. Hence, the need to check each row.

The other alternative to this (but a bit more involved) is to pull the types from the table definition when an iterator is created. This solution of pulling types on each iteration is a bit quicker to get into the codebase though.

hellkite500: [80de41f](https://github.com/NOAA-OWP/ngen/pull/673/commits/80de41f30142e1294891d48c1c7b573002091c86)

donaldwj (approved):

PR #675, opened 2023-11-09
Update ngen build timeout to 20 minutes

hellkite500:
Give runner a little more time to build ngen successfully before timing out.


hellkite500:
Hmm, this may not be needed (#670 and #567) if the slow build is an artifact of the optimization issue documented.

hellkite500:
#670 fixes the optmization/build issues from #567 that were causing excessively long build times, so the original timeout is once again reasonable.

PR #677, opened 2023-11-13
Move get_var_value_as_double up hierarchy, as it was inadvertently duplicated as a non-override

PhilMiller:
This method exists in both `Bmi_Module_Formulation` (from which it's overridden by the language-specific implementations) and `Bmi_Multi_Formulation`, with the exact same interface and apparent intended use.

## Changes

- Declare `get_var_value_as_double` at the common base class, not separately in different child classes

## Testing

- Ordinary CI

## Checklist

### DIFF
--- a/include/realizations/catchment/Bmi_Formulation.hpp
+++ b/include/realizations/catchment/Bmi_Formulation.hpp
@@ -219,7 +219,7 @@ namespace realization {
          * Get value for some BMI model variable.
          *
          * This function assumes that the given variable, while returned by the model within an array per the BMI spec,
-         * is actual a single, scalar value.  Thus, it returns what is at index 0 of the array reference.
+         * is actually a single, scalar value.  Thus, it returns what is at index 0 of the array reference.
          *
          * @param index
          * @param var_name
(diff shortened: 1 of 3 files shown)


include/realizations/catchment/Bmi_Formulation.hpp
```diff
@@ -215,6 +215,38 @@ namespace realization {
             *output_text_stream << std::setprecision(output_precision);
         }
 
+        /**
+         * Get value for some BMI model variable.
+         *
+         * This function assumes that the given variable, while returned by the model within an array per the BMI spec,
+         * is actual a single, scalar value.  Thus, it returns what is at index 0 of the array reference.
```
PhilMiller, line 222: I should line-wrap this better and fix the `actual` typo

program-- (approved):
I don't think we hit this case, but we don't have any diamond inheritance schemes, correct? (i.e. needing virtual inheritance on either `Bmi_Formulation` or `Bmi_Module_Formulation`) -- Particularly w.r.t. the BMI base class. (This is probably out of scope for this PR, but still partly relevant, so I figured I'd ask in case we noticed this) -- Other than that, this looks good to me.

PhilMiller:
There is some multiple inheritance in the current code (which I'm not thrilled about), but I don't think any of it hits diamond inheritance; it's all just interface mix-in sorts of things.

Thanks for the quick review

program-- (approved):

PR #683, opened 2023-11-22
escape regex characters that may appear in identifiers, fixes #680

### LINKED ISSUE #680: Error find forcings with catchment with irregular id format
As noted in NOAA-OWP/hydrofabric#28, there's at least one catchment in the current hydrofabric with an id in an unexpected format:  `cat-7e+05`.  An error occurs related to this catchment when trying to run the framework, which is how it was discovered:

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Forcing data could not be found for 'cat-7e+05'
```

However, a forcing file does exist at `/dmod/datasets/forcing/regridded/regrid_02/csv/cat-7e+05.csv`.   The forcing portion of the realization config is:

```json
        "forcing": {
           "file_pattern": "{{id}}.csv",
           "path": "/dmod/datasets/forcing/regridded/regrid_02/csv/",
           "provider":  "CsvPerFeature"
        }
```

hellkite500:
## Changes

- When searching for forcing files by pattern, escape regex characters that may be in the ID string before making the pattern

## TODO

- Add test case for scientific notation ID

## Checklist

- [x] Linux
- [x] MacOS

### DIFF
--- a/include/realizations/catchment/Formulation_Manager.hpp
+++ b/include/realizations/catchment/Formulation_Manager.hpp
@@ -503,6 +503,18 @@ namespace realization {
                 if (path.compare(path.size() - 1, 1, "/") != 0) {
                     path += "/";
                 }
+                
+                //In case an idntifier has regex specific characters, need to escape them
+                //This may not be the most optimal algorithm, but for the size of strings we care about, it should work ok for now.
+		std::string::size_type n = 0;
+                for( const char& c : {'.', '+'} ){
+
+                    while ( ( n = identifier.find( c, n ) ) != std::string::npos )
+                    {
+                        identifier.replace( n, 1, std::string("\\\\")+c );
+                        n += 5; //start next find after we replace
+                    }
+                }
 
                 std::string filepattern = forcing_prop_map.at("file_pattern").as_string();
 


PhilMiller:
Please limit IDs to a set of characters that exclude anything that regex would need escaped, rather than going along this path.

If that means other scripts or data sets need to be fixed up to not emit/contain those characters, we should push that through.

hellkite500:
Hydrofabric identifiers are strings and something like `cat-7e+05` _is_ a valid divide identifier (it uniquely identifies the divide in a given hydrofabric).  While it may not be consistent, it is valid (noaa-owp/hydrofabric#28).  ngen *could* enforce more stringent requirements, but it isn't clear why that should be required.  We can definitely expand the list of explicitly escaped characters to the full set of regex special characters:
`., +, *, ?, ^, $, (, ), [, ], {, }, |, \.`
which would ensure that any string could be pattern matched by the internal regex built in ngen.

PhilMiller (changes requested):
Mike agree to limit the catchment IDs to a safer set of characters. Once we have the updated hydrofabric, we can validate against the narrowed spec, and this escaping will be moot.

PR #690, opened 2023-12-22
Automated "test_all" target and Code Coverage

program--:
This PR reworks the `test_all` target to automatically amalgamate all tests defined through `ngen_add_test`, and builds on #348 to add code coverage using `gcovr` instead of `lcov`. In addition to using `gcovr`, modifications were made to `CodeCoverage.cmake` to support using `llvm-cov` with `gcov` emulation.

## Additions

- CMake option `NGEN_WITH_COVERAGE`. Dependent on `NGEN_WITH_TESTS`, and allows building the `ngen_coverage_*` targets.
- `cmake/CodeCoverage.cmake` module, which provides functions for generating HTML and XML code coverage documents.
- Targets `ngen_coverage_html` and `ngen_coverage_xml` for HTML and XML documents respectively.
  The output files for these targets are outputted to:
  + `<build_directory>/<target_name>/index.html`
  + `<build_directory>/<target_name>.xml`.

## Changes

- Target `test_all` is now automatically generated through `ngen_add_test` rather than explicitly at the end of the `test/CMakeLists.txt` file.

## Notes

- When generating code coverage documents, the `.gcda` files are generated alongside the source object files in `CMakeFiles/` directories. As a result, an out-of-source build is necessary, since it is not worth attempting to clean up these files in an in-source build. Moreover, to generate successive coverage documents, the gcov data files need to be deleted beforehand, so deleting the out-of-source build directory is the recommended approach.
- If using a custom compiler with (e.g.) `-DCMAKE_CXX_COMPILER=/path/to/bin/gcc-23`, the path to the corresponding coverage analysis tool needs to be set with `-DGCOV_PATH=/path/to/bin/gcov-23`, or `-DLLVM_COV_PATH=/directory/with/llvm/bin/llvm-cov-42` as appropriate

- Example Coverage:

  image

## Testing

- HTML code coverage generated using GCC, Clang, and IntelLLVM.

## Checklist

- [X] Linux
- [x] MacOS

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,7 +94,6 @@ endif()
 if(NGEN_WITH_COVERAGE)
     include(CodeCoverage)
     append_coverage_compiler_flags()
-    link_libraries(--coverage)
 endif()
 
 # -----------------------------------------------------------------------------
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,7 @@ option(NGEN_QUIET            "Silence output"                     OFF)
 # Syntax: cmake_dependent_option(<option> "<help_text>" <value> <depends> <force>)
 cmake_dependent_option(NGEN_WITH_ROUTING "Build with t-route integration" ON "NGEN_WITH_PYTHON" OFF)
 cmake_dependent_option(NGEN_UPDATE_GIT_SUBMODULES "Update submodules on configure" ON "GIT_FOUND;NGEN_HAS_GIT_DIR" OFF)
+cmake_dependent_option(NGEN_WITH_COVERAGE "Build with test coverage" ON "NGEN_WITH_TESTS" OFF)
 
 option(BMI_FORTRAN_ISO_C_LIB_DIR "Directory hint for middleware Fortran shared lib handling iso_c_binding" "${NGEN_ROOT_DIR}/extern/iso_c_fortran_bmi/cmake_build")
 option(BMI_FORTRAN_ISO_C_LIB_NAME "Name for middleware Fortran shared lib handling iso_c_binding" "iso_c_bmi")
@@ -85,6 +86,11 @@ if(PROJECT_NAME STREQUAL CMAKE_PROJECT_NAME)
     set(NGEN_IS_MAIN_PROJECT ON)
 else()
     set(NGEN_IS_MAIN_PROJECT OFF)
+endif()
+
+if(NGEN_WITH_COVERAGE)
+    add_compile_options(--coverage -O0)
+    link_libraries(--coverage)
 endif()
 
 # -----------------------------------------------------------------------------
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,8 +88,12 @@ else()
     set(NGEN_IS_MAIN_PROJECT OFF)
 endif()
 
+# -----------------------------------------------------------------------------
+# If Coverage is enabled, include coverage module and
+# set compiler flags/linkage to accomodate coverage.
 if(NGEN_WITH_COVERAGE)
-    add_compile_options(--coverage -O0)
+    include(CodeCoverage)
+    append_coverage_compiler_flags()
     link_libraries(--coverage)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,7 +50,7 @@ option(NGEN_QUIET            "Silence output"                     OFF)
 # Syntax: cmake_dependent_option(<option> "<help_text>" <value> <depends> <force>)
 cmake_dependent_option(NGEN_WITH_ROUTING "Build with t-route integration" ON "NGEN_WITH_PYTHON" OFF)
 cmake_dependent_option(NGEN_UPDATE_GIT_SUBMODULES "Update submodules on configure" ON "GIT_FOUND;NGEN_HAS_GIT_DIR" OFF)
-cmake_dependent_option(NGEN_WITH_COVERAGE "Build with test coverage" ON "NGEN_WITH_TESTS" OFF)
+cmake_dependent_option(NGEN_WITH_COVERAGE "Build with test coverage" OFF "NGEN_WITH_TESTS" OFF)
 
 option(BMI_FORTRAN_ISO_C_LIB_DIR "Directory hint for middleware Fortran shared lib handling iso_c_binding" "${NGEN_ROOT_DIR}/extern/iso_c_fortran_bmi/cmake_build")
 option(BMI_FORTRAN_ISO_C_LIB_NAME "Name for middleware Fortran shared lib handling iso_c_binding" "iso_c_bmi")
--- a/cmake/CodeCoverage.cmake
+++ b/cmake/CodeCoverage.cmake
@@ -1,0 +1,427 @@
+# Copyright (c) 2012 - 2017, Lars Bilke
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification,
+# are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this
+#    list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions and the following disclaimer in the documentation
+#    and/or other materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its contributors
+#    may be used to endorse or promote products derived from this software without
+#    specific prior written permission.
+#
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -87,6 +88,11 @@ else()
     set(NGEN_IS_MAIN_PROJECT OFF)
 endif()
 
+if(NGEN_WITH_COVERAGE)
+    add_compile_options(--coverage -O0)
```
PhilMiller, line 92: If this noticeably slows the testing process at all, it should probably be OK to use `-Og`

cmake/CodeCoverage.cmake
```diff
@@ -0,0 +1,427 @@
+# Copyright (c) 2012 - 2017, Lars Bilke
```
PhilMiller, line 1: I take it there are no local modifications to this, and it's just a verbatim import?

PhilMiller (commented):
Thanks for doing this. It all looks good to me. I'll give it a quick run on macOS, too

cmake/CodeCoverage.cmake
program--: I made some modifications, primarily adding `llvm-cov` support, and preventing failures if a non-Apple LLVM-based compiler is used (the original fails if GNU or Apple Clang is *not* used). I also removed the `lcov` function that was originally defined, since I didn't want to handle both `gcovr` and `lcov`. I'll add these points to the changelog comments in that file

CMakeLists.txt
program--: I didn't notice it slowing down too much, but if you notice it on your end, then we can make this change 🙂

cmake/CodeCoverage.cmake
```diff
@@ -0,0 +1,427 @@
...
+find_program( LLVM_COV_PATH llvm-cov )
+find_program( LCOV_PATH  NAMES lcov lcov.bat lcov.exe lcov.perl)
+find_program( FASTCOV_PATH NAMES fastcov fastcov.py )
+find_program( GENHTML_PATH NAMES genhtml genhtml.perl genhtml.bat )
+find_program( GCOVR_PATH gcovr PATHS ${CMAKE_SOURCE_DIR}/scripts/test)
+find_program( CPPFILT_PATH NAMES c++filt )
+
+if(NOT GCOV_PATH OR NOT LLVM_COV_PATH)
+    message(FATAL_ERROR "gcov/llvm-cov not found! Aborting...")
+endif() # NOT GCOV_PATH
```
PhilMiller, line 153: It turns out this trips on my system. The logic seems to have gotten DeMorgan's law wrong, or was just ill-considered.

Really, it seems like it should be part of the following conditional, since which one actually gets used depends on the compiler in use.

Regardless, I'll just install llvm-cov if I can, and get on with testing.

PhilMiller (commented):
Would it be possible/unreasonable to enable coverage analysis in the CI pipeline builds?

cmake/CodeCoverage.cmake
program--: Yeah, looks like I screwed that up 🤦🏾 I'll get that fixed and aligned to the toolchain in use.

PhilMiller:
I just tested successfully on macOS with compilers from Homebrew. In addition to setting `-DCMAKE_C_COMPILER=$(which gcc-13) -DCMAKE_CXX_COMPILER=$(which g++-13) -DCMAKE_Fortran_COMPILER=$(which gfortran-13)`, I had to set `-DGCOV_PATH=$(which gcov-13)` to match. The result without that, of running the system's `/usr/bin/gcov` is that it will silently run and produce output with no files or functions listed.

PhilMiller:
Once the gcov/llvm-cov selection logic is fixed, I'm enthusiastically +1 on this

PhilMiller:
Something to potentially enhance down the line would be to exclude irrelevant bits from the metrics. The worst offender is the Google Test source code in `test/googletest/googletest`, since we don't care that we use very little of it.

If it's easy to do that, it would be great to fold that in here

PhilMiller:
It may also make sense to not measure coverage of the test files as such, and only report on `src/` and `include/`, but that's a mixed bag - if there's code in the tests that's not running, it should either be activated or deleted.

PhilMiller:
I discovered in testing this on my Mac that the exclusions that are supposed to limit results to just `src/` and `include/` get confused if the paths involve a symlink. Building through absolute paths instead resolves that.

I'm trying to patch that by resolving the paths, but it's not working as desired.

cmake/CodeCoverage.cmake
```diff
@@ -0,0 +1,457 @@
...
+    CACHE STRING "Flags used by the Fortran compiler during coverage builds."
+    FORCE )
+set(CMAKE_CXX_FLAGS_COVERAGE
+    ${COVERAGE_COMPILER_FLAGS}
+    CACHE STRING "Flags used by the C++ compiler during coverage builds."
+    FORCE )
+set(CMAKE_C_FLAGS_COVERAGE
+    ${COVERAGE_COMPILER_FLAGS}
+    CACHE STRING "Flags used by the C compiler during coverage builds."
+    FORCE )
```
PhilMiller, line 208:
```suggestion
set(CMAKE_CXX_FLAGS_COVERAGE
    ${COVERAGE_CXX_COMPILER_FLAGS}
    CACHE STRING "Flags used by the C++ compiler during coverage builds."
    FORCE )
set(CMAKE_C_FLAGS_COVERAGE
    ${COVERAGE_C_COMPILER_FLAGS}
    CACHE STRING "Flags used by the C compiler during coverage builds."
    FORCE )
```
Did you mean to be using the language-specific flags tested just above? Does that flag also apply to gfortran?

cmake/CodeCoverage.cmake
```diff
@@ -0,0 +1,457 @@
...
+        COMMAND ;
+        COMMENT "Open ./${Coverage_NAME}/index.html in your browser to view the coverage report."
+    )
+
+endfunction() # setup_target_for_coverage_gcovr_html
+
+function(append_coverage_compiler_flags)
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${COVERAGE_COMPILER_FLAGS}" PARENT_SCOPE)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${COVERAGE_COMPILER_FLAGS}" PARENT_SCOPE)
+    set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} ${COVERAGE_COMPILER_FLAGS}" PARENT_SCOPE)
```
PhilMiller, line 446: Having just recently seen build-breaking issues in the SCHISM build for BMI related to modifying `CMAKE_Lang_FLAGS` variables, it's much preferable to use `add_compile_options` as was being done previously

PhilMiller: Oh, is it like this because of the `PARENT_SCOPE` thing that `add_compile_options` can't do?

PhilMiller: Making it a macro instead could maybe address that?

cmake/CodeCoverage.cmake
```diff
@@ -0,0 +1,457 @@
...
+    # Show info where to find the report
+    add_custom_command(TARGET ${Coverage_NAME} POST_BUILD
+        COMMAND ;
+        COMMENT "Open ./${Coverage_NAME}/index.html in your browser to view the coverage report."
+    )
+
+endfunction() # setup_target_for_coverage_gcovr_html
+
+function(append_coverage_compiler_flags)
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${COVERAGE_COMPILER_FLAGS}" PARENT_SCOPE)
```
PhilMiller, line 444: As above, should these be `COVERAGE_C_COMPILER_FLAGS` and such?

PhilMiller:
There's maybe some race condition if one runs `make -j 8 ngen_coverage_html` de-novo in a freshly-configured tree, where the data seem not to be fully populated in the output. I'm not sure I can be bothered to care, but again, noting it here to document ["don't hold it that way"](https://www.theguardian.com/technology/blog/2010/jun/25/iphone-reception-problems-solved)

PhilMiller:
My remaining comments are probably fine to ignore. This achieves its purpose, and isn't otherwise making anything else around it worse. Let's just get it in, reap the benefits, and we can fix nitty-gritty issues as they become practical concerns.

PhilMiller (dismissed):

program--:
@PhilMiller I've modified the CMake module to include your suggestions, but particularly, removing all the language-specific flags and using `add_compile_options`/`link_libraries` as previously done. Should be good now 👍🏾

PhilMiller (approved):

PR #695, opened 2024-01-02
utility script for making animations of forcing inputs

hellkite500:
Script used to generate forcing animations for AGU 2023 fall meeting.

### DIFF
--- a/utilities/plotting/forcing_animation.py
+++ b/utilities/plotting/forcing_animation.py
@@ -1,0 +1,221 @@
+#!/usr/bin/env python
+"""
+@author Nels J. Frazier
+@date 01-02-2024
+@version 0.0.0
+
+Utility to generate animations in gif format in both gridded form
+and catchment aggregated form.  Use forcing_animation --help for
+descriptions of options and required inputs.
+
+"""
+import geopandas as gpd
+import matplotlib.pyplot as plt
+#import PIL
+#PIL.Image.MAX_IMAGE_PIXELS = 188745354+10000
+from matplotlib.animation import FuncAnimation, PillowWriter, FFMpegWriter
+import seaborn as sns
+import xarray as xr
+from functools import partial
+import cartopy.crs as ccrs
+import cartopy.feature as cfeature
+#import pickle
+import argparse
+from pyogrio import read_dataframe
+
+def plot_catchment_frame(ax, forecast, forcing, catchments, i):
+    t = forcing[i]
+    #each dataset is a slice of time
+    ts = t.time.dt.strftime("%Y-%m-%d %H:%M:%S").item()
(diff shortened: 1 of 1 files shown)


utilities/plotting/forcing_animation.py
```diff
@@ -0,0 +1,221 @@
...
+
+args = parser.parse_args()
+
+forecast = args.cycle+" Range "+args.source
+crs = ''
+epsg = ''
+if(args.command == 'hydrofabric'):
+    epsg = '5070'
+    crs = ccrs.epsg(epsg)
+    crs = ccrs.PlateCarree()
```
PhilMiller, line 75: Which one is `crs` meant to be?

hellkite500: The pyproj 5070 epsg code enforces a projection boundary that clips off the southern part of the CONUS domain, so for now using the PlateCarree.  See comment on line 160.  I'll clean this up a bit, just wanted to get it published in some form.

utilities/plotting/forcing_animation.py
```diff
@@ -0,0 +1,221 @@
...
+cmap = sns.color_palette("light:b", as_cmap=True)
+#Reasonably good "blue ish" cube helix
+#cmap = sns.color_palette("ch:start=.2,rot=-.2", as_cmap=True)
+
+bkgrnd = sns.color_palette('flare', as_cmap=True)
+#cmap.set_under(bkgrnd(50), alpha=0.05)
+cmap.set_under("bisque", alpha=0.1)
+e_color = 'none'
+vmin = forcing['RAINRATE'].min().item()
+vmin = 0.0009
```
PhilMiller, line 99: Another duplicate assignment

utilities/plotting/forcing_animation.py
```diff
@@ -0,0 +1,221 @@
...
+    # union = gpd.GeoSeries( catchments.unary_union, crs=catchments.crs)
+    union = gpd.read_file("domain.gpkg", engine='pyogrio')
+    
+    print("Projecting domain to epsg:4326")
+    union = union.to_crs(epsg='4326')
+    catchments = catchments.to_crs(epsg='4326')
+
+    forcing = forcing.set_index({"catchment ids":"catchment_ids"}).to_array().squeeze().drop_vars('variable')
+    #Get the catchment geopandas index, strip to int for xarray index
+    cat_ids = catchments['divide_id'].apply(lambda x: int(x.split("-")[1]))
```
PhilMiller, line 142: Presumably true at the time of writing, but does the hydrofabric work guarantee unique numeric IDs of catchments across the various kinds that potentially prefix them?

Actually, are there currently or plans for catchment IDs prefixed other than `cat-NNNN`, like we have for nexuses?

@mikejohnson51

PR #720, opened 2024-01-30
Forcings Engine Data Provider Interface and Lumped Forcings Implementation

### LINKED ISSUE #705: Forcing Engine gridded forcing into ngen

program--:
This PR adds a `ForcingsEngineDataProvider` base class and `ForcingsEngineLumpedDataProvider` implementation.

### Implementation Details
`ForcingsEngineDataProvider` is a templated base class taking two types, `DataType` and `SelectionType`, and inherits from `DataProvider<DataType, SelectionType>`. The base class handles Python module instantiation and timing functions, and leaves value acquisition to derived classes.

`ForcingsEngineLumpedDataProvider` is a derived class of `ForcingsEngineDataProvider<double, CatchmentAggrDataSelector>` representing the forcings engine in lumped catchment mode (i.e. `GRID_TYPE` in the config file is set to "hydrofabric").

Since the forcings engine provides values over an entire domain, the classes are designed in a flyweight-like manner, where construction and instance referencing are through static maps for each templated base class of `ForcingsEngineDataProvider`, i.e. each derived class instance is stored in a map of its base type.

Derived classes should implement a static construction function that uses the protected function `ForcingsEngineDataProvider::set_instance`, which moves a `unique_ptr` of the base class into the static map. Then, to get an instance, callers should either use the return from the derived function (i.e. `ForcingsEngineLumpedDataProvider::lumped_instance`), or get an existing instance using `ForcingsEngineDataProvider<DataType, SelectionType>::instance`. Ownership and lifetime management are held by the static maps, so these functions return raw pointers (that do not imply any ownership).

> Note: it may make sense to return references instead.

## Additions
- `ForcingsEngineDataProvider`
- `ForcingsEngineLumpedDataProvider`, and a corresponding test fixture `ForcingsEngineLumpedDataProviderTest`.
- `assert_forcings_engine_requirements`, which checks that the forcings engine python module is installed, and checks for the `$WGRIB2` environment variable.

## Changes
- Separate `NullForcingsEngine` implementation into source file so that the `NGen::forcings` CMake target always has a source.

## Notes
- To use the forcings engine, the user's environment requires NetCDF/PNetCDF, ParallelIO, ESMF (+ python bindings), and the wgrib2 executable.
- When running the forcings engine, the `WGRIB2` environment variable must be set to the path of the wgrib2 executable. If/when pywgrib2 is integrated, this will not be a requirement.

## To-Do
- [x] Allow forcings engine to be installable as a package (https://github.com/jduckerOWP/ngen-forcing/pull/3)

## Checklist

- [x] Linux
- [ ] macOS


include/forcing/ForcingsEngineDataProvider.hpp
PhilMiller, line 1: Given that trailing return types don't get used for anything else in the code base, and there's nothing that has template arguments so as to really benefit from them, they're more distracting than helpful. In this case, I'd prefer if they were switched to traditional leading return types.

include/forcing/ForcingsEngineDataProvider.hpp
```diff
@@ -0,0 +1,65 @@
...
+    );
+
+    ~ForcingsEngineDataProvider() override;
+
+    /**
+     * @brief Set the MPI communicator that the Forcings Engine should use.
+     *
+     * @param handle int representing a communicator handle.
+     */
+    void set_communicator(int handle);
```
PhilMiller, line 36: Before I really dig into `ForcingsEngine`, I think it's probably preferable for that to live in its own header, and for this to be a method only exposed on that and not the `Provider`. There's a weird side-effect of calling this on one `Provider` instance affecting all of them.

src/forcing/ForcingsEngineDataProvider.cpp
```diff
@@ -0,0 +1,428 @@
...
+    time_current_index_++;
+};
+
+ForcingsEngine& ForcingsEngine::instance(
+    const std::string& init,
+    std::size_t time_start,
+    std::size_t time_end
+)
+{
+    static ForcingsEngine inst_{init, time_start, time_end};
```
PhilMiller, line 157: At least make this a `class` `static` variable, and not a function static. The singleton-ish nature is a little bit disrupted because of how the function arguments of the first call get used in construction.

include/forcing/ForcingsEngine.hpp
```diff
@@ -0,0 +1,172 @@
...
+
+#include "Bmi_Py_Adapter.hpp"
+#include <utilities/mdarray.hpp>
+
+namespace data_access {
+
+struct ForcingsEngine
+{
+    using size_type                 = std::size_t;
+    using clock_type                = std::chrono::system_clock;
```
PhilMiller, line 16: Eventually we'll want to switch the time keeping to something a bit more thoughtful, ideally UTC-based, but this is fine for now.

include/forcing/ForcingsEngine.hpp
```diff
@@ -0,0 +1,172 @@
...
+     * @brief Get the index of a variable in the instance cache.
+     * 
+     * @param variable 
+     * @return size_type 
+     */
+    size_type variable_index(const std::string& variable) const noexcept;
+
+    /**
+     * @brief Get a forcing value from the instance cache, or update the instance
+     *        and its cache.
```
PhilMiller, line 83: What does the "or update" refer to here?
- Is that a relic of an earlier iteration of the design?
- The return value isn't `double&`, so it can't be assigned to
- Does calling this result in the engine population the requested entry if it wasn't present in the cache?

PhilMiller: Broadly speaking, functions used to access data should probably not document their impact on any internal caches, as such. Caching is an implementation detail.

include/forcing/ForcingsEngine.hpp
```diff
@@ -0,0 +1,172 @@
...
+     */
+    size_type divide_index(const std::string& divide_id) const noexcept;
+
+    /**
+     * @brief Get the index of a variable in the instance cache.
+     * 
+     * @param variable 
+     * @return size_type 
+     */
+    size_type variable_index(const std::string& variable) const noexcept;
```
PhilMiller, line 79: There's public functions for indices of time, divide, and variable, and then the actual data access function `at()` takes the key values, and not the indices. What's the deal with that?

include/forcing/ForcingsEngine.hpp
program--: Bad phrasing on my part, but it's the last bullet point there -- calling `at` with a later `raw_time` value will execute the model's `Update()` through to the new time and cache the values.

include/forcing/ForcingsEngine.hpp
program--: `at` uses these three functions, I intended to move them to the private interface but neglected that.

src/forcing/ForcingsEngine.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+namespace data_access {
+
+std::unordered_map<std::string, std::unique_ptr<ForcingsEngine>> ForcingsEngine::instances_{};
+
+ForcingsEngine::ForcingsEngine(const std::string& init, size_type time_start, size_type time_end)
+  : time_start_(std::chrono::seconds{time_start})
+  , time_end_(std::chrono::seconds{time_end})
+{
+    check_runtime_requirements();
```
PhilMiller, line 11: It looks like the things checked in there only need to happen once per run, not once per instance created

test/forcing/ForcingsEngineDataProvider_Test.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+TEST_F(ForcingsEngineDataProviderTest, Lookback)
+{
+    // Both selectors encompass the entire temporal domain,
+    // and since they're AOIs, they should have different values.
+    // This means, to test lookback, we can evaluate one selector,
+    // then attempt to use the second selector. If the values match
+    // then lookback did not work correctly. Otherwise, lookback worked.
+    //
+    // Note that this assumes the variable values are different for the SAME variable.
```
program--, line 138: FYI, this is slightly outdated I believe. I initially was testing the difference in value between catchments, but just below I'm testing the difference between variables in the same catchment.

src/forcing/ForcingsEngine.cpp
```diff
@@ -0,0 +1,239 @@
...
+auto ForcingsEngine::time_index(time_t ctime) const noexcept
+  -> size_type
+{
+    const auto epoch = clock_type::from_time_t(ctime);
+
+    if (epoch < time_begin() || epoch > time_end()) {
+        return bad_index;
+    }
+
+    const auto offset = (epoch - time_begin()) / time_step();
```
PhilMiller, line 129: Do we need to worry about rounding and loss of precision issues with this calculation? It's not clear to me whether the `time_point` representation is integral or floating point, since cppreference just says "signed arithmetic type"

src/forcing/ForcingsEngine.cpp
```diff
@@ -0,0 +1,239 @@
...
+    return time_end_;
+}
+
+auto ForcingsEngine::time_step() const noexcept
+  -> const clock_type::duration&
+{
+    return time_step_;
+}
+
+auto ForcingsEngine::time_index(time_t ctime) const noexcept
```
PhilMiller, line 120: Probably not going to happen before this PR gets merged, but we should really work all use of `time_t` out of our code-base. It's just not really appropriate for use in anything other than system calls

include/forcing/ForcingsEngine.hpp
PhilMiller: Honestly, I'd kinda rather `at` took (or had an overload that took) the index values instead, and callers kept some awareness of indices. Juggling strings and system time types is kinda silly for a core data structure.

src/forcing/ForcingsEngine.cpp
```diff
@@ -0,0 +1,239 @@
...
+  , time_end_(std::chrono::seconds{time_end})
+{
+    check_runtime_requirements();
+
+    bmi_ = std::make_unique<models::bmi::Bmi_Py_Adapter>(
+        "ForcingsEngine",
+        init,
+        "NextGen_Forcings_Engine.BMIForcingsEngine",
+        true,
+        true,
```
PhilMiller, line 18: We should maybe change the constructor for `Bmi_Py_Adapter` to take enums instead of `bool`, but for now, could you add a small comment on each of these to denote what formal parameter the `true` literals correspond to?

src/forcing/ForcingsEngine.cpp
program--: I _think_ (heavy emphasis on think) that this is an integral operation. At the very least, `decltype(epoch - time_begin())` evaluates to a `chrono::duration<long, std::ratio<...>>`, and the division of that against `decltype(time_step())` has a common type of `long`.

src/forcing/ForcingsEngine.cpp
```diff
@@ -0,0 +1,239 @@
...
+        utils::getStdOut()
+    );
+    
+    time_step_ = std::chrono::seconds{static_cast<int64_t>(bmi_->GetTimeStep())};
+    var_outputs_ = bmi_->GetOutputVarNames();
+    var_outputs_.erase(var_outputs_.begin()); // Erase "CAT-ID" from outputs
```
PhilMiller, line 24: Maybe I'm being paranoid, but I don't feel like trusting the underlying code to always return `CAT-ID` as the *first* output variable. Could this be a `find` call instead of `begin`?

test/forcing/ForcingsEngineDataProvider_Test.cpp
program--: To clarify, this test needs to be reworked since it's not testing anything meaningful I think.

src/forcing/ForcingsEngine.cpp
```diff
@@ -0,0 +1,239 @@
...
+        // Get current values for each variable
+        const auto var_span = boost::span<const double>{
+            static_cast<double*>(bmi_->GetValuePtr(var_outputs_[vi])),
+            size
+        };
+
+        for (size_type di = 0; di < size; di++) {
+            // Set current variable value for each divide
+            var_cache_.at({{time_current_index_, di, vi}}) = var_span[di];
+        }
```
PhilMiller, line 48: This may be a pre-mature optimization, but I think it would be good to have some sort of ranged `set` operation in the `mdarray` itself that can do this a bit more efficiently. If nothing else, the "store values from a new time step" functionality is repeated below inside `at`, so it should at least be factored out to its own helper function

src/forcing/ForcingsEngine.cpp
```diff
@@ -0,0 +1,239 @@
...
+    return offset;
+}
+
+auto ForcingsEngine::divide_index(const std::string& divide_id) const noexcept
+  -> size_type
+{
+
+    const auto  id_sep   = divide_id.find('-');
+    const char* id_split = id_sep == std::string::npos
+                           ? divide_id.data()
```
PhilMiller, line 139:
```suggestion
                           ? &divide_id[0]
```
Just for reading symmetry

src/forcing/ForcingsEngine.cpp
```diff
@@ -0,0 +1,239 @@
...
+                           : &divide_id[id_sep + 1];
+    const int   id_int   = std::atoi(id_split);
+
+    const auto pos = std::lower_bound(var_divides_.begin(), var_divides_.end(), id_int);
+    if (pos == var_divides_.end() || *pos != id_int) {
+        return bad_index;
+    }
```
PhilMiller, line 146: Is this not just `find`?

src/forcing/ForcingsEngine.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+    // implicit cast to unsigned
+    return std::distance(var_divides_.begin(), pos);
```
PhilMiller, line 149: I'm a bit concerned about assumptions made in this indexing algorithm, but I'm not sure if it's warranted. Do we have a guarantee of the numerical uniqueness of every divide label in the hydrofabric? If we really need that, it should probably be getting automatically checked somewhere at startup.

src/forcing/ForcingsEngine.cpp
```diff
@@ -0,0 +1,239 @@
...
+            "Failed to get ForcingsEngine value at index {" + index_str + "}" +
+            "\n  Shape    : {" + shape_str + "}" +
+            "\n  Time     : `" + time_str  + "`" +
+            "\n  Divide   : `" + divide_id + "`" +
+            "\n  Variable : `" + variable  + "`"
+        };
+    }
+
+    // If time index is past current index, 
+    // update until new index and cache values.
```
PhilMiller, line 201: Is there any mechanism (planned) to discard data from timesteps earlier than (some window before) the current one? Or is it intentional to accumulate the forcings values for the entire run?

PhilMiller: The usual meaning of a cache is something with a fixed or limited size, providing ready access to values that would otherwise be accessed from a slower source (e.g. re-read/compute from the backing data)

src/forcing/ForcingsEngineDataProvider.cpp
```diff
@@ -0,0 +1,103 @@
...
+    const auto start = ForcingsEngine::clock_type::from_time_t(selector.get_init_time());
+    const auto end   = std::chrono::seconds{selector.get_duration_secs()} + start;
+
+    std::vector<double> values;
+    for (auto i = start; i < end; i += engine_->time_step()) {
+        const auto current_time = ForcingsEngine::clock_type::to_time_t(i);
+        values.push_back(
+            engine_->at(current_time, selector.get_id(), selector.get_variable_name())
+        );
+    }
```
PhilMiller, line 98: It seems a little bit silly to have to convert `i` back to `time_t` to pass it down to the `ForcingsEngine` when you just asked the engine to do said conversion. Why not re-type to be able to just pass `i` (and then rename the loop variable to `current_time`)?

src/forcing/ForcingsEngine.cpp
```diff
@@ -0,0 +1,282 @@
...
+  -> size_type
+{
+    const auto epoch = clock_type::from_time_t(ctime);
+    return time_index(epoch);
+}
+
+auto ForcingsEngine::time_index(clock_type::time_point epoch) const noexcept
+  -> size_type
+{
+    if (epoch < time_begin() || epoch > time_end()) {
```
PhilMiller, line 140: The way `time_dim` gets calculated, I think this may have to test `epoch >= time_end()` - i.e. it's a half-open interval, like C++'s usual `begin()/end()` iterator pairs

src/forcing/ForcingsEngineDataProvider.cpp
```diff
@@ -0,0 +1,103 @@
...
+}
+
+auto ForcingsEngineDataProvider::get_values(const CatchmentAggrDataSelector& selector, ReSampleMethod m)
+  -> std::vector<double>
+{
+    const auto start = ForcingsEngine::clock_type::from_time_t(selector.get_init_time());
+    const auto end   = std::chrono::seconds{selector.get_duration_secs()} + start;
+
+    std::vector<double> values;
```
PhilMiller, line 92: I'm not a fan of allocating a vector of length matching the number of time steps in the interval in the underlying data, storing it all, and then digesting it. If we're getting the values one at a time, we can digest them one at a time, in constant memory, without any dynamic allocation.

src/forcing/ForcingsEngineDataProvider.cpp
```diff
@@ -0,0 +1,103 @@
...
+    return timegm(&tm_);
+}
+
+ForcingsEngineDataProvider::ForcingsEngineDataProvider(
+    const std::string& init,
+    std::size_t time_start,
+    std::size_t time_end
+) : engine_(
+    &ForcingsEngine::instance(init, time_start, time_end)
+){};
```
PhilMiller, line 21:
```suggestion
){}
```

src/forcing/ForcingsEngineDataProvider.cpp
```diff
@@ -0,0 +1,103 @@
...
+ForcingsEngineDataProvider::ForcingsEngineDataProvider(
+    const std::string& init,
+    const std::string& time_start,
+    const std::string& time_end,
+    const std::string& time_fmt
+) : ForcingsEngineDataProvider(
+    init,
+    parse_time(time_start, time_fmt),
+    parse_time(time_end, time_fmt)
+){};
```
PhilMiller, line 32:
```suggestion
){}
```

PhilMiller (commented):
CI will probably complain about those semicolons too

src/forcing/ForcingsEngine.cpp
program--: I reverted this change -- I think the conditional logic may need revisiting. If we check for greater than or equal, then we'll get a `bad_index` value exactly at the end time; this happens in the `Timestepping` test, but I _think_ we'd want the last possible index if we call `get_ts_index_for_time` with the end time.

What may need to change is that if `epoch >= time_end()`, we always return the last possible index (i.e. in the case of the unit tests, that would be 5 for any time past `params.simulation_end_t`).

@PhilMiller Does that make sense to you? Or do you have a different suggestion?

PhilMiller: Before getting into the implementation details, is the interval spanning `time_begin(), time_end()` open or closed? I know there are actually deeper questions about the data semantics buried beneath this (e.g. instantaneous vs trailing average vs centered average vs etc), but I'm just wondering about implementation details. It seems implicit that queries at `time_begin()` are valid, and expected to produce whatever data. Are queries at `time_end()` valid, down the call stack? Do we expect such queries up the call stack?

I know there's some pieces of code here and in related codebases that do the indefinite extension of values past the end of the recorded observation/forecast period. I don't think we should immediately implement that here until we're very clear on whether we want it, whether it needs to be controlled in some fashion, and whether there are distinctions between requests past `time_end()` that are valid and those that are simply erroneous.

include/forcing/ForcingsEngineDataProvider.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+    long get_data_stop_time() override;
+
+    long record_duration() override;
+
+    size_t get_ts_index_for_time(const time_t& epoch_time) override;
+
+    double get_value(const CatchmentAggrDataSelector& selector, ReSampleMethod m) override;
+
+    std::vector<double> get_values(const CatchmentAggrDataSelector& selector, data_access::ReSampleMethod m) override;
```
PhilMiller, line 42: There's an inconsistency in using qualified `data_access::ReSampleMethod` vs unqualified `ReSampleMethod` here. For the interface, I think it's preferable to use the qualified name, and there's hopefully no `using` declaration in scope.

src/forcing/ForcingsEngine.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+ForcingsEngine& ForcingsEngine::instance(
+    const std::string& init,
+    std::size_t time_start,
+    std::size_t time_end
+)
+{
+    if (instances_[init] == nullptr) {
+        instances_[init] = std::make_unique<ForcingsEngine>(init, time_start, time_end);
+    }
```
PhilMiller, line 77: Is it valid to ask for a `ForcingsEngine` instance with an `init` matching some other request, but different time arguments? Given the discussion about checking begin/end time, and possibly extending data, that seems to invite trouble.

src/forcing/ForcingsEngine.cpp
program--: Following the other data providers, it should be a half-open interval, i.e. `[time_begin(), time_end)`. However, I think we want to have an index at `time_end`, because, I _think_, the semantic of `bmi.UpdateUntil(time_end)` is consistent with the half-open interval (i.e. we can update up to the end time, but not past it).

As far as the forcings engine is concerned, it will update until while its current time value is less than whatever sent to it, as long as the necessary GRIB files are available.

It probably makes sense _not_ to extend the the index to invalid times, so I agree with that last part.

src/forcing/ForcingsEngine.cpp
program--: The apparent situation that comes to mind is if we are running a separate workflow under a new temporal domain in the same execution space. To me, this seems like it should be an invalid call.

I can add two asserts to ensure the time arguments match that of the instance received, so we can see if this situation (or similar) occurs and address it then.

src/forcing/ForcingsEngine.cpp
program--: I factored out the helper function in [6edd99a](https://github.com/NOAA-OWP/ngen/pull/720/commits/6edd99acbfb710ba2294225d3e65d893c04d90ff) for now. I'm going to hold off on trying to implement a ranged set op in mdarray and probably do that in a separate PR

src/forcing/ForcingsEngine.cpp
program--: We should, but I'd have to double check. I'll take a look at implementing a check for uniqueness, but most likely directly in the Forcings Engine's python code (I _think_ it makes more sense there)

src/forcing/ForcingsEngine.cpp
PhilMiller: The asserts of matching time arguments sound like a good answer for now.

src/forcing/ForcingsEngine.cpp
PhilMiller: For a longer run, this is ultimately going to be a big concern, since it'll be a growing memory footprint without bound.

src/forcing/ForcingsEngineDataProvider.cpp
```diff
@@ -0,0 +1,105 @@
...
+    if (m == ReSampleMethod::SUM || m == ReSampleMethod::MEAN) {
+        double acc = 0.0;
+        for (auto current_time = start; current_time < end; current_time += engine_->time_step()) {
+            acc += engine_->at(current_time, id, var);
+        }
+
+        if (m == ReSampleMethod::MEAN) {
+            acc /= static_cast<double>(
+                selector.get_duration_secs() / std::chrono::duration_cast<std::chrono::seconds>(engine_->time_step()).count()
+            );
```
PhilMiller, line 81: This expression seems to have gotten long and unwieldy. Could you break the denominator calculation into one or a few local variables?

src/forcing/ForcingsEngineDataProvider.cpp
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        return acc;
+    }
+
+    throw std::runtime_error{"Given ReSampleMethod " + std::to_string(m) + " not implemented."};
+}
+
+std::vector<double> ForcingsEngineDataProvider::get_values(const CatchmentAggrDataSelector& selector, ReSampleMethod m)
```
PhilMiller, line 90: With the change to `get_value`, is there any expectation that `get_values` will see any use, or can it be deleted?

program--: `get_values` and `get_value` are both virtual functions from `data_access::GenericDataProvider` -- one or the other is potentially called around `Bmi_Module_Formulation.cpp:626`

src/forcing/ForcingsEngine.cpp
program--: My bad, I thought I replied to this one 😅. Yeah I want to reduce the memory footprint this takes up by only storing the last timestep's variable values. Just haven't refactored that in yet

src/forcing/ForcingsEngineDataProvider.cpp
program--: Resolved in [be18dff](https://github.com/NOAA-OWP/ngen/pull/720/commits/be18dffa965cfcaac4b2726140642beef712b819).

src/forcing/ForcingsEngineDataProvider.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+    if (m == ReSampleMethod::SUM || m == ReSampleMethod::MEAN) {
+        double acc = 0.0;
+        for (auto current_time = start; current_time < end; current_time += engine_->time_step()) {
+            acc += engine_->at(current_time, id, var);
+        }
+
+        if (m == ReSampleMethod::MEAN) {
+            const auto time_step_seconds = std::chrono::duration_cast<std::chrono::seconds>(engine_->time_step()).count();
+            acc /= selector.get_duration_secs() / time_step_seconds;
```
PhilMiller, line 80: Just for clarity's sake, I'd further suggest

```suggestion
            const auto time_step_seconds = std::chrono::duration_cast<std::chrono::seconds>(engine_->time_step()).count();
            const auto num_time_steps = selector.get_duration_secs() / time_step_seconds;
            acc /= num_time_steps;
```

That double division in a single statement, one of which is `/=`, is a bit brain-bending.

Also, your response checks my comprehension that we are in fact dividing by the number of time steps being sampled across, and averaged.

program--: Ah yeah that's fair, added in [40d900b](https://github.com/NOAA-OWP/ngen/pull/720/commits/40d900b133bae4644562f484faccec722d8545f3).

include/forcing/ForcingsEngineDataProvider.hpp
```diff
@@ -0,0 +1,50 @@
+#ifndef NGEN_FORCING_FE_PER_FEATURE_DATA_PROVDER_HPP
+#define NGEN_FORCING_FE_PER_FEATURE_DATA_PROVDER_HPP
+
+#include "GenericDataProvider.hpp"
+#include "ForcingsEngine.hpp"
+
+namespace data_access {
+
+struct ForcingsEngineDataProvider
```
PhilMiller, line 9: Having just gone over the `DataProvider` class hierarchy, I think what we want here may be to declare
```suggestion
template <class data_type, class selection_type>
struct ForcingsEngineDataProvider : public DataProvider<data_type, selection_type>
```

And then have further derived classes `ForcingsEngineLumpedDataProvider` (in this PR), `ForcingsEngineGridDataProvider`, and `ForcingsEngineMeshDataProvider` (each to come later).

Along the way, we could do away with `GenericDataProvider` as a separate class definition, and just make it an alias
`using GenericDataProvider = DataProvider<double, CatchmentAggrDataSelector>;`
So that `ForcingsEngineLumptedDataProvider` would be a super-type and hence pointers would be convertible.

All of the above implies that `GenericDataProvider` should probably be renamed to something like `CatchmentLumpedDataprovider`, if it's kept at all.

test/forcing/ForcingsEngineDataProvider_Test.cpp
```diff
@@ -0,0 +1,186 @@
...
+    constexpr std::array<const char*, 8> expected_variables = {
+        "U2D_ELEMENT",
+        "V2D_ELEMENT",
+        "LWDOWN_ELEMENT",
+        "SWDOWN_ELEMENT",
+        "T2D_ELEMENT",
+        "Q2D_ELEMENT",
+        "PSFC_ELEMENT",
+        "RAINRATE_ELEMENT"
+    };
```
PhilMiller, line 108:
```suggestion
    };
    ASSERT_EQ(outputs_test.size(), expected_variables.size());
```

src/forcing/ForcingsEngineLumpedDataProvider.cpp
```diff
@@ -0,0 +1,241 @@
...
+    ReSampleMethod m
+)
+{
+    const auto start = clock_type::from_time_t(selector.get_init_time());
+    const auto end   = std::chrono::seconds{selector.get_duration_secs()} + start;
+    const auto step  = std::chrono::seconds{this->record_duration()};
+    const std::string id = selector.get_id();
+    const std::string var = selector.get_variable_name();
+
+    std::vector<double> values;
```
PhilMiller, line 233:
```suggestion
    std::vector<double> values;
    values.reserve((end-start)/step);
```
?

PhilMiller (commented):
Overall, this looks pretty good to me. I think we could potentially get this all hammered out and merged in the next few days.

include/forcing/ForcingsEngineDataProvider.hpp
```diff
@@ -0,0 +1,149 @@
...
+        return inst.get();
+    }
+
+    template<typename Callable>
+    static ForcingsEngineDataProvider* set_instance(const std::string& init, Callable&& f)
+    {
+        std::unique_ptr<ForcingsEngineDataProvider> tmp = nullptr;
+        if (instances_[init] != nullptr) {
+            tmp = std::move(instances_[init]);
+        }
```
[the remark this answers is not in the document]
program--: Ah, I was a bit confused about this. That's the reason I set `tmp`, since I wasn't sure if assigning over `instances_[init]` would destroy any previous value there. Thanks for the info, I'll adjust this 🙂

test/forcing/ForcingsEngineDataProvider_Test.cpp
```diff
@@ -0,0 +1,186 @@
...
+    
+    for (decltype(auto) output : boost::combine(outputs_test, expected_variables)) {
+        EXPECT_EQ(output.get<0>(), output.get<1>());
+    }
```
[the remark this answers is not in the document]
program--: No, that's a good point -- it isn't necessary that the names be in any particular order.

PhilMiller:
Mind if we make this change as well as part of this PR?

```
diff --git a/include/forcing/GenericDataProvider.hpp b/include/forcing/GenericDataProvider.hpp
index 6bdeeb4f0..d4c6bd7ae 100644
--- a/include/forcing/GenericDataProvider.hpp
+++ b/include/forcing/GenericDataProvider.hpp
@@ -6,12 +6,7 @@
 
 namespace data_access
 {
-    class GenericDataProvider : public DataProvider<double, CatchmentAggrDataSelector>
-    {
-        public:
-
-        private:
-    };
+    using GenericDataProvider = DataProvider<double, CatchmentAggrDataSelector>;
 }
```

include/forcing/ForcingsEngineDataProvider.hpp
```diff
@@ -80,50 +61,79 @@ struct ForcingsEngineDataProvider
...
+        var_output_names_ = bmi_->GetOutputVarNames();
     }
 
+    static ForcingsEngineDataProvider* set_instance(
+        const std::string& init,
+        std::unique_ptr<ForcingsEngineDataProvider>&& instance
+    )
+    {
+          using std::swap;
+          swap(instances_[init], instance);
```
PhilMiller, line 125: I'm still not seeing why this needs to me more complicated than
```suggestion
          instances_[init] = std::move(instance);
```
Is there some subtle issue around destructor timing that I'm missing? That's the only reason I can think of for grabbing the existing value in `instances_[init]`

src/forcing/ForcingsEngineLumpedDataProvider.cpp
```diff
@@ -40,29 +42,44 @@ Provider::ForcingsEngineLumpedDataProvider(
...
-  : Provider(init, parse_time(time_begin, time_fmt), parse_time(time_end, time_fmt))
-{}
+{
+    auto time_begin_epoch = static_cast<size_t>(parse_time(time_begin, time_fmt));
+    auto time_end_epoch = static_cast<size_t>(parse_time(time_end, time_fmt));
+
+
+    auto provider = std::unique_ptr<ForcingsEngineLumpedDataProvider>{
+        new Provider{init, time_begin_epoch, time_end_epoch}
+    };
```
PhilMiller, line 79:
```suggestion
    auto provider = std::make_unique<ForcingsEngineLumpedDataProvider>(
        init, time_begin_epoch, time_end_epoch
    );
```
In the `unique_ptr` case, it doesn't make a tangible difference, but it's generally good practice to use `make_unique` and `make_shared` to avoid edge conditions around exceptions, and to get better memory allocation behavior and layout in the `shared_ptr` case.

include/forcing/ForcingsEngineDataProvider.hpp
program--: No you're right 🤦🏾 This was an artifact of some in-between changes, but after resolving that I forgot to change the swap back to a move assignment

src/forcing/ForcingsEngineLumpedDataProvider.cpp
program--: This one can't be `std::make_unique` because the constructor for `ForcingsEngineLumpedDataProvider` is private (due to constraining construction/access to the instance getter `ForcingsEngineDataProvider::instance`), so without `new` it causes a compilation error -- maybe my design isn't idiomatic though

PhilMiller: Ok, this is fine to leave as-is, then. As noted, `make_unique` really just exists for symmetry with `make_shared`.

src/forcing/ForcingsEngineLumpedDataProvider.cpp
```diff
@@ -0,0 +1,259 @@
...
+                break;
+            }
+
+            acc += this->at(id, var);
+            current_time += step;
+        }
+
+        if (m == ReSampleMethod::MEAN) {
+            const auto time_step_seconds = step.count();
+            const auto time_duration = std::chrono::duration_cast<std::chrono::seconds>(current_time.time_since_epoch()).count();
```
PhilMiller, line 228:
```suggestion
            const auto time_duration = std::chrono::duration_cast<std::chrono::seconds>(current_time - start).count();
```
I think I'm misunderstanding, but should this be relative to `start` and not the common epoch?

src/forcing/ForcingsEngineLumpedDataProvider.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+        return acc;
+    }
+
+    throw std::runtime_error{"Given ReSampleMethod " + std::to_string(m) + " not implemented."};
+}
+
+std::vector<double> Provider::get_values(
+    const CatchmentAggrDataSelector& selector,
+    ReSampleMethod m
```
PhilMiller, line 241: Does this need to constrain acceptable values of `ReSampleMethod`, or gain the same functionality as the scalar case?

program--: I'm not sure to be honest. This one is tricky for me because I don't think resampling makes much sense in context of the `get_values` function -- at least, not for `CatchmentAggrDataSelector`, maybe for gridded data selectors with a target grid spec -- so I'm not sure what that would look like in this case.

I think my understanding of the implementation was that `get_values` gets all the values for that variable & catchment at each timestep in the given selector duration, so no resampling can really be done -- however, with the `get_value` function, it's resampled across the duration, so it makes sense there. I may be missing something though...

include/forcing/ForcingsEngineDataProvider.hpp
```diff
@@ -0,0 +1,168 @@
...
+    //       a single argument: `clock_type::duration time_duration`, since the forcings engine
+    //       manages time via a duration rather than time points. !! Need to double check
+    ForcingsEngineDataProvider(
+      const std::string& init,
+      std::size_t time_begin_seconds,
+      std::size_t time_end_seconds
+    )
+      : time_begin_(std::chrono::seconds{time_begin_seconds})
+      , time_end_(std::chrono::seconds{time_end_seconds})
+    {
```
PhilMiller, line 115:
```suggestion
    {
        assert_forcings_engine_requirements();
```
I think this would be the right place to re-introduce this call, since it's on relatively low-count construction of the shared instances, not per-catchment.

src/forcing/ForcingsEngineLumpedDataProvider.cpp
```diff
@@ -0,0 +1,250 @@
...
+        const auto var_span = boost::span<const double>{
+            static_cast<double*>(bmi_->GetValuePtr(outputs[vi])),
+            size
+        };
+
+        for (std::size_t di = 0; di < size; di++) {
+            // Move 0 time (current) to 1 (previous), and set the
+            // current values from var_span
+            var_cache_.at({{1, di, vi}}) = var_cache_.at({{0, di, vi}});
+            var_cache_.at({{0, di, vi}}) = var_span[di];
```
PhilMiller, line 136: Not a request to change anything now, but for your awareness, there's a common pattern in HPC codes for old/new values of an array to avoid copying. It's to have two storage instances, and swap (or literally `swap`) between them. If we end up wanting/needing to optimize this, that's probably what we'd implement.

test/forcing/ForcingsEngineLumpedDataProvider_Test.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+    const auto outputs = provider_->get_available_variable_names();
+
+    ASSERT_EQ(outputs.size(), expected_variables.size());
+
+    // Check that each output variable exists in the list of expected variables
+    for (const auto& output : outputs) {
+        EXPECT_NE(
+            std::find(expected_variables.begin(), expected_variables.end(), output),
+            expected_variables.end()
```
PhilMiller, line 132: There's a subtle mistake here
```suggestion
    for (const auto& expected : expected_variables) {
        EXPECT_NE(
            output.find(expected),
            output.end()
```
If `outputs` consisted of N copies of a single expected variable, the test as written would have passed. We want to check that *every* expected variable is present.

Given that, (I know I'm back-tracking), do we want to assert the absence of *additional* variables? If not, then the size equality check just above could be dropped.

program--: Ah I noticed this too, but couldn't sort the logic to correctly check it in my head. Good catch 🙂 I think we don't need the additional assertion.

src/forcing/ForcingsEngineLumpedDataProvider.cpp
```diff
@@ -0,0 +1,250 @@
...
+double Provider::get_value(
+    const CatchmentAggrDataSelector& selector,
+    ReSampleMethod m
+)
+{
+    const auto start = clock_type::from_time_t(selector.get_init_time());
+    const auto end = std::chrono::seconds{selector.get_duration_secs()} + start;
+    const auto step = std::chrono::seconds{this->record_duration()};
+    const std::string id = selector.get_id();
+    const std::string var = selector.get_variable_name();
```
PhilMiller, line 203: Per Slack conversation, these should go through `divide_index()` and `variable_index()` outside the loop

src/forcing/ForcingsEngineLumpedDataProvider.cpp
```diff
@@ -0,0 +1,250 @@
...
+    const auto start = clock_type::from_time_t(selector.get_init_time());
+    const auto end   = std::chrono::seconds{selector.get_duration_secs()} + start;
+    const auto step  = std::chrono::seconds{this->record_duration()};
+    const std::string id = selector.get_id();
+    const std::string var = selector.get_variable_name();
+
+    std::vector<double> values;
+    values.reserve((end - start) / step);
+    for (auto current_time = start; current_time < end; current_time += step) {
+        values.push_back(this->at(id, var));
```
PhilMiller, line 244: Per Slack conversation, this needs to call `next()`

src/forcing/ForcingsEngine.cpp
PhilMiller: @mikejohnson51 Does the Hydrofabric make this promise that the numeric part of each divide ID is unique?

PhilMiller: Documenting discussion from the stand-up today, Justin is implementing a uniqueness check. That's the last outstanding concern I have with this PR.

program--: Added in [48e724c](https://github.com/NOAA-OWP/ngen/pull/720/commits/48e724c5bdbca5aeae89bc33b6e793ff37a55e22) -- the approach uses a temporary `std::unordered_map<int, int>` which increments the value on each divide_id, and checks for >1 values. If it finds one, a `std::runtime_error` is thrown -- let me know if there's any concerns with that

src/forcing/ForcingsEngineLumpedDataProvider.cpp
```diff
@@ -42,9 +42,19 @@ Provider::ForcingsEngineLumpedDataProvider(
...
+    std::unordered_map<int, int> uniq;
+    uniq.reserve(id_dim);
     var_divides_.reserve(id_dim);
     for (int i = 0; i < id_dim; ++i) {
-        var_divides_[cat_id[i]] = i;
+        const auto id = cat_id[i];
+    
+        uniq[id]++;
+        if (uniq[id] > 1) {
+            throw std::runtime_error{"Non-unique catchment ID found in lumped forcings engine domain: cat-" + std::to_string(id)};
```
PhilMiller, line 54: Let's not encode an assumption that the prefix is `cat-` here. Preferable to just print the number exactly as is.

Otherwise, this looks good.

program--: Revised in [8e8fa99](https://github.com/NOAA-OWP/ngen/pull/720/commits/8e8fa9926e97f38f51c9f740be6c43651369e23b)

PhilMiller (commented):
I'm happy with this. It would probably be good to get a fresh review from someone else, since I've poked at so many bits of this along the way.

include/forcing/ForcingsEngineLumpedDataProvider.hpp
```diff
@@ -0,0 +1,113 @@
...
+     * 
+     * @param divide_id Divide ID to index at
+     * @param variable Forcings variable to get
+     * @param previous If true, return the previous timestep values.
+     * @return double 
+     */
+    double at(
+        const std::string& divide_id,
+        const std::string& variable,
+        bool previous = false
```
PhilMiller, line 71: So, I feel rather late pointing this out, but `at()` is private, defaults `previous = false`, and has no internal callers for which it's passed an explicit `true`. At that point, what are we getting for doubling our memory footprint and the work of double-buffering the data in that memory?

program--: Initially I thought that requiring the previous timestep values would be necessary so that catchments (in the same process) referencing the same forcings engine instance did not pass-over a step needed by other catchments -- but looking back through the main NGen code, I think it was a premature implementation and won't be needed -- so I will take out that portion of this PR

PhilMiller:
@donaldwj @hellkite500 Could one of you review this? I'm happy with it, but I've also had a lot of say in its progression, so I don't want to be the one to sign off on it.

Issue #703, opened 2024-01-12
Simplify parallel run on single processor


hellkite500:
A QoL feature for running single process runs with a parallel build.

## Current behavior

When using a parallel build, but only a single processor, e.g. `mpirun -n 1` it is currently required to generate a single full domain partition and pass that to ngen when executing.

## Expected behavior

When run with only a single processor, don't require a partition file, and run the full hydrofabric domain.

PR #721, opened 2024-02-02
Parallel run with one proc w/o partition file

stcui007:
This PR address issue #703 in ngen framework, i.e., to run a MPI job using one processor without providing a partition file. For more than one processor, you must provide a partition file. If you insist on providing the partition file in the case of one processor, you are still ok.

## Additions

- Define a dummy class `PartitionOne` for to present a whole-hydrofabric partition for a single process, and associated test

## Removals

None

## Changes

- Use `PartitionOne` when running on a single MPI process, rather than expecting a partition file argument on the command line

## Testing

Passed all local test except RoutingPyBindTest.TestRoutingPyBind
Passing all automated testing

## Notes

Command line example to run a job without/with partition file:

    mpirun -n 1 ./cmake_build_mpi/ngen data/catchment_data.geojson '' data/nexus_data.geojson '' data/example_bmi_multi_realization_config.json
 
    mpirun -n 3 ./cmake_build_mpi/ngen data/catchment_data.geojson '' data/nexus_data.geojson '' data/example_bmi_multi_realization_config.json ./test_partition_3.json

## Checklist

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -40,8 +40,8 @@ jobs:
         ref: gh-pages
         clean: false
         
-    - name: Move generated files where git can see them
-      run: cp -rp docs/html/* .
+    - name: Prevent generated docs dir from being committed and overwriting on the next run.
+      run: rm -Rf docs/html
 
     - name: Prevent generated docs dir from being committed and overwriting on the next run.
       run: rm -Rf docs/html
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -46,9 +46,6 @@ jobs:
     - name: Prevent generated docs dir from being committed and overwriting on the next run.
       run: rm -Rf docs/html
 
-    - name: Prevent generated docs dir from being committed and overwriting on the next run.
-      run: rm -Rf docs/html
-
     - name: Update branch information
       run: git fetch --all; git branch --list -r;
       
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -39,6 +39,9 @@ jobs:
       with:
         ref: gh-pages
         clean: false
+
+    - name: Move generated files where git can see them
+      run: cp -rp docs/html/* .
         
     - name: Prevent generated docs dir from being committed and overwriting on the next run.
       run: rm -Rf docs/html
--- a/include/core/Partition_Data.hpp
+++ b/include/core/Partition_Data.hpp
@@ -1,10 +1,11 @@
 #ifndef PARTITION_DATA_H
 #define PARTITION_DATA_H
 
-using Tuple = std::tuple<int, std::string, std::string, std::string>;
+//using Tuple = std::tuple<int, std::string, std::string, std::string>;
 
 struct PartitionData
 {
+using Tuple = std::tuple<int, std::string, std::string, std::string>;
     int mpi_world_rank;
     std::unordered_set<std::string> catchment_ids;
     std::unordered_set<std::string> nexus_ids;
--- a/include/core/Partition_Data.hpp
+++ b/include/core/Partition_Data.hpp
@@ -1,11 +1,9 @@
 #ifndef PARTITION_DATA_H
 #define PARTITION_DATA_H
 
-//using Tuple = std::tuple<int, std::string, std::string, std::string>;
-
 struct PartitionData
 {
-using Tuple = std::tuple<int, std::string, std::string, std::string>;
+    using Tuple = std::tuple<int, std::string, std::string, std::string>;
     int mpi_world_rank;
     std::unordered_set<std::string> catchment_ids;
     std::unordered_set<std::string> nexus_ids;
--- a/include/core/Partition_Data.hpp
+++ b/include/core/Partition_Data.hpp
@@ -1,0 +1,14 @@
+#ifndef PARTITION_DATA_H
+#define PARTITION_DATA_H
+
+using Tuple = std::tuple<int, std::string, std::string, std::string>;
+
+struct PartitionData
+{
+    int mpi_world_rank;
+    std::unordered_set<std::string> catchment_ids;
+    std::unordered_set<std::string> nexus_ids;
+    std::vector<Tuple> remote_connections;
+};
+
+#endif  //PARTITION_DATA_H
--- a/include/core/Partition_One.hpp
+++ b/include/core/Partition_One.hpp
@@ -25,7 +25,6 @@ class Partition_One {
          */
         void generate_partition(geojson::GeoJSON& catchment_collection)
         {
-            //PartitionData partition_data;
             for(auto& feature: *catchment_collection)
             {
                 std::string cat_id = feature->get_id();
@@ -38,7 +37,6 @@ class Partition_One {
 
     	virtual ~Partition_One(){};
 
-        //PartitionData is a struct
         PartitionData partition_data;       
 };
 
--- a/include/core/Partition_One.hpp
+++ b/include/core/Partition_One.hpp
@@ -1,0 +1,47 @@
+#ifndef PARTITION_ONE_H
+#define PARTITION_ONE_H
+
+#ifdef NGEN_MPI_ACTIVE
+
+#include <iostream>
+#include <memory>
+#include <string>
+#include <vector>
+#include <unordered_set>
+#include <FeatureBuilder.hpp>
+#include "features/Features.hpp"
+#include <FeatureCollection.hpp>
+#include "Partition_Data.hpp"
+
+class Partition_One {
+
+    public:
+        Partition_One() {};
+
+        void generate_partition(geojson::GeoJSON& catchment_collection)
+        {
+            int counter = 0;
+            for(auto& feature: *catchment_collection)
+            {
+                std::string cat_id = feature->get_id();
+                catchment_ids.emplace(cat_id);
+                std::string nex_id = feature->get_property("toid").as_string();
+                nexus_ids.emplace(nex_id);
+                counter++;
+            }
+            std::cout << "counter = " << counter << std::endl;
+        }
+
+    	virtual ~Partition_One(){};
+
+        PartitionData partition_data;       
+
+    private:
+        int mpi_world_rank;
+        std::unordered_set<std::string> catchment_ids;
+        std::unordered_set<std::string> nexus_ids;
+        std::vector<Tuple> remote_connections;
+};
+
+#endif // NGEN_MPI_ACTIVE
+#endif // PARTITION_ONE_H
--- a/include/core/Partition_One.hpp
+++ b/include/core/Partition_One.hpp
@@ -25,28 +25,21 @@ class Partition_One {
          */
         void generate_partition(geojson::GeoJSON& catchment_collection)
         {
+            //PartitionData partition_data;
             for(auto& feature: *catchment_collection)
             {
                 std::string cat_id = feature->get_id();
-                catchment_ids.emplace(cat_id);
+                partition_data.catchment_ids.emplace(cat_id);
                 std::string nex_id = feature->get_property("toid").as_string();
-                nexus_ids.emplace(nex_id);
+                partition_data.nexus_ids.emplace(nex_id);
             }
             partition_data.mpi_world_rank = 0;
-            partition_data.catchment_ids = catchment_ids;
-            partition_data.nexus_ids = nexus_ids;
         }
 
     	virtual ~Partition_One(){};
 
         //PartitionData is a struct
         PartitionData partition_data;       
-
-    private:
-        int mpi_world_rank;
-        std::unordered_set<std::string> catchment_ids;
-        std::unordered_set<std::string> nexus_ids;
-        std::vector<Tuple> remote_connections;
 };
 
 #endif // NGEN_MPI_ACTIVE
--- a/include/core/Partition_One.hpp
+++ b/include/core/Partition_One.hpp
@@ -18,22 +18,25 @@ class Partition_One {
     public:
         Partition_One() {};
 
+        /**
+         * The function that parses geojson::GeoJSON data and build unordered sets of catchment_ids and nexus_ids
+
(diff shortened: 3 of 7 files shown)


include/core/Partition_Data.hpp
```diff
@@ -0,0 +1,14 @@
+#ifndef PARTITION_DATA_H
+#define PARTITION_DATA_H
+
+using Tuple = std::tuple<int, std::string, std::string, std::string>;
+
+struct PartitionData
+{
```
PhilMiller, line 7:
```suggestion
struct PartitionData
{
    using Tuple = std::tuple<int, std::string, std::string, std::string>;
```
Don't declare a global type named `Tuple`

include/core/Partition_One.hpp
```diff
@@ -0,0 +1,53 @@
...
+    	virtual ~Partition_One(){};
+
+        //PartitionData is a struct
+        PartitionData partition_data;       
+
+    private:
+        int mpi_world_rank;
+        std::unordered_set<std::string> catchment_ids;
+        std::unordered_set<std::string> nexus_ids;
+        std::vector<Tuple> remote_connections;
```
PhilMiller, line 49: These members seem to duplicate the members of `PartitionData`. Why not just use them?

include/core/Partition_Data.hpp
stcui007: I tried to make the change. But it caused some build errors in ```Partition_Parser.hpp```. Apparently, ```Partition_Parser.hpp``` uses this declaration in a few places in its code region. In its current form, the ```Partition_Parser.hpp``` would be the same as its original form in master branch.

include/core/Partition_One.hpp
stcui007: These variables are used this way following the example of ```Partition_Parser.hpp```. In this form, it seems the logic in the loop above is simpler. We could use the struct members, but we would seem to use longer variable names with somewhat overuse of the "." operator.

include/core/Partition_Data.hpp
PhilMiller: It would become `PartitionData::Tuple`.

include/core/Partition_One.hpp
PhilMiller: Syntactic simplicity isn't a good reason to duplicate data in memory and add extra assignments, as well as just adding extra declarations that anyone reading the code later has to make sense of.

include/core/Partition_Data.hpp
stcui007: Right. Implemented.

include/core/Partition_One.hpp
stcui007: Revised as requested.

include/core/Partition_Data.hpp
```diff
@@ -0,0 +1,15 @@
+#ifndef PARTITION_DATA_H
+#define PARTITION_DATA_H
+
+//using Tuple = std::tuple<int, std::string, std::string, std::string>;
+
+struct PartitionData
+{
+using Tuple = std::tuple<int, std::string, std::string, std::string>;
```
PhilMiller, line 8:
```suggestion
    using Tuple = std::tuple<int, std::string, std::string, std::string>;
```

include/core/Partition_Data.hpp
```diff
@@ -0,0 +1,15 @@
+#ifndef PARTITION_DATA_H
+#define PARTITION_DATA_H
+
+//using Tuple = std::tuple<int, std::string, std::string, std::string>;
+
```
PhilMiller, line 5:
```suggestion
```

include/core/Partition_One.hpp
```diff
@@ -0,0 +1,46 @@
...
+        Partition_One() {};
+
+        /**
+         * The function that parses geojson::GeoJSON data and build unordered sets of catchment_ids and nexus_ids
+
+         * @param catchment_collection the geojson::GeoJSON data containing all the necessary hydrofabric info
+         */
+        void generate_partition(geojson::GeoJSON& catchment_collection)
+        {
+            //PartitionData partition_data;
```
PhilMiller, line 28:
```suggestion
```

include/core/Partition_One.hpp
```diff
@@ -0,0 +1,46 @@
...
+                partition_data.catchment_ids.emplace(cat_id);
+                std::string nex_id = feature->get_property("toid").as_string();
+                partition_data.nexus_ids.emplace(nex_id);
+            }
+            partition_data.mpi_world_rank = 0;
+        }
+
+    	virtual ~Partition_One(){};
+
+        //PartitionData is a struct
```
PhilMiller, line 41:
```suggestion
```
```suggestion
        //PartitionData is a struct
```

include/core/Partition_Parser.hpp
```diff
@@ -63,8 +52,9 @@ class Partitions_Parser {
             std::string remote_nex_id;
             std::string remote_cat_id;
             std::string direction;
-            Tuple tmp_tuple;
-            std::vector<Tuple> remote_conn_vec;
+            //Tuple tmp_tuple;
```
PhilMiller, line 55:
```suggestion
```

stcui007:
Thanks Phil for the review. Will work on it first thing in the morning
Monday.

test/CMakeLists.txt
```diff
@@ -446,12 +461,15 @@ ngen_add_test(
...
         NGen::geojson
+        NGen::geopackage
         NGen::realizations_catchment
         NGen::mdarray
         NGen::mdframe
         NGen::logging
         NGen::ngen_bmi
         testbmicppmodel
+    REQUIRES
+        NGEN_WITH_SQLITE
```
PhilMiller, line 472: I'm concerned about making the entire `test_unit` build dependent on having SQLite support enabled.

include/core/Partition_One.hpp
```diff
@@ -0,0 +1,44 @@
...
+                std::string cat_id = feature->get_id();
+                partition_data.catchment_ids.emplace(cat_id);
+                std::string nex_id = feature->get_property("toid").as_string();
+                partition_data.nexus_ids.emplace(nex_id);
+            }
+            partition_data.mpi_world_rank = 0;
+        }
+
+    	virtual ~Partition_One(){};
+
```
PhilMiller, line 39:
```suggestion
```
There are no virtual functions here, and nothing at all for a destructor to even do, so we don't need to write one out at all.

include/core/Partition_One.hpp
```diff
@@ -0,0 +1,44 @@
...
+#include <FeatureBuilder.hpp>
+#include "features/Features.hpp"
+#include <FeatureCollection.hpp>
+#include "Partition_Data.hpp"
+
+class Partition_One {
+
+    public:
+        Partition_One() {};
+
```
PhilMiller, line 20:
```suggestion
```
The compiler will provide a default constructor without any need to write something out. Specifying a non-default constructor can create unintended side effects.

test/CMakeLists.txt
PhilMiller: Maybe the test case could say

```
         #if !defined(NGEN_WITH_SQLITE3)
            GTEST_SKIP() << "Compiled without SQLite support"
         #endif
```
?

test/utils/Partition_One_Test.cpp
```diff
@@ -0,0 +1,162 @@
...
+    std::set<std::string> duplicates;
+    //use set difference to identify all duplicates
+    std::set_difference(cat_id_vec.begin(), cat_id_vec.end(), unique.begin(), unique.end(), std::inserter(duplicates, duplicates.end()));
+    if( duplicates.size() > 0 ){
+        for( auto& id: duplicates){
+        }
+        exit(1);
+    }
+
+    ASSERT_TRUE(true);
```
PhilMiller, line 129:
```suggestion
    for( auto& id: duplicates){
        // Add printing code as needed
    }

    ASSERT_EQ(duplicates.size(), 0);
```
Unit tests should never be calling `exit()` directly, since that kills the whole test process, and doesn't properly report the failed test result.

The loop over duplicates doesn't need to be conditional, since if there are none, it does nothing.

Same applies to 1b below.

PhilMiller:
I pushed fixes for some of the smaller things I noted in my review. I wanted to call my concerns about the test code to your attention.

stcui007:
Could you be more explicit?

stcui007:
Ignore my previous question if you don't have any more to add. Just read your comments on the test codes.

test/CMakeLists.txt
stcui007: The problem is with the ```geopackage``` is involved in the test code.

stcui007:
I just put a couple of comments in the PR page.

test/CMakeLists.txt
program--: Haven't reviewed this, but IIRC from an earlier discussion, you could take out the GeoPackage reading, and just use GeoJSON, since it isn't needed to test the functionality of this PR.

stcui007:
I have made an update by making Partition_One an independent test in
analogy with some other packages.

.github/workflows/documentation.yml
```diff
@@ -39,9 +39,12 @@ jobs:
...
         clean: false
-        
+
     - name: Move generated files where git can see them
       run: cp -rp docs/html/* .
+        
+    - name: Prevent generated docs dir from being committed and overwriting on the next run.
+      run: rm -Rf docs/html
 
     - name: Prevent generated docs dir from being committed and overwriting on the next run.
```
PhilMiller, line 49: Why and how did this step get duplicated?

test/utils/Partition_One_Test.cpp
```diff
@@ -0,0 +1,165 @@
...
+        for (auto & path_option : parent_dir_options)
+            name_combinations.push_back(path_option + file_basename);
+
+        return utils::FileChecker::find_first_readable(name_combinations);
+    }
+
+    void read_file_generate_partition_data()
+    {
+        const std::string file_path = file_search(hydro_fabric_paths, "catchment_data.geojson");
+        if (boost::algorithm::ends_with(catchmentDataFile, "gpkg")) {
```
PhilMiller, line 56: Nothing ever sets `catchmentDataFile`, so this condition will never be true. Why does the test depend on Geopackage at all? If it's not being used, just delete it!

include/core/Partition_One.hpp
```diff
@@ -0,0 +1,40 @@
+#ifndef PARTITION_ONE_H
+#define PARTITION_ONE_H
+
+#ifdef NGEN_MPI_ACTIVE
```
PhilMiller, line 4: The guard for `NGEN_MPI_ACTIVE` isn't necessary here. MPI's not actually referenced or used by any of this code. Removing the guard would simplify test code by removing the weird `#define NGEN_MPI_ACTIVE`

.github/workflows/documentation.yml
stcui007: No idea. The only way that I can think of is during rebasing. At no time that I even touched that file. Did rebasing senses something different even human eyes see they are exactly the same? Any way, I will remove the repeat.

test/utils/Partition_One_Test.cpp
stcui007: Agree, If we need to use gpkg in the future, we can add it back with some revision.

include/core/Partition_One.hpp
stcui007: The presence of ```NGEN_MPI_ACTIVE``` does make the code look a bit cumbersome. However, Partition_One is specifically designed for this case. The source codes are under this condition. On the other hand, I assume if the test code works w/o MPI. it is probably ok.

stcui007: The ```NGEN_MPI_ACTIVE``` is needed as Partition_One.hpp needs it. Without it, the test code won't compile. The Parttion_Test.cpp using Partition_Parser.hpp works in a similar fashion. So this is kept in place.

test/utils/Partition_One_Test.cpp
```diff
@@ -0,0 +1,153 @@
...
+    std::set<std::string> duplicates;
+    //use set difference to identify all duplicates
+    std::set_difference(cat_id_vec.begin(), cat_id_vec.end(), unique.begin(), unique.end(), std::inserter(duplicates, duplicates.end()));
+
+    for( auto& id: duplicates){
+        if (!id.empty()) {
+            std::cout << "duplicates string set is not empty" << std::endl;
+            break;
+        }
+    }
```
PhilMiller, line 117:
```suggestion
    for (auto& id: duplicates) {
        std::cout << "duplicates string set contains " << id << std::endl;
    }
```
The `if` condition is testing for `duplicates` containing an element that is an empty string (`id` is a `std::string`. That makes no sense. If `duplicates` contains any elements at all, the error message applies, and it should print the element(s) found.

The exact same concern applies to test 1b below.

stcui007: The if condition tests for duplicates that's not empty by the ```negation operator```. So if that condition holds, we print out a message that says there are duplicates. Then break out of the loop so we don't need to print out a lot of lines if there are many in the duplicates.

test/utils/Partition_One_Test.cpp
```diff
@@ -0,0 +1,147 @@
...
+    };
+}
+
+TEST_F(PartitionOneTest, TestPartitionData_1a)
+{
+    read_file_generate_partition_data();
+
+    partition_one.generate_partition(catchment_collection);
+    PartitionData data_struct = partition_one.partition_data;
+    catchment_ids = data_struct.catchment_ids;
```
PhilMiller, line 95: I just realized an oversight in these tests - we also need to check that `catchment_ids` and `nexus_ids` contain the right number of elements.

Also, I just noticed that the comparison isn't the partition output against the original `catchment_collection`, but just against itself. The test as a whole should affirm that the set in the single partition is identical to the set in the input it was given.

stcui007:
Thanks for the careful review. Will work on it.

PhilMiller:
Code looks good. I'm going to wait until we've got #767 merged to rebase and merge this.

PhilMiller (approved):

PR #725, opened 2024-02-08
Remove global include from FindUDUNITS2.cmake

program--:
This PR fixes an issue where non-default UDUNITS2 installations are propagating its include path to all targets. This becomes an issue if the non-default path has a conflicting header (i.e. `bmi.h`), as files will prefer that over other search paths.

For example, if `bmi-c` is installed in the same non-default path, this can cause an issue in compiling CFE because there's a non-BMI define in its header:

```
<ngen_dir>/extern/cfe/cfe/src/bmi_cfe.c:1588:56: error: use of undeclared identifier 'BMI_MAX_LOCATION_NAME'
            strncpy(location, output_var_locations[i], BMI_MAX_LOCATION_NAME);
                                                       ^
<ngen_dir>/extern/cfe/cfe/src/bmi_cfe.c:1595:55: error: use of undeclared identifier 'BMI_MAX_LOCATION_NAME'
            strncpy(location, input_var_locations[i], BMI_MAX_LOCATION_NAME);
```

For reference, this issue was discovered by checking the exported compilation database of a macOS host attempting to build `ngen`, and noticing that `-I/usr/local/include` was added to every translation unit.

## Removals

- Removes global include statement. ~The comment shouldn't be relevant anymore, since the transition to a more targets-based build process.~ The comment is still relevant, but our approach should change so that the include issue requires user-intervention instead of forcing onto the user, i.e. see [comment below](https://github.com/NOAA-OWP/ngen/pull/725#issuecomment-1936375955).

## Checklist

- [X] Linux
- [x] macOS

### DIFF
--- a/.github/actions/ngen-build/action.yaml
+++ b/.github/actions/ngen-build/action.yaml
@@ -86,7 +86,6 @@ runs:
               echo "LD_LIBRARY_PATH=/usr/local/lib/:$LD_LIBRARY_PATH" >> $GITHUB_ENV
               echo "LDFLAGS=-L/usr/local/lib" >> $GITHUB_ENV
               echo "LIB=$LD_LIBRARY_PATH" >> $GITHUB_ENV
-              echo "INCLUDE=/usr/local/include" >> $GITHUB_ENV
               echo "NETCDF=/usr/local" >> $GITHUB_ENV
             fi
           shell: bash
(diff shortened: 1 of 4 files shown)


program--:
Huh, interesting that CI fails, but locally it works. I'll have to follow the target link paths, since this should be fine...

program--:
It seems like UDUNITS2 installations depend on package manager. [AUR](https://aur.archlinux.org/packages/udunits) installs the header directly into `/usr/include`, whereas [RHEL](https://rhel.pkgs.org/7/epel-x86_64/udunits2-devel-2.2.20-2.el7.x86_64.rpm.html) installs into `/usr/include/udunits2/`.

program--:
`__has_include` workaround works for RHEL, but not for macOS... I'm guessing now it's because of homebrew's install path, yet I don't know why CMake isn't locating this correctly -- it doesn't make much sense to me 😵‍💫

program--:
Previous errors are now resolved, in a similar manner as what was originally done. However, the primary change is that instead of using a global `include_directories()` call to add `/usr/local/include`, we should recommend setting the environment variables:
```
C_INCLUDE_PATH=/usr/local/include
CPLUS_INCLUDE_PATH=/usr/local/include
```
or, have an option to force those when configuring a NGen build.

PhilMiller (approved):

PR #750, opened 2024-02-29
Refactor: BMI adapters' virtual destructor hierarchy and general cleanup

### LINKED ISSUE #747: BMI adapter destructors are not entirely virtual
Currently, the `BMI_Adapter` class does not define a destructor, in any capacity. However, the `AbstractCLibAdapter` defines a virtual destructor that is overridden (but not explicitly with `override`) in its subclasses (C++, C, Fortran adapters).

Moreover, the BMI Python Adapater class does **not** define a destructor. Therefore, python models' `finalize()` method is never called.

## Current behavior

`BMI_Adapter` does not define a destructor.

## Expected behavior

`BMI_Adapter` should define a virtual destructor, and **all** children classes should override it explicitly with `override`.

program--:
This PR resolves #747 by pulling the virtual destructor hierarchy one level above `AbstractCLibBmiAdapter`, so that `Bmi_Adapter` defines a virtual destructor.

Moreover, this PR includes some additional cleanup of the BMI adapter classes and supporting files.

## Additions

- Adds a default virtual destructor to `Bmi_Adapter`.
- Adds an overridden destructor to `Bmi_Py_Adapter`.
- Adds source files for `AbstractCLibBmiAdapter` and `Bmi_Adapter`.

## Changes

- Reworks Bmi listfile to remove call to dynamically sourced library, and instead manually creates the CMake target to handle conditional compilation properly.
- Replaces `virtual` destructor attribute with `override` in subclasses of `Bmi_Adapter` and `AbstractCLibBmiAdapter`.
- Updates and sorts include directives in header and source files.

## Checklist

- [x] Linux
- [x] macOS

### DIFF
--- a/include/bmi/AbstractCLibBmiAdapter.hpp
+++ b/include/bmi/AbstractCLibBmiAdapter.hpp
@@ -36,7 +36,7 @@ namespace models {
              * Note that this performs the logic in the `Finalize()` function for cleaning up this object and its
              * backing BMI model.
              */
-            ~AbstractCLibBmiAdapter() override;
+            ~AbstractCLibBmiAdapter() noexcept override;
 
             /**
              * Perform tear-down task for this object and its backing model.
--- a/include/bmi/AbstractCLibBmiAdapter.hpp
+++ b/include/bmi/AbstractCLibBmiAdapter.hpp
@@ -5,8 +5,6 @@
 
 #include "Bmi_Adapter.hpp"
 
-#include "utilities/ExternalIntegrationException.hpp"
-
 namespace models {
     namespace bmi {
 
@@ -28,21 +26,9 @@ namespace models {
              */
             AbstractCLibBmiAdapter(const std::string &type_name, std::string library_file_path, std::string bmi_init_config,
                                    std::string forcing_file_path, bool allow_exceed_end, bool has_fixed_time_step,
-                                   std::string registration_func, utils::StreamHandler output)
-                    : Bmi_Adapter(type_name, std::move(bmi_init_config), std::move(forcing_file_path),
-                                     allow_exceed_end, has_fixed_time_step, output),
-                      bmi_lib_file(std::move(library_file_path)),
-                      bmi_registration_function(std::move(registration_func)) { }
+                                   std::string registration_func, utils::StreamHandler output);
 
-            AbstractCLibBmiAdapter(AbstractCLibBmiAdapter &&adapter) noexcept :
-                    Bmi_Adapter(std::move(adapter)),
-                    bmi_lib_file(std::move(adapter.bmi_lib_file)),
-                    bmi_registration_function(adapter.bmi_registration_function),
-                    dyn_lib_handle(adapter.dyn_lib_handle)
-            {
-                // Have to make sure to do this after "moving" so the original does not close the dynamically loaded library handle
-                adapter.dyn_lib_handle = nullptr;
-            }
+            AbstractCLibBmiAdapter(AbstractCLibBmiAdapter &&adapter) noexcept;
 
             /**
              * Class destructor.
@@ -50,9 +36,7 @@ namespace models {
              * Note that this performs the logic in the `Finalize()` function for cleaning up this object and its
              * backing BMI model.
              */
-            ~AbstractCLibBmiAdapter() override {
-                finalizeForLibAbstraction();
-            }
+            ~AbstractCLibBmiAdapter() override;
 
             /**
              * Perform tear-down task for this object and its backing model.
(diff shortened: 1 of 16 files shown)


include/bmi/AbstractCLibBmiAdapter.hpp
```diff
@@ -27,31 +26,17 @@ namespace models {
              */
             AbstractCLibBmiAdapter(const std::string &type_name, std::string library_file_path, std::string bmi_init_config,
```
program--, line 27: Formatting is a bit odd in the diff, but I didn't want to mix all these changes with formatting changes as well.

include/bmi/AbstractCLibBmiAdapter.hpp
```diff
@@ -192,24 +97,18 @@ namespace models {
...
-            }
+            const void *get_dyn_lib_handle();
 
         private:
 
             /** Path to the BMI shared library file, for dynamic linking. */
             std::string bmi_lib_file;
             /** Name of the function that registers BMI struct's function pointers to the right module functions. */
-            const std::string bmi_registration_function;
+            std::string bmi_registration_function;
```
program--, line 111: Removed const specifier, since it isn't necessary here. Really, the getter for this should be const instead.

include/bmi/Bmi_Adapter.hpp
```diff
@@ -20,29 +20,13 @@ namespace models {
...
-                            + this->bmi_init_config + "'. Error: "+std::strerror(errno);
-                    throw std::runtime_error(init_exception_msg);
-                }
-            }
+                        bool has_fixed_time_step, utils::StreamHandler output);
 
             Bmi_Adapter(Bmi_Adapter const&) = delete;
             Bmi_Adapter(Bmi_Adapter &&) = default;
 
+            virtual ~Bmi_Adapter() = default;
```
program--, line 28: For clarity, this is set to default instead of calling `Finalize()` to prevent undefined behavior, since `Finalize()` is virtual.

include/bmi/Bmi_Py_Adapter.hpp
```diff
@@ -44,6 +47,10 @@ namespace models {
             Bmi_Py_Adapter(Bmi_Py_Adapter const&) = delete;
             Bmi_Py_Adapter(Bmi_Py_Adapter&&) = delete;
 
+            ~Bmi_Py_Adapter() override {
+                Finalize();
+            }
```
program--, line 52: Unlike Bmi_Adapter, it should be fine to call `Finalize()` here, since we are calling the "final" version here. To make it safer, `Bmi_*_Adapter` classes should probably have the `final` specifier in the declaration.

src/bmi/AbstractCLibBmiAdapter.cpp
program--, line 1: The new source files are formatted with clang-format, which is why the formatting changed from what was originally in the headers.

src/bmi/CMakeLists.txt
```diff
@@ -1,25 +1,37 @@
-include(${PROJECT_SOURCE_DIR}/cmake/dynamic_sourced_library.cmake)
-dynamic_sourced_cxx_library(ngen_bmi "${CMAKE_CURRENT_SOURCE_DIR}")
```
program--, line 2: Dynamic sourcing is being removed because we should aim to eventually get rid of it completely. Conditional compilation is better managed by the build system rather than in the C++ files themselves, but this dynamic sourcing is preventing that from happening.

include/bmi/AbstractCLibBmiAdapter.hpp
```diff
@@ -192,24 +97,18 @@ namespace models {
...
+            void *dynamic_load_symbol(const std::string &symbol_name);
 
-            inline const std::string &get_bmi_registration_function() {
-                return bmi_registration_function;
-            }
+            const std::string &get_bmi_registration_function();
 
-            inline const void *get_dyn_lib_handle() {
-                return dyn_lib_handle;
-            }
```
program--, line 205: I could've replaced `inline` with `__attribute__((always_inline))` here instead to get the intended behavior, but opted to move definitions to a source file instead because it probably isn't worth it currently.

program-- (commented):
Added some review comments to help guide review and clarify design decisions. 😄

include/bmi/Bmi_Adapter.hpp
hellkite500: Should this then be a pure virtual function so that subclasses are required to implement a destructor (even if it is the default)?  We would have to double check the exact usage of `Bmi_Adapter` and make sure we aren't using it as a concrete type anywhere, but might be worth thinking about.

include/bmi/AbstractCLibBmiAdapter.hpp
hellkite500: Agree the getter could be marked const, but that would prevent future code added to the class from trying to change the `bmi_registration_function`, perhaps even inadvertently, which could lead to unexpected behavior.  I guess I'm trying to say, it doesn't hurt to be overly pedantic in these cases where we want the compiler to help with validation as much as possible.

include/bmi/Bmi_Adapter.hpp
program--: It can't be pure virtual because `Bmi_Adapter` has member variables, so you'll get a linker error on compilation, i.e. https://godbolt.org/z/o9rK41qdT

program--: After a brief discussion, this is actually more of a syntactical choice. We can define `~Bmi_Adapter() = 0`, and then implement `= default` in its source file to make it explicit that derived classes *should* implement a destructor, but there's no way to force that.

include/bmi/AbstractCLibBmiAdapter.hpp
hellkite500: I would consider keeping at least these three small, simple functions in the header and marked `inline`.  They are relatively simple wrapper functions which are used for each adapter and at scale (800k calls at initialization of conus) could save some overhead in intialization.  The library load and load symbol functions are more complex, and inlining them would likely have some similar initializaiton optimizations avoiding additional function calls in setting up the adapters.  I don't think these are called from many different places, so the resulting binary size and build times should be minimal.
We may want to at least profile some large init times and see what the overhead in intialization is from NOT inlining these.

program--: Reverted in [7fc98b0](https://github.com/NOAA-OWP/ngen/pull/750/commits/7fc98b0dd9aa81f6da5a40c160a9d619ad9a3038)

include/bmi/Bmi_Adapter.hpp
program--: Added in [e2678b6](https://github.com/NOAA-OWP/ngen/pull/750/commits/e2678b61e43188ce517202a43c0ce3a06321740f)

include/bmi/AbstractCLibBmiAdapter.hpp
program--: Reverted in [4f83059](https://github.com/NOAA-OWP/ngen/pull/750/commits/4f83059438c162f0cfe8fae81899a8c4e8891356)

include/bmi/AbstractCLibBmiAdapter.hpp
```diff
@@ -2,9 +2,8 @@
 #define NGEN_ABSTRACTCLIBBMIADAPTER_HPP
 
 #include <dlfcn.h>
```
PhilMiller, line 4: This `#include` can now be moved down to the .cpp file, since the calls to `dlopen` and `dlsym` are there.

include/bmi/Bmi_Py_Adapter.hpp
PhilMiller: I'd like to see those `final` specifiers actually applied.

include/bmi/AbstractCLibBmiAdapter.hpp
```diff
@@ -27,31 +26,17 @@ namespace models {
...
-                    bmi_lib_file(std::move(adapter.bmi_lib_file)),
-                    bmi_registration_function(adapter.bmi_registration_function),
-                    dyn_lib_handle(adapter.dyn_lib_handle)
-            {
-                // Have to make sure to do this after "moving" so the original does not close the dynamically loaded library handle
-                adapter.dyn_lib_handle = nullptr;
-            }
+                                   std::string registration_func, utils::StreamHandler output);
+
+            AbstractCLibBmiAdapter(AbstractCLibBmiAdapter &&adapter) noexcept = default;
```
PhilMiller, line 31: The body of the destructor means that the previous explicit body of the move constructor was necessary. A moved-from instance should not be calling `dlclose` on its `dyn_lib_handle`, since that would result in a doubled call.

Rather than restoring/fixing it, does the code build with this
```suggestion
            AbstractCLibBmiAdapter(AbstractCLibBmiAdapter &&adapter)= delete;
```
If so, I think that's preferable for making it enforceably clear that `AbstractCLibBmiAdapter` and its descendants are not value types, and should never be treated as such - consumers should always be constructing in place and assigning pointers, and doing otherwise is likely a mistake.

PhilMiller: I'll give that a quick test and report back.

program--: Ah, yeah I messed this one up. I don't remember why I made this change (I assume I thought that the original handle would be set to nullptr when moved, but that shouldn't work) 🤦🏾 thanks for catching that.

program--: At the very least, unit tests for C and Fortran pass when the move constructor is deleted.

PhilMiller: I've got everything passing locally. Hopefully CI agrees.

PhilMiller:
I'm trying to take a look at the ASan failure in the unit test. Needing to update some software locally to do that.

PhilMiller:
Ok, finally re-ran the tests locally with ASan, and didn't see the failure. Going to need some more investigation, I guess.

I may try backing out the very last commit, and see if that changes things

program--:
> Ok, finally re-ran the tests locally with ASan, and didn't see the failure. Going to need some more investigation, I guess.
> 
> I may try backing out the very last commit, and see if that changes things

I'm having the same experience. I don't get any ASan errors locally. As a note, it seems master is also experiencing the ASan issues (and the other PR I recently put up).

PhilMiller:
The ASan failures may be a transient thing due to weird configuration on Github's CI runners. In particular, ASan functionally requires that ASLR *not* be disabled, because there's certain ELF loading behavior in the Linux kernel that only happens with ASLR enabled. I pushed on that getting fixed about a decade ago, for other reasons, but never got anywhere.

The upshot is that we may see the test start passing if we re-run it again tomorrow.

PhilMiller (dismissed):

hellkite500 (approved):

Issue #746, opened 2024-02-27
Handle remapping BMI module output variable names


aaraney:
Configuring a multi bmi realization with two modules that share a common output variable _name_ can produces errors or unexpected results. Similarly, if the outer multi bmi module's `main_output_variable` _is_ the shared common output variable _name_, the last module listed in the `modules` section wins. Meaning the last module's output variable with the name associated with the multi bmi's `main_output_variable` section will be the output.

To explore this problem I created a simple BMI python module with a single `input` scalar variable and a single `output` scalar variable. The `output` is simply assigned to `input`. I put the module "back to back" in a multi bmi realization (see below realization configs) to show case these issues.

Failing to remapping either module's output name produces the following error:

```shell
std::runtime_error: Multi BMI cannot be created with module BmiA with output variable output because a previous module is using this output variable name/alias.
```

<details>
<summary>realization file</summary>

```json5
{
  "global": {
    "formulations": [
      {
        "name": "bmi_multi",
        "params": {
          "model_type_name": "",
          "allow_exceed_end_time": true,
          "main_output_variable": "output", // which module's `output` should this be? Currently it is always the last.
          "init_config": "/dev/null",
          "uses_forcing_file": false,
          "modules": [
            {
              "name": "bmi_python",
              "params": {
                "python_type": "model.BmiA",
                "model_type_name": "BmiA",
                "name": "python_bmi_1",
                "init_config": "green",
                "allow_exceed_end_time": true,
                "main_output_variable": "output",
                "uses_forcing_file": false,
                "variables_names_map": {
                  "input": "DLWRF_surface"
                }
              }
            },
            {
              "name": "bmi_python",
              "params": {
                "python_type": "model.BmiA",
                "model_type_name": "BmiA",
                "name": "python_bmi_2",
                "init_config": "blue",
                "allow_exceed_end_time": true,
                "main_output_variable": "output",
                "uses_forcing_file": false,
                "variables_names_map": {
                  "input": "DLWRF_surface"
                }
              }
            }
          ]
        }
      }
    ],
    "forcing": {
      "file_pattern": "cat-27.csv",
      "path": "./data/",
      "provider": "CsvPerFeature"
    }
  },
  "time": {
    "start_time": "2015-12-01 00:00:00",
    "end_time": "2015-12-01 23:00:00",
    "output_interval": 3600
  }
}
```

</details>

[...]

PR #753, opened 2024-03-06
docs: clarify usage of `variables_names_map`

aaraney:
Clarify usage of `variables_names_map` in multi-bmi context.

### DIFF
--- a/doc/BMI_MODELS.md
+++ b/doc/BMI_MODELS.md
@@ -122,7 +122,9 @@ There are some special BMI formulation config parameters which are required in c
 ### Optional Parameters
 * `variables_names_map`
   * can specify a mapping of model variable names (input or output) to supported standard names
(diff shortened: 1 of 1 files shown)


aaraney:
related to #746.

robertbartel:
@aaraney, I've opened #754 with some similar changes I'd been working on.  I think we should probably go with that one, as its a little more extensive, but if there is anything here that I'm missing there, let me know.

In particular, I'm leaning toward deliberately _not mentioning_ that output variable aliases are many-to-one, and potentially filing that as a bug for us to remove (albeit one with the lowest possible priority).  I'm assume you have it in there because you tested it and found that it works.   I don't remember intending that, and I don't think there are any scenarios where it does something that couldn't be accomplished by configuring the alias of an input.  And, to me, that seems like it might contribute to user confusion.  We can discuss that further and get other opinions, though.

aaraney:
closing in favor of #754.

PR #754, opened 2024-03-06
Improving documentation on BMI and BMI multi-formulation configuration

robertbartel:
Updating the _BMI_MODELS.md_ file with better detail on variable names mapping, behavior of `bmi_multi` formulation configurations, and usage (or omission) of default variable values.

Additions to docstrings for wrapper provider types to clarify details regarding when they are ready to provide, which has relevance with respect to the need for default values in the _OptionalWrappedDataProvider_.

### DIFF
--- a/doc/BMI_MODELS.md
+++ b/doc/BMI_MODELS.md
@@ -511,6 +511,10 @@ This introduces a special case for when there is no previous time step.  Not eve
 > [!NOTE]
 > Currently only `double` default values are supported.
 
+##### Example Look-Back Config
+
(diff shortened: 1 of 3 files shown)


doc/BMI_MODELS.md
program--, line 1: Would it be possible to add an example of a BMI multi-formulation that has a look-back input dependency? or rather just extend the default values example to include modules?

robertbartel: Sure, I've modified the example config and added a bit more description to hopefully better illustrate things.

program-- (approved):

Issue #628, opened 2023-09-05
Embedded Python interpreter on macOS puts system `site-packages` path before venv paths


mattw-nws:
In testing #626, it was discovered that on macOS, when a venv is active, the interpreter embedded in ngen reports `sys.path` directories including a system-level `site-packages` path like `*/Frameworks/Python.framework/Versions/3.*/lib/python3.*/site-packages` of the same version as the venv... this does not occur (or has not been observed) on Linux.

Initially it was thought that this might be a result of #540 (or possibly #592), but checking out the commit before #540 (c261af4) shows the same behavior. This can cause several problems, and it was observed that if there is a mismatched NumPy version in the system path from the on in a venv that this would trigger the mismatch exception added in #558.

Other (worse) things could happen because of this, for example if different BMI module versions were installed in a venv vs. the system/Python.framework site-packages paths, then an unexpected version could be transparently used (this could be one source of the problems described in #441). Generally installing things in the system Python path (or user site) is likely to cause problems like this, but is not uncommon--especially NumPy seems to end up there frequently, given search results. 

It seems as if this may be different behavior of pybind11 (or the embedded interpreter tooling of CPython, which is used by pybind11?) on macOS, but the cause is largely unknown. [Some interesting info on pybind11's embedding settings that may be relevant (or not)](https://github.com/pybind/pybind11/issues/4471).

## Current behavior

ngen built on macOS will include a path like the 4th line in the paths returned from `sys.path` in the embedded pybind11 Python interpreter:

```
Python Environment Info:
  VIRTUAL_ENV environment variable: /Volumes/Expendable/venv-trm-311
  Discovered venv: /Volumes/Expendable/venv-trm-311
  System paths:
    
    /usr/local/Cellar/python@3.11/3.11.1/Frameworks/Python.framework/Versions/3.11/lib/python311.zip
    /usr/local/Cellar/python@3.11/3.11.1/Frameworks/Python.framework/Versions/3.11/lib/python3.11
    /usr/local/Cellar/python@3.11/3.11.1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/lib-dynload
    /usr/local/Cellar/python@3.11/3.11.1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages
    /Volumes/Expendable/venv-trm-311/lib/python3.11
    /Volumes/Expendable/venv-trm-311/lib/python3.11/site-packages
```
...While in the same environment, `python` itself reports different paths:
```
[...]

PR #755, opened 2024-03-06
Fix for embedded Python interpreter issues on macOS

robertbartel:
Fixes issues described in #628.

Moving to a newer version of pybind11 that has [this issue](https://github.com/pybind/pybind11/issues/4471) fixed.  Also modifying the order in which items (e.g., virtual environment directories) are added to the interpreter  system path, but only on macOS.

These changes allow for successful and correct loading of numpy 1.26.3 from the virtual environment (as was used during build system generation), where previously 1.26.2 was being loaded from the system-level installed packages.

### DIFF
--- a/include/utilities/python/InterpreterUtil.hpp
+++ b/include/utilities/python/InterpreterUtil.hpp
@@ -146,7 +146,11 @@ namespace utils {
                 py::object requestedDirPath = Path(directoryPath);
                 if (py::bool_(requestedDirPath.attr("is_dir")())) {
                     if (std::find(sys_path_vector.begin(), sys_path_vector.end(), directoryPath) == sys_path_vector.end()) {
+#ifdef __APPLE__
+                        sys_path.attr("insert")(1, py::str(directoryPath));
+#else
                         sys_path.attr("insert")(sys_path_vector.size(), py::str(directoryPath));
+#endif
                     }
                 }
                 else {


PhilMiller:
Why the macOS change? The reasoning for the pybind11 update is quite clear, but that one's not. Especially why `insert` at `1` rather than `0`

PhilMiller:
Ping. What's the motivation for the macOS path order difference?

robertbartel:
@PhilMiller, basically the reason was that this is intended to fix an issue that seemingly only affects MacOS.  This is described in #628, and I experienced the issue myself.  I suppose I am assuming here that Linux environments truly aren't having this problem (I haven't actually confirmed this myself), though, even beyond Matts original description, I'd expect to have seen a lot more attention on the problem if Linux were also impacted.

Given that the issue is a MacOS-only problem, I was attempting to minimize the impact of the changes.  But we don't have to if consistency seems more important.

robertbartel:
As for why to insert at `1`,  it was because the path at index `0` was the current directory.  I wanted to move the virtual environment packages to a higher priority when searching for imports, but felt that searching in the current directory probably should still be done first.

robertbartel:
Also, to clarify something I might have earlier missed was being asked ... 

Moving to a new version of pybind11 was not enough to resolve the problem.  Both a newer pybind11 and modifications to adding directories to the interpreter's system path were necessary to get the embedded interpreter to import the correct numpy package.

PhilMiller (approved):
I'm a bit uneasy about the inconsistency, but this does empirically fix things for me on a Mac that I'm not experiencing on Linux, so it's maybe best to limit the change's impact for now.

PR #769, opened 2024-03-19
provide option to skip initializing and add general build flags to sub-module build

hellkite500:
Provide some flexibility in the submodule build action, so a caller can avoid initializing the module in case they want to replace the submodule code with custom code before the compile steps.

Also take a generic build flag input allowing custom cmake flags such as macro defines to be passed to to the cmake configure step.

This would replace #768, which would address noaa-owp/noaa-owp-modular#104 by allowing custom workflows to change to replace the code being built by the action.

Should also strongly consider adding a `workdir` input to allow a `cd ${{ inputs.workdir }}` before the configure/build steps.  This would allow some customization with the actions/checkout which allows checking out different repositories to new directories, but the build action would need to know which path to initiate the build steps under.

### DIFF
--- a/.github/actions/ngen-submod-build/action.yaml
+++ b/.github/actions/ngen-submod-build/action.yaml
@@ -17,6 +17,14 @@ inputs:
     mod-dir:
       required: true
       description: 'Path to the submodule with CMakeLists.txt to build'
+    initialize:
+      description: 'Initialize the ngen submodule, default'
+      required: false
+      default: 'true'
+    cmake-flags:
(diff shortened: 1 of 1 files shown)


aaraney (approved):
Looks good to me. Thanks, @hellkite500!

PhilMiller (approved):

madMatchstick (approved):
@hellkite500 - Thanks for spearheading this. 

For reference, this will require changes to NOM (thanks for the template/[pr#105](https://github.com/NOAA-OWP/noah-owp-modular/pull/105/files)), topmodel & pet workflows.

program-- (approved):

PR #770, opened 2024-03-20
reafctor: Move Formulation::get_output_line_for_timestep implementati…

PhilMiller:
…on from duplicates per-language to Bmi_Module_Formulation

## Additions

-

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

1.

- [ ] Linux


PhilMiller:
I recommend that this be reviewed commit-by-commit to make things much more comprehensible.

include/realizations/catchment/Bmi_Py_Formulation.hpp
```diff
@@ -95,7 +71,7 @@ namespace realization {
 
     protected:
 
-        std::shared_ptr<models::bmi::Bmi_Py_Adapter> construct_model(const geojson::PropertyMap &properties) override;
+        std::shared_ptr<models::bmi::Bmi_Adapter> construct_model(const geojson::PropertyMap &properties) override;
```
program--, line 74: I'm curious about this change. What's the reason to return the base class here instead of the derived class? Since it's covariant, it should be okay? (though, probably superfluous)

include/realizations/catchment/Bmi_Module_Formulation.hpp
```diff
@@ -44,18 +37,12 @@ namespace realization {
          * @param output_stream
          */
         Bmi_Module_Formulation(std::string id, std::shared_ptr<data_access::GenericDataProvider> forcing_provider, utils::StreamHandler output_stream)
-                : Bmi_Formulation(std::move(id), forcing_provider, output_stream) { };
+                : Bmi_Formulation(std::move(id), forcing_provider, output_stream) { }
 
-        virtual ~Bmi_Module_Formulation() {};
+        virtual ~Bmi_Module_Formulation() = default;
```
program--, line 42: Not an issue, but since we're modifying this line anyways, could we change the syntax to use `override` instead of `virtual` to show that the virtual destructor hierarchy starts above this class?

include/realizations/catchment/Bmi_Py_Formulation.hpp
PhilMiller: The signature has to exactly match the parent class, since this is a `virtual` member function.

Part of the change is to remove the templating of the base class that had this returning the derived concrete class.

include/realizations/catchment/Bmi_Module_Formulation.hpp
PhilMiller: I was kinda tempted to just leave off `virtual` too. I think technically, it's not an `override`, since the parent class destructor is still called as part of object instance destruction. Syntactically, it may be allowed, though.

PhilMiller: Ok, yeah, I see `override` is allowed as a special case. I'll make the change to add it.

include/realizations/catchment/Bmi_Py_Formulation.hpp
program--: Ah, I see. I thought covariant return types worked with smart pointers, but sadly they don't i.e. https://godbolt.org/z/nGE9rjEe6. Yeah nevermind then 🙂

PhilMiller: They don't, but keep in mind that's not why I made the change - as I said, the base class declaring this virtual function doesn't know the concrete type anymore (because I dropped the template), and the signature has to precisely match the base class's declaration of this function.

program-- (approved):
I think @hellkite500 wants to review first before merging, but this looks great to me!

hellkite500 (approved):

PR #797, opened 2024-04-19
Fix the few most pressing warnings reported by -Wall

PhilMiller:
## Changes

- Fix occurrences of `-Wpessimizing-move` reported by either GCC or AppleClang
- Fix a mis-use of bitwise `|` instead of logical `||`
- Fix range-for loop over `vector<string>` working on values rather than references
- Quiet some `-Wmaybe-uninitialized` by setting output-argument pointer variables to `nullptr`
- Fix catching polymorphic `boost::bad_get` by value

## Testing

1. Local testing with GCC 13
2. Local testing with AppleClang 15
3. CI with GCC 11
4. CI with AppleClang 14

## Notes

This gets us a large portion of the way to being able to build with `-Wall` for C++

- There are a lot of un/signed comparison warnings that I think we can reasonably suppress
- There are some constructor initializer list warnings about orders not matching declarations, that we should probably just fix, but aren't a big deal
- There are a bunch of unused variables and functions that should eventually be investigated and likely deleted
- The BMI C++ spec misses a virtual destructor, so we get warnings about that until https://github.com/csdms/bmi-cxx/issues/11 gets addressed, and we incorporate the fix in our modules

## Checklist

### DIFF
--- a/.github/actions/ngen-build/action.yaml
+++ b/.github/actions/ngen-build/action.yaml
@@ -182,7 +182,7 @@ runs:
           run: |
             export BOOST_ROOT="$(pwd)/boost_1_79_0"
             export CFLAGS="-fsanitize=address -O1 -g -fno-omit-frame-pointer -Werror"
-            export CXXFLAGS="-fsanitize=address -O1 -g -fno-omit-frame-pointer -pedantic-errors -Werror"
+            export CXXFLAGS="-fsanitize=address -O1 -g -fno-omit-frame-pointer -pedantic-errors -Werror -Wpessimizing-move -Wparentheses -Wrange-loop-construct"
             . .venv/bin/activate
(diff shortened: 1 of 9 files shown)


PhilMiller:
If we merge this, then the actual `ngen` code and its tests compiles with `-Werror -Wall -Wno-reorder -Wno-unused -Wno-sign-compare -Wno-catch-value` on GCC 13. I definitely think it's worth fixing the `-Wcatch-value` reports as well. The others are less definitively problematic.

PhilMiller:
The selection of flags added to CI is based on which ones are available for both GCC and AppleClang. Otherwise, the build just fails with unrecognized command line arguments.

program-- (approved):

hellkite500 (approved):

PR #801, opened 2024-04-20
re-throw exceptions as std::runtime_error with original error

hellkite500:
When getting time units from the base BMI class, exception context is lost from the subclasses which raise them.  For example, delegating to to the python adapter where the python module raises an exception, the pybind::error_already_set exception and message is lost, and we are left with just an uncaught std::exception with no context.  Catching the generic exception and copying the message into a runtime_error at least provides the error context provided by the original exception.

## Changes

- `get_time_convert_factor` now does a try/catch and throws runtime_error if any std::exception is caught

## Checklist

- [x] Linux
- [x] MacOs

### DIFF
--- a/src/bmi/Bmi_Adapter.cpp
+++ b/src/bmi/Bmi_Adapter.cpp
@@ -37,7 +37,19 @@ Bmi_Adapter::~Bmi_Adapter() = default;
 
 double Bmi_Adapter::get_time_convert_factor() {
     double value             = 1.0;
-    std::string input_units  = GetTimeUnits();
+    std::string input_units("-");
+    try{
+        input_units  = GetTimeUnits();
+    }
+    catch(std::exception &e){
+        //Re-throwing any exception as a runtime_error so we don't lose
+        //the error context/message.  We will lose the original exception type, though
+        //When a python exception is raised from the py adapter subclass, the
+        //pybind exception is lost and all we see is a generic "uncaught exception"
+        //with no context.  This way we at least get the error message wrapped in
+        //a runtime error.
+        throw std::runtime_error(e.what());
+    }
     std::string output_units = "s";
     return UnitsHelper::get_converted_value(input_units, value, output_units);
 }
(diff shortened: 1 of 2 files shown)


src/bmi/Bmi_Adapter.cpp
```diff
@@ -37,7 +37,19 @@ Bmi_Adapter::~Bmi_Adapter() = default;
...
+    try{
+        input_units  = GetTimeUnits();
+    }
+    catch(std::exception &e){
+        //Re-throwing any exception as a runtime_error so we don't loose
+        //the error context/message.  We will loose the original exception type, though
+        //When a python excpetion is raised from the py adapter subclass, the
+        //pybind exception is lost and all we see is a generic "uncaught exception"
+        //with no context.  This way we at least get the error message wrapped in
+        //a runtime error.
```
program--, line 50:
```suggestion
        //Re-throwing any exception as a runtime_error so we don't lose
        //the error context/message.  We will lose the original exception type, though
        //When a python exception is raised from the py adapter subclass, the
        //pybind exception is lost and all we see is a generic "uncaught exception"
        //with no context.  This way we at least get the error message wrapped in
        //a runtime error.
```

hellkite500: squashed into ef347aed6ec12b9d8cafdc83243d93c46da352a7

PhilMiller:
Were we actually getting exceptions thrown from `GetTimeUnits`? Were they actually related to that, or something coming from a first call to a BMI module post-`Initialize`?

hellkite500:
> Were we actually getting exceptions thrown from `GetTimeUnits`? Were they actually related to that, or something coming from a first call to a BMI module post-`Initialize`?

If a python module, for example, implemented `get_time_units` as
```python
    def get_time_units(self) -> str:
        raise NotImplementedError
```
Then this function would error with just
```sh
libc++abi: terminating due to uncaught exception of type std::exception: std::exception
```
Other time functions produced the same.

I did a quick test on other BMI functionality used via ngen and didn't see the same uncaught exception message.  With the proposed changes, the above sitaution produces
```sh
libc++abi: terminating due to uncaught exception of type std::runtime_error: NotImplementedError: <EMPTY MESSAGE>

At:
  <path/to/module.py(<line>): get_time_units
```

PhilMiller:
That change in error message is definitely a big improvement.

In the BMI C adapter, we check every return, and throw on failures. Should each language BMI adapter be responsible for doing similarly? I.e. should `Bmi_Py_Adapter` catch whatever potentially gets thrown from any method, and wrap it up in a type that our higher-level language-independent code will recognize?

donaldwj (approved):

PR #802, opened 2024-04-20
Drop fatal error else case when trying to update submodules

hellkite500:
The error condition this was originally trapping doesn't seem relevant, and results in a fatal build error when the submodule build option is disabled, even thought it shouldn't be.

## Changes

- don't error build if ngen submodules are not requested

## Checklist

## Testing checklist

1. Build tested locally

- [x] Linux
- [x] MacOS


PhilMiller:
The change looks good to me, but I'm a bit confused about the code that's being removed. The error message seems incongruous with the logic

program-- (approved):
If I remember correctly, the `else()` was there for a different reason (that I can't remember), that became unnecessary after many evolutions of changes (which is why, I think, the logic doesn't make much sense at this point).

In any case, this looks good to me as well.

PR #804, opened 2024-04-23
ci: fix macos brew include paths. differ based on arch

aaraney:
Homebrew puts headers files in `/usr/local/include` on x86 and `/opt/homebrew/include` on apple arm. Tests are failing in multiple repos b.c. they cannot find `udunits2.h` when the runner is apple arm. This fixes that issue and generalizes how the include path is determined.


aaraney:
See https://github.com/NOAA-OWP/LGAR-C/actions/runs/8804814183/job/24166001315#step:13:396 as an example failure.

PhilMiller (approved):
Please quote the text of the failure. The job logs expire pretty quickly. 

Change looks good

PR #814, opened 2024-05-03
Replace old macros with consistent NGEN_WITH_FOO

PhilMiller:
Replace all NGen-specific build condition macros that weren't in the current standard form in the codebase with the standard form.

Resolves #813 

## Testing

1. Local testing with various flags enabled/disabled
2. CI

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows project standards (link if applicable)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

PR #815, opened 2024-05-05
Clean up code for Formulation et al in advance of Coastal introduction

PhilMiller:

## Additions

-

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

1.

- [ ] Linux


include/realizations/catchment/Bmi_C_Formulation.hpp
```diff
@@ -18,42 +18,6 @@ namespace realization {
...
-         *
-         * The BMI spec requires for variable values to be passed to/from models via as arrays.  This function
-         * essentially  treats the variable array reference as if it were just a raw pointer and returns the `0`-th
-         * array value.
-         *
-         * @param t_index The index of the time step for which to run model calculations.
-         * @param d_delta_s The duration, in seconds, of the time step for which to run model calculations.
-         * @return The total discharge of the model for the given time step.
-         */
-        double get_response(time_step_t t_index, time_step_t t_delta) override;
```
donaldwj, line 55: When moving get_var_value_as_double() would it be a good time to replace it with a more correct set of functions or possibly a templated function. Not all value are double or can be stored in one.

include/realizations/catchment/Bmi_C_Formulation.hpp
```diff
@@ -135,6 +87,7 @@ namespace realization {
          * @return
          */
         double get_var_value_as_double(const int& index, const std::string& var_name) override;
+        using Bmi_Formulation::get_var_value_as_double;
```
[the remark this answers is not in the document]
PhilMiller: Yes, it is (was?) needed, because otherwise the single `override` actually hides the overload set. I was a bit surprised. There may be more going on, or it may have been transient the overloads defined at different layers of the hierarchy as I had changes in progress.

src/realizations/catchment/Bmi_Fortran_Formulation.cpp
```diff
@@ -94,77 +90,4 @@ double Bmi_Fortran_Formulation::get_var_value_as_double(const int &index, const
...
-        else
-            get_bmi_model()->UpdateUntil(model_initial_time + t_delta_model_units);
-        // TODO: again, consider whether we should store any historic response, ts_delta, or other var values
-        next_time_step_index++;
-    }
-    return get_var_value_as_double( get_bmi_main_output_var());
-}
-
-#endif // NGEN_BMI_FORTRAN_ACTIVE
+#endif // NGEN_WITH_BMI_FORTRAN
```
PhilMiller, line 93: Oops, looks like a rebase mistake led me to mis-match the `#if`. I'll clear that up before this gets merged

include/realizations/catchment/Bmi_Formulation.hpp
donaldwj, line 1: Why remove virtual from this? It does not need to be pure virtual but some formulation class may need to use a different implementation, for example caching variables that are expensive to access as they required a gather operation.

include/realizations/catchment/Bmi_C_Formulation.hpp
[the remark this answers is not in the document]
PhilMiller: For the sake of this cleanup, it seems potentially sensible to eliminate the overload, and just have the handful of call sites that want to assume a scalar value pass an index 0 argument. On the flip side, having them call this overload would let us add a check that the value being requested really is a scalar and not part of a larger array.

donaldwj (commented):
I dont see any major issues. I am curious on some of the instances of removing virtual from functions that are now implemented but those are not an actual problem. I do think fully implementing get_value would make sense here.

src/realizations/catchment/Bmi_Fortran_Formulation.cpp
PhilMiller: I think I'll let #814 deal with that, and leave it alone here to avoid merge conflicts.

PhilMiller:
I think this is mostly in shape to clear the path for Coastal needs. I'm still uncertain about `create_formulation`, but that could be addressed in a subsequent PR.

donaldwj (approved):

PR #833, opened 2024-06-11
ci: fix troute build

aaraney:
Fix `t-route` CI builds after changes introduced in https://github.com/NOAA-OWP/t-route/pull/780. https://github.com/NOAA-OWP/t-route/pull/780 at its core addressed the need for the python `wheel` package as a build time dependency. `wheel` was not installed in CI before attempting to build `t-route`. This addresses that.


aaraney:
Testing this with the pinned `pip` dep now just to make sure this solves the issue. Once the tests pass, I will remove the pinning of the `pip` version in CI and include that as part of this PR.

hellkite500 (approved):

PR #839, opened 2024-06-25
Forcings Engine Data Provider Base Interface

program--:
> [!IMPORTANT]
> **This PR replaces #720.**

This PR adds the **`ForcingsEngineDataProvider<DataType, SelectionType>`** class. This class is intended to serve as the base class for Lumped, Gridded, and Unstructured forcings engine data providers. The base class handles timing and updating the underlying BMI instance, so derived classes need only implement the value querying requirements.

Internally, this PR also adds the **`detail::ForcingsEngineStorage`** class, which composes a map of initialization file names to Python BMI Adapter shared pointers. The design of this is to allow the Forcings Engine data providers to store a mask of some sort to simplify querying from the forcings engine, i.e. for a Gridded instance, the mask may be a bounding box. This need is due to the implementation of the forcings engine, and should simplify the external requirements for a NGen run using the Python Forcings Engine.

Additionally, this PR includes some minor changes to the `forcings` library. Namely, the `GenericDataProvider` class is refactored into a type alias, and the `NullForcingProvider` implementation is split into header/source files rather than being header-only. The latter change is to ensure the `forcings` CMake target is always build-able.

## Additions

- `ForcingsEngineDataProvider`: templated base class.
- `detail::ForcingsEngineStorage`: shared storage between data providers.
- `detail::assert_forcings_engine_requirements`: helper for ensuring runtime dependencies are available.
- `detail::parse_time`: helper for parsing timestamp strings.
-  `constexpr` constants for default time format string and forcings engine python information.

## Changes

- `NullForcingProvider`: split implementation into header/source file.
- `GenericDataProvider`: converted into type alias. 

## Checklist

### DIFF
--- a/include/forcing/ForcingsEngineDataProvider.hpp
+++ b/include/forcing/ForcingsEngineDataProvider.hpp
@@ -1,5 +1,9 @@
 #pragma once
 
+#include <NGenConfig.h>
+
+#if NGEN_WITH_PYTHON
+
 #include <cmath>
 #include <chrono>
 #include <memory>
@@ -211,3 +215,5 @@ struct ForcingsEngineDataProvider
 };
 
 } // namespace data_access
+
+#endif // NGEN_WITH_PYTHON
--- a/include/forcing/ForcingsEngineDataProvider.hpp
+++ b/include/forcing/ForcingsEngineDataProvider.hpp
@@ -60,7 +60,7 @@ struct ForcingsEngineStorage {
     }
 
     //! Associate a Forcings Engine instance to a file path.
-    //! @param key Initiailiation file path for Forcings Engine instance.
+    //! @param key Initialization file path for Forcings Engine instance.
     //! @param value Shared pointer to a Forcings Engine BMI instance.
     void set(const key_type& key, value_type value)
     {
--- a/include/forcing/ForcingsEngineDataProvider.hpp
+++ b/include/forcing/ForcingsEngineDataProvider.hpp
@@ -178,7 +178,6 @@ struct ForcingsEngineDataProvider
                 "ForcingsEngine",
                 init,
                 forcings_engine_python_classpath,
-                /*allow_exceed_end=*/true,
                 /*has_fixed_time_step=*/true
             );
 
--- a/include/forcing/ForcingsEngineDataProvider.hpp
+++ b/include/forcing/ForcingsEngineDataProvider.hpp
@@ -144,11 +144,31 @@ struct ForcingsEngineDataProvider
     )
       : time_begin_(std::chrono::seconds{time_begin_seconds})
       , time_end_(std::chrono::seconds{time_end_seconds})
-      , bmi_(storage_type::instances.get(init))
-      , time_step_(std::chrono::seconds{static_cast<int64_t>(bmi_->GetTimeStep())})
-      , time_current_index_(std::chrono::seconds{static_cast<int64_t>(bmi_->GetCurrentTime())} / time_step_)
-      , var_output_names_(bmi_->GetOutputVarNames())
-    {}
+    {
+        // Get a forcings engine instance if it exists for this initialization file
+        auto bmi_ = storage_type::instances.get(init);
+
+        // If it doesn't exist, create it and assign it to the storage map
+        if (bmi_ == nullptr) {
+            // Outside of this branch, this->bmi_ != nullptr after this
+            bmi_ = std::make_shared<models::bmi::Bmi_Py_Adapter>(
+                "ForcingsEngine",
+                init,
+                forcings_engine_python_classpath,
+                /*allow_exceed_end=*/true,
+                /*has_fixed_time_step=*/true,
+                utils::getStdOut()
+            );
+
+            storage_type::instances.set(init, bmi_);
+        }
+
+        // Now, initialize the BMI dependent instance members
+        // NOTE: using std::lround instead of static_cast will prevent potential UB
+        time_step_ = std::chrono::seconds{std::lround(bmi_->GetTimeStep())};
+        time_current_index_ = (std::chrono::seconds{std::lround(bmi_->GetCurrentTime())} / time_step_); // resolves to long
+        var_output_names_ = bmi_->GetOutputVarNames();
+    }
 
     
     //! Update the Forcings Engine instance to the next timestep.
@@ -165,12 +185,11 @@ struct ForcingsEngineDataProvider
     //!                  since 3600 * 4 = 14400 but 14400 < 14401.
     void next(double time)
     {
-      const auto start = bmi_->GetCurrentTime();
-      bmi_->UpdateUntil(time);
-      const auto end = bmi_->GetCurrentTime();
-      time_current_index_ += static_cast<int64_t>(
-        (end - start) / bmi_->GetTimeStep()
-      );
+        const auto start = bmi_->GetCurrentTime();
+        bmi_->UpdateUntil(time);
+        const auto end = bmi_->GetCurrentTime();
+        const auto diff = end - start;
+        time_current_index_ += std::chrono::seconds{std::lround(diff)} / time_step_;
     }
 
     //! Forcings Engine instance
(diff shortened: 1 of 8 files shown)


include/forcing/ForcingsEngineDataProvider.hpp
```diff
@@ -0,0 +1,258 @@
...
+    bool
+  > = true
+>
+std::unique_ptr<typename Tp::base_type> make_forcings_engine(
+  const std::string& init,
+  const std::string& time_begin,
+  const std::string& time_end,
+  const std::string& time_fmt = default_time_format,
+  Args&&...          args
+)
```
program--, line 223: It probably makes more sense to just return `std::unique_ptr<Tp>`, and let polymorphism -- if needed for some reason -- be handled on assignment by consumers of this function.

src/forcing/ForcingsEngineDataProvider.cpp
```diff
@@ -0,0 +1,47 @@
...
+// Initialize instance storage
+ForcingsEngineStorage ForcingsEngineStorage::instances{};
+
+time_t parse_time(const std::string& time, const std::string& fmt)
+{
+    std::tm tm_ = {};
+    std::stringstream tmstr{time};
+    tmstr >> std::get_time(&tm_, fmt.c_str());
+
+    // Note: `timegm` is available for Linux and BSD (aka macOS) via time.h, but not Windows.
```
PhilMiller, line 18: Is this thus missing `#include <time.h>`?

program--: I think `<iomanip>` transitively includes `<ctime>`, but it's probably better to explicitly include that. I'll add this in.

include/forcing/ForcingsEngineDataProvider.hpp
program--: Now that I'm taking the 100th look at this, but with a clearer head, I'm pretty sure this isn't needed.

- The instance getting/construction logic can be moved to the protected constructor
- The assertion is planned to move outside of this to a single call point
- All that's left is `std::make_unique<Tp>`, which doesn't require a convenience function, and on its own removes the need for any form of type checking.

include/forcing/ForcingsEngineDataProvider.hpp
```diff
@@ -0,0 +1,211 @@
...
+//! Check that requirements for running the forcings engine
+//! are available at runtime. If requirements are not available,
+//! then this function throws.
+void assert_forcings_engine_requirements();
+
+//! Storage for Forcings Engine-specific BMI instances.
+struct ForcingsEngineStorage {
+    using key_type   = std::string;
+    using bmi_type   = models::bmi::Bmi_Py_Adapter;
+    using value_type = std::shared_ptr<bmi_type>;
```
PhilMiller, line 39: It would be good to comment these types with descriptions of the particular contents of their instances -- i.e. config file path, and corresponding instantiations of the above-named Python class

include/forcing/ForcingsEngineDataProvider.hpp
```diff
@@ -0,0 +1,211 @@
...
+    ForcingsEngineDataProvider(
+        std::string init,
+        std::size_t time_begin_seconds,
+        std::size_t time_end_seconds
+    )
+      : time_begin_(std::chrono::seconds{time_begin_seconds})
+      , time_end_(std::chrono::seconds{time_end_seconds})
+    {
+        // Get a forcings engine instance if it exists for this initialization file
+        auto bmi_ = storage_type::instances.get(init);
```
PhilMiller, line 149: As written, this shadows the member variable `bmi_`, and so doesn't actually initialize it. I think you wanted simply
```suggestion
        bmi_ = storage_type::instances.get(init);
```

include/forcing/ForcingsEngineDataProvider.hpp
```diff
@@ -0,0 +1,211 @@
...
+                utils::getStdOut()
+            );
+
+            storage_type::instances.set(init, bmi_);
+        }
+
+        // Now, initialize the BMI dependent instance members
+        // NOTE: using std::lround instead of static_cast will prevent potential UB
+        time_step_ = std::chrono::seconds{std::lround(bmi_->GetTimeStep())};
+        time_current_index_ = (std::chrono::seconds{std::lround(bmi_->GetCurrentTime())} / time_step_); // resolves to long
```
PhilMiller, line 169: `time_current_index_` is initialized and updated, but it's `private` and I never see it being used within this class definition. Is it just a relic of some sort?

PhilMiller:
The shadowed variable `bmi_` is a substantial concern. The other comments are inconsequential.

include/forcing/ForcingsEngineDataProvider.hpp
program--: Good catch, yeah I missed that when I removed the factory function. Thanks!

include/forcing/ForcingsEngineDataProvider.hpp
program--: Yeah I think this was a relic, since I didn't see any reference to it in the gridded or lumped implementations in a previous PR. Removed it, thanks!

include/forcing/ForcingsEngineDataProvider.hpp
```diff
@@ -0,0 +1,211 @@
...
+        auto pos = data_.find(key);
+        if (pos == data_.end()) {
+          return nullptr;
+        }
+
+        return pos->second;
+    }
+
+    //! Associate a Forcings Engine instance to a file path.
+    //! @param key Initiailiation file path for Forcings Engine instance.
```
PhilMiller, line 63:
```suggestion
    //! @param key Initialization file path for Forcings Engine instance.
```
Typo there

PhilMiller (dismissed):
Maybe good to let others have a look, or see what the derived classes look like, but I think this is all good.

include/forcing/ForcingsEngineDataProvider.hpp
```diff
@@ -0,0 +1,212 @@
...
+        // If it doesn't exist, create it and assign it to the storage map
+        if (bmi_ == nullptr) {
+            // Outside of this branch, this->bmi_ != nullptr after this
+            bmi_ = std::make_shared<models::bmi::Bmi_Py_Adapter>(
+                "ForcingsEngine",
+                init,
+                forcings_engine_python_classpath,
+                /*allow_exceed_end=*/true,
+                /*has_fixed_time_step=*/true,
+                utils::getStdOut()
```
PhilMiller, line 180:
```suggestion
                /*has_fixed_time_step=*/true
```

PhilMiller: To match the removal of the `StreamHandler` parameters

include/forcing/ForcingsEngineDataProvider.hpp
```diff
@@ -0,0 +1,212 @@
...
+
+  protected:
+    using storage_type = detail::ForcingsEngineStorage;
+
+    //! Forcings Engine Data Provider Constructor
+    //!
+    //! @note Derived implementations should delegate to this constructor
+    //!       to acquire a shared forcings engine instance.
+    ForcingsEngineDataProvider(
+        std::string init,
```
PhilMiller, line 161:
```suggestion
        std::string const& init,
```

include/forcing/ForcingsEngineDataProvider.hpp
```diff
@@ -0,0 +1,212 @@
+#pragma once
+
+#include <memory>
+#include <unordered_map>
+#include <chrono>
```
PhilMiller, line 5:
```suggestion
#include <chrono>
#include <string>
```

PhilMiller:
The main header here can only be included with `NGEN_WITH_PYTHON` enabled, but I'm not sure that's actually something we need to guard against in the header itself. Implementations and clients are going to need to guard against it anyway. I guess it would save clients conditionalizing both the `#include` and the usage site.

include/forcing/ForcingsEngineDataProvider.hpp
```diff
@@ -0,0 +1,214 @@
...
+        bmi_ = storage_type::instances.get(init);
+
+        // If it doesn't exist, create it and assign it to the storage map
+        if (bmi_ == nullptr) {
+            // Outside of this branch, this->bmi_ != nullptr after this
+            bmi_ = std::make_shared<models::bmi::Bmi_Py_Adapter>(
+                "ForcingsEngine",
+                init,
+                forcings_engine_python_classpath,
+                /*allow_exceed_end=*/true,
```
PhilMiller, line 181: #842, if merged first, will require deleting this line. I'm happy to hold off and make the update to this there, though.

PhilMiller: OK, that's merged. I'll clear this line.

src/forcing/ForcingsEngineDataProvider.cpp
```diff
@@ -0,0 +1,48 @@
+#include <forcing/ForcingsEngineDataProvider.hpp>
```
PhilMiller, line 1: I think this file should be guarded by `NGEN_WITH_PYTHON`

include/forcing/ForcingsEngineDataProvider.hpp
```diff
@@ -0,0 +1,213 @@
...
+//! Default time format used for parsing timestamp strings
+static constexpr auto default_time_format = "%Y-%m-%d %H:%M:%S";
+
+namespace detail {
+
+//! Parse time string from format.
+//! Utility function for ForcingsEngineLumpedDataProvider constructor.
+time_t parse_time(const std::string& time, const std::string& fmt);
+
+//! Check that requirements for running the forcings engine
```
donaldwj, line 34: Since this is probably a fatal error only this function is fine I would still want a check form of the function but it is not important

src/forcing/ForcingsEngineDataProvider.cpp
program--: Just to clarify, you mean `ForcingsEngineDataProvider.hpp` should have a `NGEN_WITH_PYTHON` guard inside of it?

include/forcing/ForcingsEngineDataProvider.hpp
```diff
@@ -0,0 +1,213 @@
...
+
+    //! Output variable names
+    std::vector<std::string> var_output_names_{};
+
+  private:
+    //! Initialization config file path
+    std::string init_;
+
+    //! Calendar time for simulation beginning
+    clock_type::time_point time_begin_{};
```
donaldwj, line 204: We probably need an official style rule on member data names (or names in general), we have at least 4 styles used between various files

src/forcing/ForcingsEngineDataProvider.cpp
donaldwj: Anything that requires python should be

donaldwj (approved):
I dont see any problems

donaldwj (approved):

PR #843, opened 2024-06-26
Update Formulation_Manager.hpp to use boost regex

JoshCu:
# Update the formulation manager to use boost::regex instead of std.

When matching complex regex for finding forcing files, std::regex becomes a significant portion of Ngen init. Especially with large numbers of catchments per core.

For an extreme example, when running serially for wb-479197 and it's upstreams ~6500 catchments. NGen::init takes ~153s, if I use the line "forcing": {"file_pattern": ".\*{{id}}.\*.csv" in my realization.
removing those wildcards reduces the time to 69 seconds.

For use cases where ngen is being run over a large area, this init time can become a non-insignificant portion of the total runtime, in the serial example, it took 153 seconds to init, 43s to run the models, 19s to route for a 24h simulation.

## Changes

- std::regex -> boost::regex

## Testing

This was all tested on ngiab images using ngen [f91e2ea](https://github.com/NOAA-OWP/ngen/commit/f91e2ea)
The run package was generated using the [ngiab_preprocessor](https://github.com/CIROH-UA/NGIAB_data_preprocess)
24h run of ~6500 catchments, sloth + cfe + noaa-owp-modular + troute, subset from wb-479197 on hydrofabric v20.1

<details>
  <summary>Hardware used</summary>
Model name:            Intel(R) Xeon(R) CPU E5-2697 v3 @ 2.60GHz
    Thread(s) per core:  2
    Core(s) per socket:  14
    Socket(s):           2
    CPU max MHz:         3600.0000
    CPU min MHz:         1200.0000
RAM 2100mhz DDR4
Storage ~500mbs sata ssd
</details>

<details>
 <summary> Realization.json </summary>

```json  

{
    "global": {
        "formulations": [
            {
                "name": "bmi_multi",
                "params": {
                    "name": "bmi_multi",
                    "model_type_name": "bmi_multi",
                    "main_output_variable": "Q_OUT",
                    "forcing_file": "",
                    "init_config": "",
                    "allow_exceed_end_time": true,
                    "modules": [
                        {
                            "name": "bmi_c++",
                            "params": {
                                "name": "bmi_c++",
                                "model_type_name": "SLOTH",
                                "main_output_variable": "z",
                                "init_config": "/dev/null",
                                "allow_exceed_end_time": true,
                                "fixed_time_step": false,
                                "uses_forcing_file": false,
                                "model_params": {
                                    "sloth_ice_fraction_schaake(1,double,m,node)": 0.0,
                                    "sloth_ice_fraction_xinanjiang(1,double,1,node)": 0.0,
                                    "sloth_soil_moisture_profile(1,double,1,node)": 0.0
                                },
                                "library_file": "/dmod/shared_libs/libslothmodel.so",
                                "registration_function": "none"
                            }
                        },
                        {
                            "name": "bmi_fortran",
                            "params": {
                                "name": "bmi_fortran",
                                "model_type_name": "NoahOWP",
                                "library_file": "/dmod/shared_libs/libsurfacebmi.so",
                                "forcing_file": "",
                                "init_config": "./config/cat_config/NOAH-OWP-M/{{id}}.input",
                                "allow_exceed_end_time": true,
                                "main_output_variable": "QINSUR",
                                "variables_names_map": {
                                    "PRCPNONC": "precip_rate",
                                    "Q2": "SPFH_2maboveground",
                                    "SFCTMP": "TMP_2maboveground",
                                    "UU": "UGRD_10maboveground",
                                    "VV": "VGRD_10maboveground",
                                    "LWDN": "DLWRF_surface",
                                    "SOLDN": "DSWRF_surface",
                                    "SFCPRS": "PRES_surface"
                                },
                                "uses_forcing_file": false
                            }
                        },
                        {
                            "name": "bmi_c",
                            "params": {
                                "name": "bmi_c",
                                "model_type_name": "CFE",
                                "main_output_variable": "Q_OUT",
                                "init_config": "./config/cat_config/CFE/{{id}}.ini",
                                "allow_exceed_end_time": true,
                                "fixed_time_step": false,
                                "uses_forcing_file": false,
                                "registration_function": "register_bmi_cfe",
                                "variables_names_map": {
                                    "water_potential_evaporation_flux": "EVAPOTRANS",
                                    "atmosphere_water__liquid_equivalent_precipitation_rate": "QINSUR",
                                    "ice_fraction_schaake": "sloth_ice_fraction_schaake",
                                    "ice_fraction_xinanjiang": "sloth_ice_fraction_xinanjiang",
                                    "soil_moisture_profile": "sloth_soil_moisture_profile"
                                },
                                "library_file": "/dmod/shared_libs/libcfebmi.so.1.0.0"
                            }
                        }
                    ],
                    "uses_forcing_file": false
                }
            }
        ],
        "forcing": {
            "file_pattern": "{{id}}.csv",
            "path": "./forcings/by_catchment/",
            "provider": "CsvPerFeature"
        }
    },
    "time": {
        "start_time": "2010-01-01 00:00:00",
        "end_time": "2010-01-02 00:00:00",
        "output_interval": 3600,
        "nts": 288.0
    },
    "routing": {
        "t_route_config_file_with_path": "/ngen/ngen/data/config/ngen.yaml"
    },
    "output_root": "/ngen/ngen/data/outputs/ngen"
}

```

</details>

`boost+exact` first checks if file_pattern matched the file exactly, then runs regex if it does not match

`exact path` is where the "path" formulation variable accepts the `{{id}}` placeholder allowing for file_pattern to be omitted and no regex being run. 

### Serial
| library | .\*{{id}}.\*.cat <br> | {{id}}.cat |
| ------- | --------------------- | ---------- |
| std     | 153s                  | 69s        |
| boost   | 62s                   | 55s        |
| re2     | 37s                   | 36s        |
| exact path | NA | 27s |

### 10 mpi ranks
| library | {{id}}.cat |
| ------- | ---------- |
| std     | 8.6s      |
| boost   |  7.4s        |
| re2     |  5.2s        |
| exact path | 4.2s  |

### 56 mpi ranks
| library | {{id}}.cat |
| ------- | ---------- |
| std     | 4.8s        |
| boost   |  4.5s        |
| re2     |  3.9s        |
| exact path |  3.2s |  

## Screenshots

[interactive version of the graph here](https://claude.site/artifacts/7ab67fc0-f7f1-4e48-99e9-ba225bb37014)
image
<details>
<summary>std, boost, re2, on 10 cores</summary>   

image
</details>

<details>
<summary>perf Flamegraph of unmodified fomulation manager running serially with double wildcards</summary>   

kernel
</details>

## Notes
- re2 is faster, but I don't know if it's worth adding the dependency when we're already using boost
- If it's worth it, I can open a PR for a version using re2
- I'm not sure how common these wildcarded regex use cases are, or how many people would be running serially for large numbers of catchments
- read.cpp in the geopackage code also uses std::regex, but it's called so infrequently that the performance benefit is negligible. I left it unchanged to reduce code modification, although I'm unsure if that's the correct decision.
- changing {{id}}.cat to {{id}}\\.cat reduces all times further, but I noticed the lack of escapement after I'd finished all my testing
- I can upload the various docker images used to test to dockerhub if needed

## Future work
I wanted to keep changes minimal because I'm not too familiar with the rest of ngen, but would it be worth modifying the formulation manager further to optionally disable regex completely?
For my use-case, the forcing files are just named cat-1234.csv and there is one per catchment so I don't need to use regex at all after the formulation manager replaces the {{id}} placeholder.

### DIFF
--- a/include/realizations/catchment/Formulation_Manager.hpp
+++ b/include/realizations/catchment/Formulation_Manager.hpp
@@ -463,8 +463,7 @@ namespace realization {
                 if (directory != nullptr) {
                     bool match;
                     while ((entry = readdir(directory))) {
-                        match = boost::regex_match(entry->d_name, pattern);
-                        if( match ) {
+                        if (filepattern == entry->d_name || boost::regex_match(entry->d_name, pattern)) {
                             // If the entry is a regular file or symlink AND the name matches the pattern, 
                             //    we can consider this ready to be interpretted as valid forcing data (even if it isn't)
                             #ifdef _DIRENT_HAVE_D_TYPE
--- a/include/realizations/catchment/Formulation_Manager.hpp
+++ b/include/realizations/catchment/Formulation_Manager.hpp
@@ -9,8 +9,8 @@
 #include <functional>
 #include <dirent.h>
 #include <sys/stat.h>
-#include <regex>
 
+#include <boost/regex.hpp>
 #include <boost/property_tree/ptree.hpp>
 #include <boost/property_tree/json_parser.hpp>
 #include <FeatureBuilder.hpp>
@@ -410,7 +410,7 @@ namespace realization {
                 }
 
                 // Create a regular expression used to identify proper file names
-                std::regex pattern(filepattern);
+                boost::regex pattern(filepattern);
 
                 // A stream providing the functions necessary for evaluating a directory:
                 //    https://www.gnu.org/software/libc/manual/html_node/Opening-a-Directory.html#Opening-a-Directory
@@ -463,7 +463,7 @@ namespace realization {
                 if (directory != nullptr) {
                     bool match;
                     while ((entry = readdir(directory))) {
-                        match = std::regex_match(entry->d_name, pattern);
+                        match = boost::regex_match(entry->d_name, pattern);
                         if( match ) {
                             // If the entry is a regular file or symlink AND the name matches the pattern, 
                             //    we can consider this ready to be interpretted as valid forcing data (even if it isn't)


PhilMiller:
Hi Josh, and things for noticing the issue, looking into it, and proposing this fix.

We'll need to check whether `boost::regex` needs to be built, installed, and linked against, or is effectively 'header-only'. Everything else we use from Boost is in the latter category right now, so there may be some resistance to changing that. This sort of performance improvement may be worth it, though.

We'd definitely like to bring down excessive run times like you've observed. Would you be open to making some other changes and testing them, since you've already got this teed up?

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -463,7 +463,7 @@ namespace realization {
                 if (directory != nullptr) {
                     bool match;
                     while ((entry = readdir(directory))) {
-                        match = std::regex_match(entry->d_name, pattern);
+                        match = boost::regex_match(entry->d_name, pattern);
                         if( match ) {
```
PhilMiller, line 467:
```suggestion
                        if (filepattern == entry->d_name || boost::regex_match(entry->d_name, pattern)) {
```

JoshCu:
> Hi Josh, and thanks for noticing the issue, looking into it, and proposing this fix.
> 
> We'll need to check whether `boost::regex` needs to be built, installed, and linked against, or is effectively 'header-only'. Everything else we use from Boost is in the latter category right now, so there may be some resistance to changing that. This sort of performance improvement may be worth it, though.
> 
> We'd definitely like to bring down excessive run times like you've observed. Would you be open to making some other changes and testing them, since you've already got this teed up?

Yeah of course! Happy to do any additional testing, just let me know what I should try.

As for boost regex being header only, I don't think it needs to be built? the dockerfile I'm using just downloads boost.bzip, unzips it, then adds the folder to path before building ngen. I don't have to make install it or anything if that answers the question?

my docker patch is just two sed commands with no additional changes to how ngen's built 
```dockerfile
RUN sed -i 's|#include <regex>|#include <boost/regex.hpp>|g' include/realizations/catchment/Formulation_Manager.hpp
RUN sed -i 's/std::regex/boost::regex/g' include/realizations/catchment/Formulation_Manager.hpp
```
the dockerfile is [this one](https://github.com/JoshCu/NGIAB-CloudInfra/blob/build_by_release/docker/Dockerfile) if it's of any use/interest :)

include/realizations/catchment/Formulation_Manager.hpp
JoshCu: I've tested this and it works, it's also inspired me to see if there's a way to filter the directory with regex, then loop over the matches, rather than having to check each file one at a time. It looks possible with boost::filesystem, but that might be a bit too cumbersome for this

JoshCu:
@PhilMiller What about changing `path` property to also accept {{id}}?
Changing this
```c++ 
if(forcing_prop_map.count("path") != 0){
    path = forcing_prop_map.at("path").as_string();
}
```
to
```c++
if(forcing_prop_map.count("path") != 0){
    path = forcing_prop_map.at("path").as_string();
    int id_index = path.find("{{id}}");
    if (id_index != std::string::npos) {
        path = path.replace(id_index, sizeof("{{id}}") - 1, identifier);
    }
}
```
means that I can put the {{id}} in the path, then remove file_pattern from my realization so that it hits the early return before any regex is used.
Testing it gives me 27s in serial and 3.2s in parallel 56 cores

PhilMiller:
The idea for `{{id}}` substitution in path sounds good to me. I don't know job setups too well, though, so I'll ask Bobby and/or Austin to comment on that.

Meanwhile, I added lines in the PR description's timing tables for "boost+exact" describing the code as amended to check against literal `filepattern`. Could you please put up numbers for that the current PR code?

PhilMiller:
If you're feeling diligent, `std+exact` might be helpful too, if we conclude that we don't want to add the dependency on Boost regex.

aaraney:
@PhilMiller, thanks for looping me in. I don't foresee this being an issue. @robertbartel, do you feel differently?

robertbartel:
Well, according to [this](https://www.boost.org/doc/libs/1_85_0/more/getting_started/unix-variants.html#header-only-libraries), _Boost.Regex_ is not header-only (someone sanity check me to make sure I'm not missing something). 

I don't see a problem with adding support for `{{id}}` replacement within `forcing.path`.  It seems to me like something worth doing anyway.  We don't organize data like this in DMOD right now, though we could consider adapting to it, and I could easily see some users naturally wanting (i.e., regardless of performance implications) to arrange forcings files and other data like BMI configs together in catchment-specific directories.

JoshCu:
I was surprised the timings with the filepattern exact check before attempting the regex weren't measurably faster. Without attaching a debugger I'd guess that the filepattern match only returns true at most once per pattern, but still runs the regex against every other non-exact file match. So in this example it would only prevent regex being used 1/6500th of the time? I'll rerun to confirm and update the tables

PhilMiller:
Oh, yeah, if we're going to try for exact match, we should just try opening the specific file, rather than testing against the entire enumerated contents of the directory.

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -463,8 +463,7 @@ namespace realization {
                 if (directory != nullptr) {
```
JoshCu, line 463: How about something like this?
I just copied the code that stats the file after the regex match.
I tested it just using std and it matched the performance of adding {{id}} to the path variable, 27s serial 3.2s parallel 56x

```suggestion
                if (directory != nullptr) {
                    // check if directory + file_pattern is a file before attempting to iterate
                    struct stat st;
                    if (stat((path + filepattern).c_str(), &st) == 0) {
                        if (S_ISREG(st.st_mode)) {
                            return forcing_params(
                                path + filepattern,
                                provider,
                                simulation_time_config.start_time,
                                simulation_time_config.end_time
                            );
                        }
                    }
```

JoshCu: correction*  3.7s in 56x parallel. So not as fast as changing `path` to accept {{id}}, but probably only because I didn't stat the file when I did that, I just accepted the path without verifying if it existed. the speed difference serially is ~27.2 vs ~27.8 but my testing it just manually running it 5 times so I haven't been including the decimal place in the longer timings

PhilMiller: I just noticed an apparent bug in that code. We probably shouldn't be checking `S_ISREG`, since that would (I think) exclude symbolic links, which should be legitimate, as long as they point at an existent file.

All of this code kinda suffers from the syndrome of "asking permission, rather than forgiveness" - we should ideally just be trying to directly open the files in question.

JoshCu: I think I misunderstood the original code a bit too, it's looping over all the files and checking matches, then returning at the first match it finds right? I mistook it to return all matching files. If it's only returning the first match I can have a go at reworking it to

1. try an open the file without stat
2. if that fails, build a list of every file in the directory
3. run regex against that list
4. if more than one match is found try to open the first match

The line between performance optimization and hack is blurry but I was wondering if it would be faster to get a string of all filenames, separated by some illegal filename character like hash or pipe, then run regex once against that string to pull out the matching files. Rather than running regex once per file?

JoshCu: hopefully this is more clear than my fumbling with suggested changes https://github.com/NOAA-OWP/ngen/compare/master...JoshCu:ngen:open_files_first_ask_questions_later
How about something like this? I modified it to just try and open the file immediately, then if that fails fall back to the regex matching, which also just tries to open the file rather than using stat. The bit at the top of the diff is just inverting the `if directory != nullptr` logic to unindent the code one layer.

JoshCu:
@PhilMiller If this is still of interest, what are the next steps?

1. Add boost regex?
2. Add {{id}} substitution to the path variable?
3. Try to open the file, if that fails, then begin regex matching?

**1** | This is the smallest code change, a moderate speed improvement, but boost.regex is **not header only**

**2** | This is the fastest, slightly larger change, but doesn't verify that the file exists before accepting it

**3** | Slightly slower than 2, much larger change,  seems to be a good option assuming [my code is ok?](https://github.com/NOAA-OWP/ngen/compare/master...JoshCu:ngen:open_files_first_ask_questions_later)

I'm happy to redo any changes and put then in a new pull request if needed :)

JoshCu:
Is that why I didn't have to do anything other than building boost for the regex change to work? The highest version in rocky 9 epel is/was 1.75 so I had to build 1.79 from source anyway https://github.com/JoshCu/NGIAB-CloudInfra/blob/7cef18572883f4b9ec04471b650a87f860e488e3/docker/Dockerfile#L28

program--:
> Is that why I didn't have to do anything other than building boost for the regex change to work? ...

Yeah that'd be my assumption. Another verification that boost.regex doesn't need any additional building: our CI builds the changes successfully, and we only link to the `Boost::boost`/`Boost::headers` CMake target https://github.com/NOAA-OWP/ngen/actions/runs/9687130620/job/26813637727?pr=843#step:3:594

For older boost versions, boost.regex requires the `regex` component when calling `find_package(Boost ...)` to include linking to the compiled code AFAIK

robertbartel:
> As a note, boost.regex is header-only in the general case (at least as of 1.79), as noted [here](https://www.boost.org/doc/libs/1_79_0/libs/regex/doc/html/boost_regex/install.html). There is a build requirement for C++03 and below, and when using ICU support, but I don't think that is needed in this PR unless I'm missing something.

Thanks, @program--.  Looks like they (still) haven't updated Boost's _Getting Started_ page to clearly reflect that and I didn't look hard enough earlier.  If what's stated [here](https://github.com/NOAA-OWP/ngen/blob/master/doc/DEPENDENCIES.md#version-requirements-1) in the dependency doc is still correct, a non-header-only scenario for this shouldn't be possible.

A bit of an aside:  I wanted to quickly see if any other Boost libraries listed as "must be built" didn't really belong there, so I took a brief look at _Boost.Chrono_ (the first one in that list).  Besides details on _Boost.Chrono_ directly, I came across this interesting line about its _Boost.System_ dependency:

> Boost.System has an undocumented feature (use of macro BOOST_ERROR_CODE_HEADER_ONLY) to make it header only.

So, buyer beware on any of the _Getting Started_  "must be built" libraries.

JoshCu:
Closing this as it's included in https://github.com/NOAA-OWP/ngen/pull/897

The route I ended up going was to just try and open the file immediately, then if that failed begin the regex matching. https://github.com/CIROH-UA/ngen/pull/7/commits/4486ed0e7497c246c7f5de26b70449b3d7961c28

PR #845, opened 2024-07-01
Lumped Forcings Engine Data Provider

program--:
This PR adds the **`ForcingsEngineLumpedDataProvider`** class, inheriting from **`ForcingsEngineDataProvider<double, CatchmentAggrDataSelector>`**.

This provider interacts with the python-based Forcings Engine when running in "hydrofabric" mode (i.e. in the configuration file, `GRID_TYPE="hydrofabric"`). The following variables are accessible through the data provider interface:

BMI Variable       | Alias      | Description                                | Units
------------------ | ---------- | ------------------------------------------ | -----
`U2D_ELEMENT`      | `U2D`      | 10-m U-component of wind                   | m/s
`V2D_ELEMENT`      | `V2D`      | 10-m V-component of wind                   | m/s
`T2D_ELEMENT`      | `T2D`      | 2-m Air Temperature                        | K
`Q2D_ELEMENT`      | `Q2D`      | 2-m Specific Humidity                      | kg/kg
`LWDOWN_ELEMENT`   | `LWDOWN`   | Surface downward long-wave radiation flux  | W/m^2
`SWDOWN_ELEMENT`   | `SWDOWN`   | Surface downward short-wave radiation flux | W/m^2
`PSFC_ELEMENT`     | `PSFC`     | Surface Pressure                           | Pa
`RAINRATE_ELEMENT` | `RAINRATE` | Surface Precipitation Rate                 | mm/s

Within the engine's catchment domain, each variable has an aggregated scalar value for each catchment. As such, every constructed lumped data provider is associated with a single catchment (but, the underlying forcings engine instance is shared between them).

~This PR also includes a minor change to the parent base class, which moves the time and config file path member variables into the `protected` scope to allow printing of them in derived functions/exceptions and prevent the need to perform some conversions within `std::chrono` if they were kept in the `private` scope.~

> These changes were incorporated in #851.

## Additions

- Adds `data_access::ForcingsEngineLumpedDataProvider`
- Adds `ForcingsEngineLumpedDataProviderTest` test fixture using AORC data from AWS S3.
- Includes a ESMF-compliant mesh for a hydrofabric subset along USGS Gage-01073000.

## Testing

- Unit tests using GCC 13 and Clang 17.

## Notes

In order to run the forcings engine unit tests prerequisites must be met:

- Set the `$WGRIB2` environment variable with the path to your `wgrib2` executable, e.g.
  ```
  export WGRIB2=$(which wgrib2)
  ```
- Ensure ESMF and esmpy are installing either system-wide or in your virtual environment.
- The master branch of https://github.com/NOAA-OWP/ngen-forcing must be installed.

## Checklist


data/forcing/forcings-engine/hydrofabric_vpu06_esmf_mesh.nc
PhilMiller, line 1: We probably shouldn't add a 15 MB test data file into the main repo. Is there some other way to approach this?

src/forcing/ForcingsEngineLumpedDataProvider.cpp
```diff
@@ -0,0 +1,144 @@
...
+{
+    divide_id_ = convert_divide_id_stoi(divide_id);
+
+    // Check that CAT-ID is an available output name, otherwise we most likely aren't
+    // running the correct configuration of the forcings engine for this class.
+    const auto cat_id_pos = std::find(var_output_names_.begin(), var_output_names_.end(), "CAT-ID");
+    if (cat_id_pos == var_output_names_.end()) {
+        throw std::runtime_error{
+            "Failed to initialize ForcingsEngineLumpedDataProvider: `CAT-ID` is not an output variable of the forcings engine."
+            " Is it running with `GRID_TYPE` set to 'hydrofabric'?"
```
PhilMiller, line 37: For this error message and the one below, it would be good to include the value of `init` so it's easy to know where to look for the misconfiguration.

include/forcing/ForcingsEngineLumpedDataProvider.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include <forcing/ForcingsEngineDataProvider.hpp>
+#include <forcing/DataProviderSelectors.hpp>
+
+namespace data_access {
+
+struct ForcingsEngineLumpedDataProvider final :
+  public ForcingsEngineDataProvider<double, CatchmentAggrDataSelector>
+{
+    using data_type      = double;
+    using selection_type = CatchmentAggrDataSelector;
```
PhilMiller, line 16: Rather than having these type aliases in each derived class, could you please add them to the base `DataProvider` class template?

E.g.

```diff
diff --git a/include/forcing/DataProvider.hpp b/include/forcing/DataProvider.hpp
index a8ffb5604..89da39f15 100644
--- a/include/forcing/DataProvider.hpp
+++ b/include/forcing/DataProvider.hpp
@@ -21,7 +21,7 @@ namespace data_access
             BACK_FILL
     };
 
-    template <class data_type, class selection_type> class DataProvider
+    template <class DataType, class SelectionType> class DataProvider
     {
         /** This class provides a generic interface to data services
         *
@@ -29,6 +29,9 @@ namespace data_access
 
         public:
 
+        using selection_type = SelectionType;
+        using data_type = DataType;
+
         virtual ~DataProvider() = default;
```

data/forcing/forcings-engine/hydrofabric_vpu06_esmf_mesh.nc
program--: There's two ways I think we could look at:

1. Generate the mesh in-memory via a hydrofabric subset (on the python side maybe? Not sure how to approach this) -- better for integration I think, since it simplifies the config file a bit, but not sure on other aspects.

2. Generate a subset of that mesh file to hold only one or two divides -- quick solution. I don't know exactly how Jason generated that, so I'll need to reach out. A CONUS mesh or meshes for each VPU are probably needed anyways, so it'd be good to have that workflow available.

(1) takes a bit more work, but I think it would be good -- assuming that it's possible -- so that users don't need to effectively duplicate data to use the forcings engine. A way to make this work could be to have an input variable on the forcings engine side that accepts WKB (i.e. `uint8_t*`) of the divide geometry, then parse and construct the nodes/elements from that.

Though, if you have any suggestions outside of that, I'm open to anything!

program--: I am currently going through with (2). Awaiting mesh files from Jason.

test/forcing/ForcingsEngineLumpedDataProvider_Test.cpp
```diff
@@ -0,0 +1,125 @@
...
+    struct { int initialized = 0, finalized = 0, size = 1, rank = 0; } mpi_{};
+
+    static void SetUpTestSuite();
+    static void TearDownTestSuite();
+};
+
+using TestFixture = ForcingsEngineLumpedDataProviderTest;
+
+constexpr const char* TestFixture::config_file;
+const std::time_t TestFixture::time_start = data_access::detail::parse_time("2024-01-17 01:00:00");
```
program--, line 44: One note that I just realized I need to change: `time_start` has no actual impact on the Forcings Engine's time domain, so while this is currently incorrect, the instance still starts at 2023-01-17 01:00:00. From what I can think of currently, there isn't an easy way to assert that the time matches, since only the time duration is exposed from the forcings engine (i.e. `get_data_start_time` and others report in seconds with an epoch of 0).

This relates back to #858.

PhilMiller: @jduckerOWP For the Forcings Engine in particular, could it expose a `start_time` variable via BMI that we can use to cross-check that the model engine's simulation start time matches the forcings engine's configured/available-data start time?

PhilMiller: Alternately, could it expose that as a pseudo-variable that we can set via BMI, rather than living in the config file?

data/forcing/forcings-engine/.gitignore
```diff
@@ -0,0 +1,8 @@
+gfs/**/*.grib2
+
+# Temp directory used by the forcings engine instances
+scratch/*
+!scratch/.keep
+
+# Cropped GRIB files
+!gfs/gfs.20240117/00/atmos/gfs.t00z.sfluxgrbf*.grib2
```
PhilMiller, line 8: It would be nice if tests, examples, and such were configured in a way that these weren't going to be written in a path beneath the repository, but rather somewhere in the build tree. Is that a reasonable possibility?

data/forcing/forcings-engine/config_aorc.yml
```diff
@@ -0,0 +1,64 @@
...
+SubOutputHour: 0
+SubOutFreq: 0
+ScratchDir: "data/forcing/forcings-engine/scratch"
+Output: 1
+compressOutput: 0
+floatOutput: 0
+AnAFlag: 0
+LookBack : -9999
+RefcstBDateProc: "202301170000"
+RefcstEDateProc: "202301170100"
```
PhilMiller, line 19: These are the time configurations that conflict with `time_start` and `duration` arguments higher up, right?

program--: Yeah, correct.

data/forcing/forcings-engine/.gitignore
program--: Yeah that makes sense. I think I can just configure the input config file via CMake -- should be fairly straightforward.

src/forcing/ForcingsEngineLumpedDataProvider.cpp
```diff
@@ -0,0 +1,132 @@
...
+        assert(start >= time_begin_);
+
+        const auto end = std::chrono::seconds{selector.get_duration_secs()} + start;
+        assert(end <= time_end_);
+
+        auto current = start;
+        while (current < end) {
+            current += time_step_;
+            bmi_->UpdateUntil(std::chrono::duration_cast<std::chrono::seconds>(current - start).count());
+            acc += *static_cast<double*>(bmi_->GetValuePtr(variable));
```
PhilMiller, line 91: Should this be 
```suggestion
            acc += *static_cast<double*>(bmi_->GetValuePtr(variable)[divide_idx_]);
```
I don't see `divide_index()` or `divide_idx_` being used anywhere in actually getting the value.

Same applies to `get_values` too, I think

src/forcing/ForcingsEngineLumpedDataProvider.cpp
```diff
@@ -0,0 +1,132 @@
...
+    const auto cat_id_span = boost::span<const int>(
+        static_cast<const int*>(bmi_->GetValuePtr("CAT-ID")),
+        size_id_dimension
+    );
+
+    auto divide_id_pos = std::find(cat_id_span.begin(), cat_id_span.end(), divide_id_);
+    if (divide_id_pos == cat_id_span.end()) {
+        // throw std::runtime_error{"Unable to find divide ID `" + divide_id + "` in given Forcings Engine domain"};
+        divide_idx_ = static_cast<std::size_t>(-1);
+    }
```
PhilMiller, line 57: It looks like this isn't actually setting `divide_idx_` in the success case, unless I'm missing it.

test/forcing/ForcingsEngineLumpedDataProvider_Test.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+    #if NGEN_WITH_MPI
+    PMPI_Finalize();
+    #endif
+}
+
+/**
+ * Tests for the flyweight-like design of provider storage by getting
+ * a new instance of the forcings engine and verifying that it points
+ * to the same address as the static initialized `provider_` member.
```
PhilMiller, line 81: Since I had to dig into the code to refresh my memory and understanding, I think it's worth calling out in the comments:
```suggestion
 * to the same address as the static initialized `provider_` member,
 * based on matching `init`, and shared over distinct `divide_id`.
```

PhilMiller:
Other than the `divide_idx_` remarks, the rest of this looks pretty good.

test/forcing/ForcingsEngineLumpedDataProvider_Test.cpp
jduckerOWP: @PhilMiller we certainly can expose a start time via BMI for cross-checking timestamps between the Forcings Engine and the framework. Regarding @program-- comment for the timestamp discrepancy we can discuss this more in our meeting today but generally the NWM operational forecast iterations are initialize at some start time (0z,6z,12z,18z etc.) and the first available forecast data is at forecast hour 1 of the simulation. Therefore, generally hour 0 (or start time of simulation) will basically need to be forced with some "hot start" fields if models require forcings at the actual start time. For retrospective datasets however, we can modify the Forcings Engine to directly provide meteorological forcings at that start time hour we'll just need some revisions to the workflow to make that work.

test/forcing/ForcingsEngineLumpedDataProvider_Test.cpp
```diff
@@ -0,0 +1,125 @@
...
+{
+    constexpr std::array<const char*, 8> expected_variables = {
+        "U2D_ELEMENT",
+        "V2D_ELEMENT",
+        "LWDOWN_ELEMENT",
+        "SWDOWN_ELEMENT",
+        "T2D_ELEMENT",
+        "Q2D_ELEMENT",
+        "PSFC_ELEMENT",
+        "RAINRATE_ELEMENT"
```
jduckerOWP, line 105: With the recent upgrades pushed to the NextGen Forcings Engine, some operational configurations (HRRR, RAP, GFS, NBM) can now also extract and advertise a new BMI field called "LQFRAC_ELEMENT" (and "LQFRAC_NODE" if unstructured mesh domain configuration). This field is described as the "liquid fraction of precipitation" and has units of % (0-100 range). For this first iteration of the NextGen Forcings Engine, it'll be okay if we leave it out for this PR and we can just merge the data later on if desired. The complexity of that variable field is that at times, it will or won't be available pending on the NWM operational configuration we apply. Will be happy to discuss these details further in our meeting!

test/forcing/ForcingsEngineLumpedDataProvider_Test.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+    #if NGEN_WITH_MPI
+    PMPI_Finalize();
+    #endif
+}
+
+/**
+ * Tests for the flyweight-like design of provider storage by getting
+ * a new instance of the forcings engine and verifying that it points
+ * to the same address as the static initialized `provider_` memberm
```
PhilMiller, line 79:
```suggestion
 * to the same address as the static initialized `provider_` member,
```
Tiny typo

program--: 🤦🏾 thanks

test/forcing/ForcingsEngineLumpedDataProvider_Test.cpp
PhilMiller: Since it won't always be present, we shouldn't require it in the test.

test/forcing/ForcingsEngineLumpedDataProvider_Test.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+    ForcingsEngineLumpedDataProviderTest()
+    {
+        #if NGEN_WITH_MPI
+        MPI_Comm_size(MPI_COMM_WORLD, &mpi_.size);
+        MPI_Comm_rank(MPI_COMM_WORLD, &mpi_.rank);
+        #endif
+    }
+
+    static constexpr const char* config_file = NGEN_LUMPED_CONFIG_PATH;
```
PhilMiller, line 31: Does this actually end up expanding to a string literal like intended? I guess if CMake ensures that the argument as written gets shell-escaped so that the macro definition includes the `"` around the name, then it would. That's just surprising to me, now that I noticed it.

src/forcing/ForcingsEngineLumpedDataProvider.cpp
```diff
@@ -0,0 +1,134 @@
+#include "DataProvider.hpp"
+#include <chrono>
+#include <forcing/ForcingsEngineLumpedDataProvider.hpp>
+
+namespace data_access {
+
+using Provider     = ForcingsEngineLumpedDataProvider;
+using BaseProvider = Provider::base_type;
+
+std::size_t convert_divide_id_stoi(const std::string& divide_id)
```
PhilMiller, line 10: Is there an existing utility function elsewhere that already serves this purpose? I'm concerned about having a copy here that might need to change if the format of divide IDs changes.

test/forcing/ForcingsEngineLumpedDataProvider_Test.cpp
program--: Yeah, the compilation db after configuring contains the following for this file:

```
-DNGEN_LUMPED_CONFIG_PATH=\\\"/path/to/ngen/build/test/config_aorc.yml\\\"
```

It may not be the best approach, but nothing else immediately comes to my mind as an alternative. But, if you prefer a different approach, I'm open to it!

src/forcing/ForcingsEngineLumpedDataProvider.cpp
PhilMiller: Ok, it's fine.

src/forcing/ForcingsEngineLumpedDataProvider.cpp
```diff
@@ -0,0 +1,134 @@
...
+{
+    auto separator = divide_id.find('-');
+    const char* split = (
+        separator == std::string::npos
+        ? &divide_id[0]
+        : &divide_id[separator + 1]
+    );
+
+    return std::atol(split);
```
donaldwj, line 19: I normally prefer strtol() to atol(). For most linux compiliers the actual conversion will be a call internal call to strtol so it is easier for strtol calls to be implemented.

donaldwj (approved):
I requested one small change but it is not important. Otherwise I see no issues

src/forcing/ForcingsEngineLumpedDataProvider.cpp
program--: Oh this is a good point -- plus, thinking about it a it more, strtol() is safer. I'll make an issue as a reminder to change this. Thanks 😄

PR #846, opened 2024-07-02
Remove MPI_Barriers before routing to increase speed.

JoshCu:
# Improve ngen-parallel Performance with T-route

## Problem Statement
When running ngen-parallel with `-n` near or equal to core count, T-route's performance is severely degraded. T-route doesn't parallelize well with MPI, and moving the finalize step to before the routing frees up CPU resources for T-route to use different parallelization strategies.

## Changes Made
- Moved MPI_Finalize before the routing step
- Applied additional T-route performance patch (link to PR will be added)

The troute change is semi related as it converts a for loop that consumes the majority of t-route's execution time to use multiprocessing. That performance improvement doesn't work while MPI wait is consuming every CPU cycle.

## Performance Results
Testing setup: ~6500 catchments, 24 timesteps, dual Xeon 28C 56T total (same as #843)

| Configuration | NGen::routing Time (s) |
|---------------|------------------------|
| Serial | 13.9106 |
| Parallel (-n 56 unpatched) | 21.6952 |
| Parallel (-n 56 patched) | 10.761 |
| Parallel (-n 56 patched + T-route perf patch) | ~6 (more testing needed) |  

## Explanation
1. T-route has a step that transposes per catchment/nexus files into per timestep files.
2. This transpose step is the longest part of T-route's execution and scales poorly with more catchments or timesteps.
3. In the unpatched version, all ranks except zero progress to MPI_Finalize and then begin polling while waiting for the routing thread to finish.
4. This polling maxes out the CPU on every rank while rank 0 computes the routing.
5. Moving Finalize before routing ensures all threads are finished before routing begins, preventing CPU maxout.

## Future Work
- Consider reworking the file transposition step for more efficient I/O (out of scope for this change)

## Testing
- Used the same setup as #843
- Manual runs show consistent improvement, but automated testing may be required for more accurate results (especially for simulation that shows a ~10% improvement with a similar deviation)

## Performance Visualization
Performance Flamegraph
*Perf flamegraph of the entire ngen run (unpatched), should be interactive if downloaded*

## Additional Notes
1. Moving the finalize before timing log output appears to make the ngen simulation step ~10% faster, but may affect timing accuracy.
2. This use case exaggerates T-route performance degradation due to using 56 MPI ranks on 56 cores.
3. ngen was built using mpich 3.3.2. I saw online that changing the MPI wait policy to poll less frequently should help, but I couldn't get it to work
4. I'm not entirely sure of the implications of calling `return 0;` right after finalize for `mpi_rank != 0;`. I thought finalize would kill/return the subprocesses but without explicitly returning them I got segfaults. Reccomendations seem to be that as little as possible should be performed after calling finalize, but seeing as all computation after that point is done by one thread I can just return the others?

## Next Steps
- [ ] Conduct automated benchmarks to verify performance improvements to troute
- [ ] Conduct automated benchmarks to verify performance improvements to ngen simulation

### DIFF
--- a/src/NGen.cpp
+++ b/src/NGen.cpp
@@ -558,57 +558,46 @@ int main(int argc, char *argv[]) {
     } //done time
 
 #if NGEN_WITH_MPI
-    MPI_Barrier(MPI_COMM_WORLD);
+    MPI_Finalize();
+    if (mpi_rank != 0)
+    {
+        return 0;
+    }
 #endif
 
-    if (mpi_rank == 0)
-    {
-        std::cout << "Finished " << manager->Simulation_Time_Object->get_total_output_times() << " timesteps." << std::endl;
-    }
+    std::cout << "Finished " << manager->Simulation_Time_Object->get_total_output_times() << " timesteps." << std::endl;
 
     auto time_done_simulation = std::chrono::steady_clock::now();
     std::chrono::duration<double> time_elapsed_simulation = time_done_simulation - time_done_init;
 
-#if NGEN_WITH_MPI
-    MPI_Barrier(MPI_COMM_WORLD);
-#endif
-
 #if NGEN_WITH_ROUTING
-    if (mpi_rank == 0)
-    { // Run t-route from single process
-        if(manager->get_using_routing()) {
-          //Note: Currently, delta_time is set in the t-route yaml configuration file, and the
-          //number_of_timesteps is determined from the total number of nexus outputs in t-route.
-          //It is recommended to still pass these values to the routing_py_adapter object in
-          //case a future implmentation needs these two values from the ngen framework.
-          int number_of_timesteps = manager->Simulation_Time_Object->get_total_output_times();
+    // Run t-route from single process only rank zero reaches this point
+    if(manager->get_using_routing()) {
+      //Note: Currently, delta_time is set in the t-route yaml configuration file, and the
+      //number_of_timesteps is determined from the total number of nexus outputs in t-route.
+      //It is recommended to still pass these values to the routing_py_adapter object in
+      //case a future implmentation needs these two values from the ngen framework.
+      int number_of_timesteps = manager->Simulation_Time_Object->get_total_output_times();
 
-          int delta_time = manager->Simulation_Time_Object->get_output_interval_seconds();
-          
-          router->route(number_of_timesteps, delta_time); 
-        }
+      int delta_time = manager->Simulation_Time_Object->get_output_interval_seconds();
+      
+      router->route(number_of_timesteps, delta_time); 
     }
 #endif
 
     auto time_done_routing = std::chrono::steady_clock::now();
     std::chrono::duration<double> time_elapsed_routing = time_done_routing - time_done_simulation;
 
-    if (mpi_rank == 0)
-    {
-        std::cout << "NGen top-level timings:"
-                  << "\n\tNGen::init: " << time_elapsed_init.count()
-                  << "\n\tNGen::simulation: " << time_elapsed_simulation.count()
+    std::cout << "NGen top-level timings:"
+              << "\n\tNGen::init: " << time_elapsed_init.count()
+              << "\n\tNGen::simulation: " << time_elapsed_simulation.count()
 #if NGEN_WITH_ROUTING
-                  << "\n\tNGen::routing: " << time_elapsed_routing.count()
+              << "\n\tNGen::routing: " << time_elapsed_routing.count()
 #endif
-                  << std::endl;
-    }
+              << std::endl;
 
   manager->finalize();
 
-#if NGEN_WITH_MPI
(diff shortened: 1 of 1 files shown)


PhilMiller:
I *really* like the idea here, but there's a fundamental challenge. As MPI is specified, we can't actually be confident that anything after a call to `MPI_Finalize` actually runs. I'll take a deeper look at whether there's a nicer way we can do this.

PhilMiller:
Ok, I've looked at this a little bit more, and here's my tentative suggestion:

Rather than having all of the non-0 processes finalize and exit after the catchment simulation is complete, have them all call `MPI_Ibarrier`, followed by a loop of `MPI_Test` on the `Ibarrier` request and `sleep` for a decent time slice (e.g. 100 ms) until it's complete. Rank 0 would only enter the barrier after completing routing.

Is that something you want to try to implement? If not, I can find some time to work up an alternative patch. In the latter case, do you mind if I push it to your branch for this PR?

Ultimately, we expect to more deeply integrate t-route with parallelization directly into ngen with BMI. Until that's implemented, though, it's pretty reasonable to find expedient ways to improve performance.

JoshCu:
This version works so far as the waiting mpi threads now no longer max out the CPU, but it's tricky to say what I'm actually measuring in terms of performance since this went through https://github.com/NOAA-OWP/t-route/pull/795 
I need to run a test on some data with more than 25 timesteps as the difference with and without this fix is ~<1s,  ~9s vs ~10s.

I also wasn't sure if `manager->finalize();` could go before the barrier or after it? If the routing was using MPI, would we need to wait for all threads to reach the post routing barrier before calling finalize?

JoshCu:
# Performance Testing Results
In summary: it makes routing ~1.3x-1.4x faster
Apologies for the previously inconsistent performance testing. The routing configuration has been optimized, and the output format changed from CSV to NetCDF. Here are the updated results using the current version of t-route and an optimized routing configuration:

## Test Setup

- **Cores:** 96
- **Subnetwork target size:** 100 (determined to be the fastest from multiple tests)
- **t-route output format:** NetCDF
- **Catchments:** 60,000
- **Time period:** 240 hours (in five-minute timesteps in t-route)

## Results

1. **t-route run standalone in NGIAB**
   - Method: Running `python -m nwm_routing` manually
   - Execution time: **115 seconds**

2. **t-route run via ngen in NGIAB using ngen master**
   - Execution time: **161 seconds**

3. **t-route run via ngen in NGIAB using this PR**
   - Execution time: **120 seconds**

```bash
2024-09-04 23:27:24,582 - root - INFO - [__main__.py:340 - main_v04]: ************ TIMING SUMMARY ************
2024-09-04 23:27:24,583 - root - INFO - [__main__.py:341 - main_v04]: ----------------------------------------
2024-09-04 23:27:24,583 - root - INFO - [__main__.py:342 - main_v04]: Network graph construction: 20.63 secs, 17.12 %
2024-09-04 23:27:24,583 - root - INFO - [__main__.py:349 - main_v04]: Forcing array construction: 29.57 secs, 24.55 %
2024-09-04 23:27:24,583 - root - INFO - [__main__.py:356 - main_v04]: Routing computations: 59.15 secs, 49.09 %
2024-09-04 23:27:24,583 - root - INFO - [__main__.py:363 - main_v04]: Output writing: 10.99 secs, 9.12 %
2024-09-04 23:27:24,583 - root - INFO - [__main__.py:370 - main_v04]: ----------------------------------------
2024-09-04 23:27:24,583 - root - INFO - [__main__.py:371 - main_v04]: Total execution time: 120.33999999999999 secs
Finished routing
NGen top-level timings:
	NGen::init: 103.262
	NGen::simulation: 234.654
	NGen::routing: 120.578
```
and that init time can be reduced to 26 seconds this https://github.com/NOAA-OWP/ngen/compare/master...JoshCu:ngen:open_files_first_ask_questions_later

JoshCu:
Recording this here with for reference:
I've been doing other testing and the first mpi barrier after model execution, before the routing also has the same issue. On a single machine, when a rank completes simulation it polls as fast as possible, pinning that core at 100% which results in my machine lowering the clock speed of the cpu. It's only slower because of the reduction in clock speed and not as a result of other processes being unable to use the cores (like with t-route) so it's <5% slowdown. It's a marginal difference on a desktop and I'm guessing this wouldn't be a problem on servers with fixed clock speed cpus. Might save some electricity though!

JoshCu:
Closing as this is included in https://github.com/NOAA-OWP/ngen/pull/897

https://github.com/NOAA-OWP/ngen/pull/897/commits/15748991408c4404fa54b25e33c3cf7bc6f9fef8

In summary, I ended up using an Ibarrier after the end of the model simulation section to very slightly speed up that section. As well as replacing the barrier with a loop that polls every 100ms to check if the t-route rank is finished before finalizing.

PR #849, opened 2024-07-03
Add `const` specifiers to various virtual member functions

program--:
This PR adds `const` specifiers as applicable to classes associated with `DataProvider<...>`.

The primary reason for this change is to ensure const-correctness within the DataProvider class hierarchy, as there's no reason not to (at least no reasons that I can think of immediately).

## Changes

`const` qualifies the following:

<details><summary><strong><code>data_access::DataProvider</code></strong></summary>

+ `get_available_variable_names`
+ `get_data_start_time`
+ `get_data_stop_time`
+ `record_duration`
+ `get_ts_index_for_time`
+ `is_property_sum_over_time_step`

</details>

<details><summary><strong><code>realization::Formulation</code></strong></summary>

+ `get_formulation_type`
+ `get_required_parameters`

</details>

<details><summary><strong><code>realization::Catchment_Formulation</code></strong></summary>

+ `get_output_header_line`

</details>

<details><summary><strong><code>realization::Bmi_Formulation</code></strong></summary>

+ `convert_model_time`
+ `get_bmi_input_variables`
+ `get_bmi_model_start_time_forcing_offset_s`
+ `get_bmi_output_variables`
+ `get_config_mapped_variable_name`
+ `get_model_current_time`
+ `get_model_end_time`
+ `get_model_type_name`
+ `is_bmi_input_variable`
+ `is_bmi_model_time_step_fixed`
+ `is_bmi_output_variable`
+ `is_model_initialized`

</details>

## Checklist

### DIFF
--- a/include/forcing/DataProvider.hpp
+++ b/include/forcing/DataProvider.hpp
@@ -40,18 +40,18 @@ namespace data_access
 
         /** Return the variables that are accessable by this data provider */
 
-        virtual boost::span<const std::string> get_available_variable_names() const noexcept = 0;
+        virtual boost::span<const std::string> get_available_variable_names() const = 0;
 
         /** Return the first valid time for which data from the request variable  can be requested */
 
-        virtual long get_data_start_time() const noexcept = 0;
+        virtual long get_data_start_time() const = 0;
 
         /** Return the last valid time for which data from the requested variable can be requested */
 
-        virtual long get_data_stop_time() const noexcept = 0;
+        virtual long get_data_stop_time() const = 0;
 
         /** Return the stride in the time dimension */
-        virtual long record_duration() const noexcept = 0;
+        virtual long record_duration() const = 0;
 
         /**
          * Get the index of the data time step that contains the given point in time.
@@ -93,7 +93,7 @@ namespace data_access
          */
         virtual std::vector<data_type> get_values(const selection_type& selector, ReSampleMethod m=SUM) = 0;
 
-        virtual bool is_property_sum_over_time_step(const std::string& name) {return false; }
+        virtual bool is_property_sum_over_time_step(const std::string& name) const {return false; }
 
         private:
     };
--- a/include/forcing/DataProvider.hpp
+++ b/include/forcing/DataProvider.hpp
@@ -40,18 +40,18 @@ namespace data_access
 
         /** Return the variables that are accessable by this data provider */
 
-        virtual boost::span<const std::string> get_available_variable_names() = 0;
+        virtual boost::span<const std::string> get_available_variable_names() const noexcept = 0;
 
         /** Return the first valid time for which data from the request variable  can be requested */
 
-        virtual long get_data_start_time() = 0;
+        virtual long get_data_start_time() const noexcept = 0;
 
         /** Return the last valid time for which data from the requested variable can be requested */
 
-        virtual long get_data_stop_time() = 0;
+        virtual long get_data_stop_time() const noexcept = 0;
 
         /** Return the stride in the time dimension */
-        virtual long record_duration() = 0;
+        virtual long record_duration() const noexcept = 0;
 
         /**
          * Get the index of the data time step that contains the given point in time.
@@ -62,7 +62,7 @@ namespace data_access
          * @return The index of the forcing time step that contains the given point in time.
          * @throws std::out_of_range If the given point is not in any time step.
          */
-        virtual size_t get_ts_index_for_time(const time_t &epoch_time) = 0;
+        virtual size_t get_ts_index_for_time(const time_t &epoch_time) const = 0;
 
         /**
          * Get the value of a forcing property for an arbitrary time period, converting units if needed.
--- a/include/forcing/ForcingsEngineDataProvider.hpp
+++ b/include/forcing/ForcingsEngineDataProvider.hpp
@@ -104,27 +104,27 @@ struct ForcingsEngineDataProvider
 
     ~ForcingsEngineDataProvider() override = default;
 
-    boost::span<const std::string> get_available_variable_names() override
+    boost::span<const std::string> get_available_variable_names() const noexcept override
     {
         return var_output_names_;
     }
 
-    long get_data_start_time() override
+    long get_data_start_time() const noexcept override
     {
         return clock_type::to_time_t(time_begin_);
     }
 
-    long get_data_stop_time() override
+    long get_data_stop_time() const noexcept override
     {
         return clock_type::to_time_t(time_end_);
     }
 
-    long record_duration() override
+    long record_duration() const noexcept override
     {
         return std::chrono::duration_cast<std::chrono::seconds>(time_step_).count();
     }
 
-    size_t get_ts_index_for_time(const time_t& epoch_time) override
+    size_t get_ts_index_for_time(const time_t& epoch_time) const override
     {
         const auto epoch = clock_type::from_time_t(epoch_time);
 
--- a/include/forcing/ForcingsEngineDataProvider.hpp
+++ b/include/forcing/ForcingsEngineDataProvider.hpp
@@ -104,22 +104,22 @@ struct ForcingsEngineDataProvider
 
     ~ForcingsEngineDataProvider() override = default;
 
-    boost::span<const std::string> get_available_variable_names() const noexcept override
+    boost::span<const std::string> get_available_variable_names() const override
     {
         return var_output_names_;
     }
 
-    long get_data_start_time() const noexcept override
+    long get_data_start_time() const override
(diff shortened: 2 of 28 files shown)


include/forcing/DataProvider.hpp
```diff
@@ -40,18 +40,18 @@ namespace data_access
 
         /** Return the variables that are accessable by this data provider */
```
PhilMiller, line 41: I just noticed a typo here
```suggestion
        /** Return the variables that are accessible by this data provider */
```
Probably not suitable to fix in this PR.

PhilMiller:
I think `is_property_sum_over_time_step` should also be `const`

PhilMiller:
Re `noexcept`:
* I don't think it's a good idea to start applying that qualifier without really thinking about it project-wide
* In this particular case, I don't think we can actually apply it accurately to all implementations of these `DataProvider` methods. They may call into BMI routines that might fail for whatever reason. We want that exception to get thrown and potentially handled at whatever level, and not just immediately call `terminate` when they hit this layer of the call stack.

So, my take is that we should go ahead with adding `const`, and hold off on `noexcept` for now.

program--:
> Re `noexcept`:
> 
> ...

Yeah I agree. I forgot the extent of the DataProvider hierarchy (thankfully CI reminded me 😄), but I'm happy with only `const` being added.

include/realizations/catchment/Bmi_Module_Formulation.hpp
```diff
@@ -306,9 +306,9 @@ namespace realization {
          *
          * @return Shared pointer to the backing model object that implements the BMI.
          */
-        std::shared_ptr<models::bmi::Bmi_Adapter> get_bmi_model();
+        std::shared_ptr<models::bmi::Bmi_Adapter> get_bmi_model() const;
```
program--, line 309: This change I'm not entirely sure on -- it's needed for some of the other const qualified members but, I think to be correctly const qualified, the return type should be

```cpp
std::shared_ptr<const models::bmi::Bmi_Adapter>
```

But is that a possible change? Need to check

include/forcing/ForcingsEngineDataProvider.hpp
```diff
@@ -104,27 +104,27 @@ struct ForcingsEngineDataProvider
 
     ~ForcingsEngineDataProvider() override = default;
 
-    boost::span<const std::string> get_available_variable_names() override
+    boost::span<const std::string> get_available_variable_names() const noexcept override
```
program--, line 107: Missed a few `noexcept`

include/realizations/catchment/Bmi_Module_Formulation.hpp
program--: `std::shared_ptr<const Bmi_Adapter>` is valid (which should've been obvious) -- however, it's not possible to use this result in other const qualified members because the C++ BMI interface does not const qualify its virtual functions, i.e. `bmi->GetOutputVarNames()` is not `const`.

In order for this to work, it'd require changes in the BMI header. An alternative is for member functions to use the `bmi_model` member directly, which circumvents this, but requires more changes... or just leave it and accept it as is 🥲

PhilMiller: I think it's fine to leave `get_bmi_model()` as non-`const` - it's `protected`, so it's not like we need to reason about unrelated classes messing with things.

The idiomatic C++ way to address this is an overload pair along the lines of 
```c++
std::shared_ptr<models::bmi::Bmi_Adapter> get_bmi_model();
std::shared_ptr<const models::bmi::Bmi_Adapter> get_bmi_model() const;
```
Obviously, that doesn't actually work in this case because that `const Bmi_Adapter` is pretty useless.

src/realizations/catchment/Bmi_Multi_Formulation.cpp
```diff
@@ -186,18 +186,11 @@ const bool &Bmi_Multi_Formulation::get_allow_model_exceed_end_time() const {
...
-boost::span<const std::string> Bmi_Multi_Formulation::get_available_variable_names() {
-    if (is_model_initialized() && available_forcings.empty()) {
-        for (const nested_module_ptr &module: modules) {
-            for (const std::string &out_var_name: module->get_bmi_output_variables()) {
-                available_forcings.push_back(module->get_config_mapped_variable_name(out_var_name));
-            }
-        }
-    }
+boost::span<const std::string> Bmi_Multi_Formulation::get_available_variable_names() const {
     return available_forcings;
```
PhilMiller, line 190: Where does this get initialized in the `Bmi_Multi_Formulation` case? I don't think it shares that code with `Bmi_Module_Formulation`

include/realizations/catchment/Bmi_Module_Formulation.hpp
PhilMiller: Was `get_available_variable_names` the only place that the limitation on use of `get_bmi_model` came up? If so, great.

program--: No, it's used in quite a few functions intended to be `const`. I think it may easy enough to leave `get_bmi_model()` as non-`const` and replace them with direct member access, since most of the calls are of the form `get_bmi_model()->GetSomething()`.

PhilMiller: It looks like the codebase was ultimately happy with this revision as made. I don't see any collateral damage changes in the PR, and it compiled (and passed the tests).

PhilMiller: If you're worried about "`get_bmi_model()` should be deeply-`const` but that's infeasible", I wouldn't fret over it.

src/realizations/catchment/Bmi_Multi_Formulation.cpp
program--: Hmm I think for each model in the multi formulation, when `init_nested_module<T>` is called, it looks like it calls

```
Bmi_Module_Formulation::create_formulation ->
  Bmi_Module_Formulation::inner_create_formulation ->
  if (model_initialized) { ... available_forcings.push_back ... }
```

so, eventually they are initialized via `Bmi_Module_Formulation` -- but I'm not sure that it's correctly initialized, since I don't see any logic to handle duplicate names prior to `push_back`. I assumed the tests would check, but looking over them now, I don't see a call to `available_forcings` or `get_available_variable_names()`.

include/realizations/catchment/Bmi_Module_Formulation.hpp
program--: Yeah, that was my only concern. It irks me a bit, but I guess there's no real harm in it 😅

src/realizations/catchment/Bmi_Multi_Formulation.cpp
PhilMiller: That's initializing the `available_forcings` member of the individual modules, though. Without the deleted code here, where would the `Bmi_Multi_Formulation`'s member `available_forcings` get initialized as the concatenation of those?

(ignoring the fact that it should be checked as the union of disjoint sets - that's maybe an issue to file, but out of scope for this PR)

program--: Ohhh right 🤦🏾 I see now. I'll add back in the deleted code and move it within `create_multi_formulation` and that should handle it correctly then.

PhilMiller: I think I want to challenge you to add a test that exercises this. I think it never comes up currently because it would require something weird like a nested `Bmi_Multi_Formulation`, I guess inside another `Bmi_Multi_Formulation`. Otherwise, there's perhaps nothing that will actually call `Bmi_Multi_Formulation::get_available_variable_names()`

PhilMiller: I think there's maybe also a case where the realization configuration doesn't specify any output variables, but I'd need to read the code more carefully to be sure of that one hitting this.

If you want to investigate this somewhat dynamically, just add a `throw` in here, and see if anything ever runs into it.

program--: I'm not sure if it's what you had in mind, but if you get a chance to take a look, I added a simple test to ensure `get_available_variable_names()` returns expected values after construction of a Multi-BMI formulation. With the code deleted, as expected, the available forcings vector isn't filled.

PhilMiller (approved):

PhilMiller:
This all looks good, but I'm sick enough that I'm not going to press "merge" on anything myself today.

PR #851, opened 2024-07-03
Minor changes to Forcings Engine Data Provider for derived implementations

program--:
This PR makes 2 changes to support derived implementations:

1. Moves timing and configuration file path member variables outside of `private` scope into `protected` so derived implementations can directly access these.
2. Adds `ForcingsEngineDataProvider::ensure_variable`, which is a helper function that helps handle variable name aliasing from what NGen expects to what the forcings engine expects.

## Additions

- `ForcingsEngineDataProvider::ensure_variable`

## Changes

- `init_`, `time_begin_`, `time_end_`, and `time_step_` from `private` to `protected` scope.

## Checklist

### DIFF
--- a/include/forcing/ForcingsEngineDataProvider.hpp
+++ b/include/forcing/ForcingsEngineDataProvider.hpp
@@ -194,6 +194,27 @@ struct ForcingsEngineDataProvider
         var_output_names_ = bmi_->GetOutputVarNames();
     }
 
+    std::string ensure_variable(std::string name, const std::string& suffix = "_ELEMENT") const
+    {
+        // TODO: use get_available_var_names() once const
+        auto vars = boost::span<const std::string>{var_output_names_};
+
+        if (std::find(vars.begin(), vars.end(), name) != vars.end()) {
+          return name;
+        }
+
+        auto suffixed_name = std::move(name) + suffix;
+        if (std::find(vars.begin(), vars.end(), suffixed_name) != vars.end()) {
+          return suffixed_name;
+        }
+
+        throw std::runtime_error{
+          "ForcingsEngineDataProvider: neither variable `"
+          + suffixed_name.substr(0, suffixed_name.size() - suffix.size())
+          + "` nor `" + suffixed_name + "` exist."
+        };
+    }
+
     //! Forcings Engine instance
     std::shared_ptr<models::bmi::Bmi_Py_Adapter> bmi_ = nullptr;
 
--- a/include/forcing/ForcingsEngineDataProvider.hpp
+++ b/include/forcing/ForcingsEngineDataProvider.hpp
@@ -200,10 +200,6 @@ struct ForcingsEngineDataProvider
     //! Output variable names
     std::vector<std::string> var_output_names_{};
 
-  private:
-    //! Initialization config file path
-    std::string init_;
-
     //! Calendar time for simulation beginning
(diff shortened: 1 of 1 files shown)


include/forcing/ForcingsEngineDataProvider.hpp
```diff
@@ -100,7 +100,7 @@ struct ForcingsEngineDataProvider
 {
     using data_type = DataType;
     using selection_type = SelectionType;
-    using clock_type = std::chrono::system_clock;
+    using clock_type = std::chrono::steady_clock;
```
program--, line 103: Looking over this again, this change may be a bit premature; I'm not fully convinced on it just yet.

The main reason is that, this affects `get_data_start_time` and `get_data_stop_time` to not necessarily be calendar time. In a simulation, that's *typically* fine, since we're concerned with time relative to the start of the simulation, but that gets a bit hairy when time series data is introduced.

May need to think on this a bit more.

PhilMiller: The right thing to do is, I think, for us to actually introduce a `simulation_clock` that is not one of the standard library types. From there, we can actually define things consistently, and explicitly without reference to external time.

I was digging way into the C++ standards documentation on this, thinking that satisfying the requirement (until C++23, with [P2212](https://wg21.link/P2212)) for such a type to have a member `static time_point now()` would be problematic, but I don't think it is. It means we can't have distinct instances that advance independently, but that's not really an issue with current usage, I don't think.

program--: I have the same thought, and I think it coincides with what the `Simulation_Time` class achieves, but more idiomatic.

As far as meeting the `static time_point now()` requirement, it might be sufficient enough to have something *like* a static instance of `Simulation_Time` that the clock type relies on, and which controls the main simulation loop -- but yeah I'm not sure the implication of something like that.

Nonetheless I also don't think it'll pose a big problem though

PhilMiller: It's really not much: https://en.cppreference.com/w/cpp/named_req/Clock

It might be as simple as
```c++
struct Simulation_Clock
{
  using rep = double;
  using period = std::ratio<1>;
  using duration = std::chrono::duration<rep, period>;
  using time_point = std::chrono::time_point<Simulation_Clock>;

  static bool is_steady() { return false; }
  static time_point now(); // Defined elsewhere, from value of global/static variable
};
```

We may have to fiddle around a bit with `rep = int64_t` or `period = std::milli` or such, but those would be minor in comparison to definitively not referencing a system implementation, and them implementing conversion appropriate functions to convert to calendar timestamps from configs. The conversion would be roughly of the form `sim_time_from_utc(in) { return in - sim_start_time; }`, but the types in that are a bit finicky.

PhilMiller: (edited to correct form of `is_steady`)

program--: I've made issue #858 for continuing this discussion as needed, as it's probably not suitable to work into this PR.

program--: As far as the original comment, I think I will revert from `steady_clock` back to `system_clock` for now.

PhilMiller (approved):

PR #853, opened 2024-07-05
chore: update to `pybind11` to `v2.11.1`

aaraney:
The only seemingly relevant change was introduced in `v2.11.0`: 

> `py::initialize_interpreter()` using `PyConfig_InitPythonConfig()` instead of `PyConfig_InitIsolatedConfig()`, to obtain complete `sys.path`. [#4473](https://github.com/pybind/pybind11/pull/4473)

This should address python virtual environment resolution issues on linux systems.

## Changes

- Update vendored `pybind11` version to `v2.11.1` (see [release notes](https://pybind11.readthedocs.io/en/stable/changelog.html#version-2-11-1-july-17-2023)).


aaraney:
Testing this locally on macOS, this does not seem to solve the issues outlined in #441 and #628.

Issue #848, opened 2024-07-03
No output written if `output_root` is specified in realization config and does not exist or is a file


aaraney:
## Current behavior

If `output_root` is specified in a realization config and points to an existing file or a non-existent directory, NextGen does not write output.

For history sake, this feature was introduce in #531.

## Expected behavior

I think this is up for debate. There is an argument for `NextGen` attempting to create the output directory if it does not exist, but I can also see the other side. In either case, if `output_root` points to a file an exception should be raised. Likewise, if the expected behavior is that `NextGen` attempts to create the directory but cannot for whatever reason, an exception should be raised. It is desirable that `NextGen` performs these checks before entering the simulation loop.

## Steps to replicate behavior

Modified [example realization config](https://github.com/NOAA-OWP/ngen/blob/c53584a27ff1c9bdc38955eb98782c6bc07424f2/data/example_realization_config.json)

<details><summary>realization config</summary>

```json
{
    "global": {
      "formulations": [
        {
          "name": "bmi_c++",
          "params": {
            "model_type_name": "test_bmi_cpp",
            "library_file": "./extern/test_bmi_cpp/cmake_build/libtestbmicppmodel",
            "init_config": "./data/bmi/c/test/test_bmi_c_config.ini",
            "main_output_variable": "OUTPUT_VAR_2",
            "variables_names_map" : {
              "INPUT_VAR_2": "TMP_2maboveground",
              "INPUT_VAR_1": "precip_rate"
            },
            "uses_forcing_file": false
          }
        }
      ],
      "forcing": {
          "file_pattern": ".*{{id}}.*.csv",
          "path": "./data/forcing/"
      }
    },
    "time": {
        "start_time": "2015-12-01 00:00:00",
        "end_time": "2015-12-30 23:00:00",
        "output_interval": 3600
    },
    "output_root": "./output/"
}
```

</details>

Cases:

- `output_root` does not exist:
    ```shell
    test ! -e output && \
    cmake_build/ngen data/catchment_data.geojson all data/nexus_data.geojson all above_realization_config.json && \
    test ! -e output 
    ```

- `output_root` is a file:
    ```shell
    test ! -d output && \
    touch output && \
    test -f output && \
    cmake_build/ngen data/catchment_data.geojson all data/nexus_data.geojson all above_realization_config.json && \
    test ! -d output
    ```

stcui007:
[...]

PR #855, opened 2024-07-17
Fix output_root

stcui007:
This PR fixes the problem of directory that `output_root` pointed to does not exist. If this is the case, the code creates a sub-directory based in the name given by the `output_root` in the realization. The code defaults to writing the outputs to the project directory if `output_root` does not exist as in the current realization.

This is an attempt to solve the problem raised in issues #857 and #848, where large number of output files are written. This is a common problem for running ngen on any large basin.

## Additions

-

## Removals

-

## Changes

`get_output_root() `function and unit test.
Two realizations with `output_root `for demonstration. One realization in its original form, i.e., w/o `output_root` for demonstration

## Testing

1. Local run ngen in serial and parallel modes, and unit test

## Notes

-

## Todos

-

## Checklist

1.

- [x] Linux


include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -285,17 +289,20 @@ namespace realization {
...
-                           ? *output_root
-                           : *output_root + "/";
-                }
+                const std::string str = *output_root;
+                const char* dir = str.c_str();
 
-                return "./";
+                //use C++ system function to check if there is a dir match that defined in realization
+                struct stat sb;
+                if (stat(dir, &sb) == 0)
```
PhilMiller, line 299:
```suggestion
                if (stat(dir, &sb) == 0 && S_ISDIR(sb.mode))
```

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -285,17 +289,20 @@ namespace realization {
...
              */
-            std::string get_output_root() const noexcept {
+            std::string get_output_root() const {
                 const auto output_root = this->tree.get_optional<std::string>("output_root");
-                if (output_root != boost::none && *output_root != "") {
-                    // Check if the path ends with a trailing slash,
-                    // otherwise add it.
-                    return output_root->back() == '/'
-                           ? *output_root
-                           : *output_root + "/";
```
program--, line 295: For reference, this manipulation was added because the output file paths are constructed by concatenation, i.e.

https://github.com/NOAA-OWP/ngen/blob/0917f9b021e081f01d42966ad92ee6370f3eca54/src/core/HY_Features.cpp#L35

https://github.com/NOAA-OWP/ngen/blob/0917f9b021e081f01d42966ad92ee6370f3eca54/src/NGen.cpp#L443

Without checking/adding the trailing slash, or modifying the code at the call sites, we may have paths of the form `/some/output/dirnex-01_output.csv` instead of `/some/output/dir/nex-01_output.csv` for example.

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -285,17 +289,20 @@ namespace realization {
...
-                }
+                const std::string str = *output_root;
```
program--, line 294: This is potentially UB I believe, since `output_root` may be uninitialized. Need to keep the check on `boost::none` at least, or replace `this->tree.get_optional<std::string>` with a different `get*` call.

include/realizations/catchment/Formulation_Manager.hpp
stcui007: Implemented. mode->st_mode.

include/realizations/catchment/Formulation_Manager.hpp
program--: I'd probably just replace line 293 with this:

```cpp
std::string output_root = this->tree.get<std::string>("output_root");
```

Since this will throw if `output_root` doesn't exist, and removes the need for a separate check.

Phil's suggestion ensures the path references a directory, but it won't handle the trailing slash case.

stcui007: I have my proposed version ready adn tested. But will try this one.

stcui007: This approach doesn't work when `output_rout` doesn't have a "/" at the end. So I kept the check.

stcui007: The problem is when you do have a directory. Instead of writing to the spefied directory, it write to the project directory with the concatenation you pointed out.

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -285,17 +291,28 @@ namespace realization {
              * 
              * @return std::string of the output root directory
              */
-            std::string get_output_root() const noexcept {
+            std::string get_output_root() const {
                 const auto output_root = this->tree.get_optional<std::string>("output_root");
+                std::string str;
```
aaraney, line 296: I think this should fix your issue.
```suggestion
                std::string str = "./";
```

stcui007: You mean the failing tests or something else. I don't want to write to "./" though.

aaraney: I'm fairly certain the call to `stat` will fail if you provide it the empty string like you are right now in the case when `output_root` is not present. I think that is why the `Build_and_Run_Model` action is failing.

I think you should return `"./"` if `output_root` is not present (`boost::none`) like in the previous code. We can always assume that the directory `ngen` is run from, `./` , exists. 

One solution is:

```c++
                if (output_root != boost::none && *output_root != "") {
                    // Check if the path ends with a trailing slash,
                    // otherwise add it.
                    output_root->back() == '/'
                           ? *output_root
                           : *output_root + "/";

                     //use C system function to check if there is a dir match that defined in realization
                    struct stat sb;
                    if (stat(output_root->c_str(), &sb) == 0 && S_ISDIR(sb.st_mode))
                        return *output_root;
                    else {
                        throw std::runtime_error("output_root directory does not exist, please create one matching that in realization");
                    }
                }
                return "./";
```

Aside, I would think we would want to check that the calculated `output_root` can be written to by the current user. That is out of the scope of this PR and should be handled elsewhere if deemed necessary.

stcui007: (1) It looks like this code would return "./" if output_root directory does not exist.
(2) This code would seem not get a chance to throw based on the outer if, which is not what we want.
(3) This block of code is a bit tricky:
```
                    output_root->back() == '/'
                           ? *output_root
                           : *output_root + "/";
```
where does the value land?
(4) The failed test looks like come from this line in Example_model_run.yml:
     ` ./ngen data/catchment_data.geojson '' data/nexus_data.geojson '' data/example_realization_config.json`
It takes realization file from the master, not from the PR, right?
(5) In the latest realization file, I set output_root to "./", so it automatically exists.

aaraney: This code handles the output root for `ngen`, so if there is one in a realization config file under the key, `output_root`, that one will be used, otherwise, the default is to write to the directory where `ngen` was called, `"./"`.

I don't think you pushed an updated `data/example_realization_config.json` file. Looks like you pushed up `data/example_bmi_multi_realization_config.json` and `data/example_bmi_multi_realization_config_w_noah_pet_cfe.json`.

program--:
Based on the comments, I think this PR is partially solving #848 but needs a bit more discussion, since it feels like an XY problem.

There's 2 separate issues here:

(1) `output_root` fails to write outputs if the specified path is invalid (doesn't exist, isn't writable, etc.) as outlined in #848 -- There are two approaches to solving this I think: either create the directory path (i.e. emulate `mkdir -p`) if possible, or throw an error. This PR currently performs the latter via `stat`.

(2) If no `output_root` is specified, potentially lots of files will be outputted into the project directory since `output_root` defaults to `./`, as outline in #857. I don't think this is actually an issue as the solution is to just specify an `output_root` in the config file.

I think the real question for this PR as is: do we want `output_root` to be **optional** or **required**? Or rather, if `output_root` has an invalid path, should we try to make the directories or should we always throw an error?

stcui007:
The issue is, if we can still call it an issue, if someone runs a big job
and forgot to create the directory or entered wrong format in realization,
then when the job finishes, he ends up having a lot of problems. This PR
tries to stop that happening at the start.

It seems to me in the long run, requireing an output_root doesn't have any
downside. The decision needs to be made by Nels. Not sure if there is any
ripple effect to calibration if we change realization.

Maybe as you alluded to in (1). We could create  a
directory programmatically.

aaraney:
Great, summary @program--. I share the same feelings.

> I think the real question for this PR is: do we want `output_root` to be **optional** or **required**? Or rather, if `output_root` has a non-existent path, should we try to make the directories or should we always throw an error?

My preference would be that `output_root` stayed optional. For instance in calibration, we take advantage of the absence of `output_root` so that the output is written to where ever we call `ngen` from. Right, of course we could add functionality to account for this, but this is just my personal preference.

Likewise, if possible, emulating `mkdir -p` is desirable IMO. I would expect that a program would create the output directory I specified if it didn't already exist. For sure a matter of option, so I am going to ping @PhilMiller, @hellkite500, and @donaldwj to get their thoughts as well.

From a practicality stance, we are already including `sys/stat.h`, so we have access to `mkdir` but there are likely other issues that I might not be considering.

stcui007:
@Austin Raney - NOAA Affiliate ***@***.***>
You are right, I didn't. I thought that test used the multiple bmi
realization.

aaraney:
> The issue is, if we can still call it an issue, if someone runs a big job and forgot to create the directory or entered wrong format in realization, then when the job finishes, he ends up having a lot of problems. This PR tries to stop that happening at the start.

Yeah, that's a good point. Like you have started to do in this PR, transitioning our examples in `data` to use `output_root` I think is a great middle ground. Thoughts?

aaraney:
@stcui007, no worries!

stcui007:
@Austin Raney - NOAA Affiliate ***@***.***>
I assume you know quite bit about calibration. So changing the Examples
would not affect them? We probably need to ask Nels. At the same time, we
need to carefully check our code so it doesn't have effect on that.

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -10,6 +10,10 @@
 #include <dirent.h>
 #include <sys/stat.h>
 #include <regex>
+#include <sys/types.h>
+#include <sys/stat.h>
```
program--, line 14: Duplicate header, already included on line 11

```suggestion

```

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -271,7 +277,9 @@ namespace realization {
             }
 
             /**
-             * @brief Get the formatted output root
+             * @brief Get the formatted output root: check the existence of the output_root directory defined
+             * in realization. If true, return the directory name. Otherwise, throw a runtime error and 
+             * request the user to create that directory
```
program--, line 282:
```suggestion
             * @brief Get the formatted output root: check the existence of the output_root directory defined
             * in realization. If true, return the directory name. Otherwise, try to create the directory
             * or throw an error on failure.
```

The documentation is slightly out of date from a previous commit

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -285,16 +293,29 @@ namespace realization {
...
 
+                    const char* dir = str.c_str();
+
+                    //use C++ system function to check if there is a dir match that defined in realization
+                    struct stat sb;
+                    if (stat(dir, &sb) == 0 && S_ISDIR(sb.st_mode))
+                        return dir;
+                    else {
+                        //throw std::runtime_error("output_root directory does not exist, please create one matching that in realization");
+                        mkdir(dir, 0755);
```
program--, line 313: Two things for `mkdir` that I think we need to ensure.

1. What happens when `mkdir` fails? (i.e. the directory where the requested path needs to be created isn't writable). I think we should check the [return code](https://pubs.opengroup.org/onlinepubs/007904875/functions/mkdir.html) and propagate the error code as a `std::runtime_error`. Additionally, you could skip the `stat` check in favor of checking `mkdir`'s return code for `EEXISTS`, but that's not super necessary.

2. What happens when the requested path requires multiple levels of directories? i.e. given a path `/path/to/ngen` with an `output_root` set to `/path/to/ngen/a/b`, does `mkdir` successfully create `a/` and `a/b/`? I suspect this will fail, so we might want to create a test case to check that.

program-- (changes requested):

include/realizations/catchment/Formulation_Manager.hpp
program--: Note: doesn't need to be exactly these changes, I wrote this with respect to my comment on `mkdir`

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -22,6 +26,8 @@
 #include "realizations/config/config.hpp"
 #include "realizations/config/layer.hpp"
 
+#include<iostream>
```
program--, line 29: One other thing I almost missed, is this needed? I didn't see anything specifically that required `<iostream>`

program--: Nevermind, I didn't realize this was `Formulation_Manager.hpp` and I know there's at least `std::cerr` outputs in here, so this is fine (especially since it probably includes it transitively anyways)

include/realizations/catchment/Formulation_Manager.hpp
stcui007: Yeah. I for got about it. Good catch.

include/realizations/catchment/Formulation_Manager.hpp
stcui007: (1) We probably should put in a guard there when mkdir fails. Throwing a `std::runtime_error` would be a good option. I am reasonably happy with the stat check the way it is, unless you feel strongly about it.

(2) My initial research suggests `mkdir `does have the capability to generate multi-level directories. However, for our purpose, do we really have a need for that?

program--: For (1), as long as we throw an error, that's fine.

For (2), where are you seeing that `mkdir` can create multi-level directories? 

This example results in an error:
```c
#include <stdio.h>
#include <string.h>
#include <sys/stat.h>
#include <errno.h>
#include <unistd.h>

void test_path(const char* path) {
    errno = 0;
    int result_code = mkdir(path, 0775);
    if (result_code == -1)
        printf("[%s] error: %s\n", path, strerror(errno));

    unlink(path);
}

int main(int argc, const char* argv[]) {
    // This passes (unless current directory is not writable)
    test_path("example_path");

    // This fails unless 'example_nested_path/' is created
    test_path("example_nested_path/subdir");
    
    // This fails unless '/opt/ngen/outputs/' is created
    test_path("/opt/ngen/outputs/22July2024_1");
    return 0;
}
```

For your second point in (2), the purpose is for consumers. It's a reasonable assumption that if `output_root` is set to a directory, and it can create a top-level directory, then it should be able to create multi-level directories. However, worst-case, as long as we're throwing the errors, then the consumer will be informed.

include/realizations/catchment/Formulation_Manager.hpp
stcui007: <iostream> is not needed. It was used dusing testing.

stcui007:
This link:
https://stackoverflow.com/questions/55994752/adding-multiple-directories-to-mkdir

stcui007:
A more careful reading of the site suggests it might have to be done
recursively.

stcui007:
However, worst-case, as long as we're throwing the errors, then the
consumer will be informed.

I agree. If users are warned, then create a multi-level directory manually,
it should always work.

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -285,16 +290,31 @@ namespace realization {
...
+                    //use C++ system function to check if there is a dir match that defined in realization
+                    struct stat sb;
+                    if (stat(dir, &sb) == 0 && S_ISDIR(sb.st_mode))
+                        return dir;
+                    else {
+                        int result = mkdir(dir, 0755);      
+                        if (result == 0)
+                            return dir;
+                        else
+                            throw std::runtime_error("mkdir " + std::string(dir) + " failed, please create " + std::string(dir));
```
program--, line 313: Not critical, but I think we should make this error message more informational by checking what `errno` is set to (might need to include `<cerrno>`). Other than that, it looks okay.

```suggestion
                            throw std::runtime_error("failed to create directory '" + str + "': " + strerror(errno));
```

stcui007: Implemented as suggested.

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -9,7 +9,11 @@
 #include <functional>
 #include <dirent.h>
 #include <sys/stat.h>
+#include <errno.h>
```
program--, line 12:
```suggestion
#include <cerrno>
```

nitpick: `<errno.h>` is for the C standard library, `<cerrno>` is the C++ counterpart.

program-- (commented):
@aaraney If you could do a final review just to see if this satisfies #848 that'd be great! Otherwise, this looks just about done

include/realizations/catchment/Formulation_Manager.hpp
program--: Oops, I think it actually should be `std::strerror`, my bad.

stcui007:
miracurolly, it actually worked. At least no complaint.

include/realizations/catchment/Formulation_Manager.hpp
stcui007: Missed this one.

stcui007:
Changed the included header.

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -285,16 +291,32 @@ namespace realization {
...
                            : *output_root + "/";
-                }
 
+                    const char* dir = str.c_str();
+
+                    //use C++ system function to check if there is a dir match that defined in realization
+                    struct stat sb;
+                    if (stat(dir, &sb) == 0 && S_ISDIR(sb.st_mode))
+                        return dir;
+                    else {
```
PhilMiller, line 309:
```suggestion
                    if (stat(dir, &sb) == 0 && S_ISDIR(sb.st_mode)) {
                        return dir;
                    } else {
```

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -285,16 +291,32 @@ namespace realization {
...
                            ? *output_root
                            : *output_root + "/";
-                }
 
+                    const char* dir = str.c_str();
+
+                    //use C++ system function to check if there is a dir match that defined in realization
+                    struct stat sb;
+                    if (stat(dir, &sb) == 0 && S_ISDIR(sb.st_mode))
+                        return dir;
```
PhilMiller, line 308:
```suggestion
                        return str;
```
We've already made the string. There's no reason to re-construct a copy like this.

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -285,16 +291,32 @@ namespace realization {
...
+                    else {
+                        errno = 0;
+                        int result = mkdir(dir, 0755);      
+                        if (result == 0)
+                            return dir;
```
PhilMiller, line 313:
```suggestion
                            return str;
```

PhilMiller:
The logic looks fine to me.

include/realizations/catchment/Formulation_Manager.hpp
stcui007: Yeah. It is also more consistent with the function definition.

hellkite500 (approved):

hellkite500:
Looks like this adds some reasonable QoL logic to help with output management. Thanks!

PhilMiller (approved):

PR #881, opened 2024-09-11
Implement `simulation_clock` type

### LINKED ISSUE #858: Implement a simulation clock type
Currently, across Data Providers and BMI instances, we're primarily using C time constructs.

<details><summary><em>Example</em></summary>

https://github.com/NOAA-OWP/ngen/blob/9fc3182f908efbf58a88b019df516932f5e85ec2/include/forcing/DataProvider.hpp#L47-L54

https://github.com/NOAA-OWP/ngen/blob/9fc3182f908efbf58a88b019df516932f5e85ec2/include/simulation_time/Simulation_Time.hpp#L171-L173

</details>

However, in some recent changes, we've started using `std::chrono` types.

<details><summary><em>Example</em></summary>

https://github.com/NOAA-OWP/ngen/blob/9fc3182f908efbf58a88b019df516932f5e85ec2/include/forcing/ForcingsEngineDataProvider.hpp#L208-L214

</details>

One of the current issues with `std::chrono` clock types is that they either represent Unix time or relative time (though not clearly). A potential solution for this is to define our own clock type[^1], `simulation_clock`, as discussed [here](https://github.com/NOAA-OWP/ngen/pull/851#discussion_r1664731024).

A benefit is that this would allow us to use `std::chrono` types while maintaining control of the simulation ticks however we see fit.

[^1]: https://en.cppreference.com/w/cpp/named_req/Clock

program--:
This PR closes #858 by implementing a `simulation_clock` type that implements the [_TrivialClock_ named requirement](https://en.cppreference.com/w/cpp/named_req/TrivialClock). However, unlike the convention `std::chrono` clocks, this implementation is a controlled clock meaning that `simulation_clock::now()` does not change unless `simulation_clock::tick()` is called.

## Additions

- `ngen::simulation_clock` type.
- Corresponding unit tests.

## Checklist

### DIFF
--- a/include/simulation_time/simulation_clock.hpp
+++ b/include/simulation_time/simulation_clock.hpp
@@ -1,0 +1,205 @@
+#pragma once
+
+#include <chrono>
+#include <stdexcept>
+#include <utility>
+
+namespace ngen {
+
+//! Simulation Clock
+//!
+//! Represents simulation ticks monotonically increasing from zero, and
+//! provides conversion utilities between simulation ticks and
+//! calendar time.
+//!
+//! Implements named requirements: Clock (https://en.cppreference.com/w/cpp/named_req/Clock)
+//!
+struct simulation_clock final
+{
+    /* Clock named requirements */
+
+    using rep = long; // TODO: double vs std::int64_t
+    using period = std::ratio<1>;
+    using duration = std::chrono::duration<rep, period>;
+    using time_point = std::chrono::time_point<simulation_clock, duration>;
+    static constexpr bool is_steady = false;
+
+    //! Current Simulation Time
+    //! \returns The current point in time the simulation is at.
+    //! \post Return value must have a non-negative representation value..
+    static time_point now() noexcept;
+
+    /* simulation_clock-specific types/interface */
+
+    // explicit simulation time type aliases //
+    using simulation_rep = rep;
+    using simulation_period = period;
+    using simulation_duration = duration;
+    using simulation_time_point = time_point;
+
+    // explicit calendar time type aliases //
+    using calendar_clock = std::chrono::system_clock;
+    using calendar_rep = calendar_clock::rep;
+    using calendar_period = calendar_clock::period;
+    using calendar_duration = calendar_clock::duration;
+    using calendar_time_point = calendar_clock::time_point;
+
+    // simulation-calendar conversion functions //
+
+    //! Convert a calendar time point to a simulation time point.
+    //! \param calendar_time Calendar time point.
+    //! \returns The corresponding simulation time point.
+    //! \pre simulation_clock::set_epoch must have been called prior to calling this.
+    static simulation_time_point from_calendar(const calendar_time_point& calendar_time);
+
+    //! Convert a simulation time point to a calendar time point.
+    //! \param sim_time Simulation time point.
+    //! \returns The corresponding calendar time point.
+    //! \pre simulation_clock::set_epoch must have been called prior to calling this.
+    static calendar_time_point to_calendar(const simulation_time_point& sim_time);
+
+    //! Convert a std::time_t value to a simulation time point.
+    //! \see simulation_clock::from_calendar
+    //! \param time Unix epoch-based time value
+    //! \returns The corresponding simulation time point.
+    //! \pre simulation_clock::set_epoch must have been called prior to calling this.
+    static simulation_time_point from_time_t(std::time_t t);
+
+    //! Convert a simulation time point to a std::time_t value.
+    //! \see simulation_clock::to_calendar
+    //! \param sim_time Simulation time point.
+    //! \returns The corresponding std::time_t value.
+    //! \pre simulation_clock::set_epoch must have been called prior to calling this.
+    static std::time_t to_time_t(const simulation_time_point& sim_time);
+
+    // simulation management functions //
+
+    //! Unitialized epoch value, aka the epoch of typename simulation_clock::calendar_clock.
+    static constexpr auto uninitialized_epoch = calendar_time_point{};
+
+    //! Get the simulation epoch.
+    //! \see simulation_clock::calendar_time_point
+    //! \see simulation_clock::set_epoch
+    //! \returns
+    //! Calendar time point representing the simulation start date/time.
+    //! If the epoch has not been set, then simulation_clock::uninitialized_epoch
+    //! is returned.
+    static calendar_time_point epoch() noexcept;
+
+    //! Set the simulation epoch.
+    //! \note This function must only be called once.
+    //! \param calendar_time Calendar time set as the epoch.
+    //! \throws std::logic_error Thrown when this function is called more than once.
+    static void set_epoch(calendar_time_point calendar_time);
+
+    //! Check if the simulation epoch has been set.
+    //! \returns true when the epoch is set, and false otherwise.
+    static bool has_epoch() noexcept;
+
+    //! Forward the simulation by N ticks.
+    //! \param steps Number of steps to tick.
+    //! \note Ticking does *not* require that simulation_clock::set_epoch has been called.
+    //! \see simulation_clock::period
+    //! \pre `steps` must be non-negative.
+    //! \throws std::domain_error Throws when the precondition is violated.
+    static void tick(rep steps = 1);
+
+    //! Forward the simulation by some duration.
+    //! \param duration Duration to move simulation forward.
+    //! \note Ticking does *not* require that simulation_clock::set_epoch has been called.
+    //! \pre `duration` must be non-negative.
+    //! \throws std::domain_error Throws when the precondition is violated.
+    template<typename Rep, typename Period>
+    static void tick(const std::chrono::duration<Rep, Period>& duration);
+
+  private:
+    //! Simulation epoch, the calendar start time of the simulation.
+    //! \returns Mutable reference to the internal calendar epoch.
+    static calendar_time_point& internal_epoch();
+
+    //! Current simulation tick.
+    //! \returns Mutable reference to the current internal simulation tick.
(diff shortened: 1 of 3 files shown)


include/simulation_time/simulation_clock.hpp
```diff
@@ -0,0 +1,205 @@
...
+    //! \see simulation_clock::to_calendar
+    //! \param sim_time Simulation time point.
+    //! \returns The corresponding std::time_t value.
+    //! \pre simulation_clock::set_epoch must have been called prior to calling this.
+    static std::time_t to_time_t(const simulation_time_point& sim_time);
+
+    // simulation management functions //
+
+    //! Unitialized epoch value, aka the epoch of typename simulation_clock::calendar_clock.
+    static constexpr auto uninitialized_epoch = calendar_time_point{};
```
program--, line 78: `static constexpr` isn't implicitly inline before C++17, it seems? https://en.cppreference.com/w/cpp/language/constexpr

Alternatively, this could be a `static constexpr` member function.

include/simulation_time/simulation_clock.hpp
```diff
@@ -0,0 +1,205 @@
...
+    //! \returns true when the epoch is set, and false otherwise.
+    static bool has_epoch() noexcept;
+
+    //! Forward the simulation by N ticks.
+    //! \param steps Number of steps to tick.
+    //! \note Ticking does *not* require that simulation_clock::set_epoch has been called.
+    //! \see simulation_clock::period
+    //! \pre `steps` must be non-negative.
+    //! \throws std::domain_error Throws when the precondition is violated.
+    static void tick(rep steps = 1);
```
program--, line 105: Just a thought, do we need thread-safety guarantees?

> I don't think so...

include/simulation_time/simulation_clock.hpp
```diff
@@ -0,0 +1,205 @@
...
+
+inline simulation_clock::time_point simulation_clock::from_time_t(std::time_t t)
+{
+    return simulation_clock::from_calendar(calendar_clock::from_time_t(t));
+}
+
+inline std::time_t simulation_clock::to_time_t(const simulation_time_point& sim_time)
+{
+    return calendar_clock::to_time_t(simulation_clock::to_calendar(sim_time));
+}
```
program--, line 165: TODO: these are missing assertions for the documented precondition.

include/simulation_time/simulation_clock.hpp
program--: Converted to a `static constexpr` member function.

include/simulation_time/simulation_clock.hpp
PhilMiller: No, I don't think so. Any multi-threading we eventually, hypothetically do should be contained within a single simulation time-step, with suitable synchronization around ticks.

program--: Cool, that's in line with what I was thinking.

PhilMiller:
Out of curiosity, did you pick up the term "controlled clock" from somewhere in particular? It seems apt, and sounds like a standard-ese-ish coinage, I've just never encountered it before. A quick Google search didn't turn it up.

program--:
> Out of curiosity, did you pick up the term "controlled clock" from somewhere in particular? ...

No I just made that up 😄

test/simulation_time/simulation_clock_Test.cpp
```diff
@@ -0,0 +1,47 @@
...
+TEST(SimulationClockTest, TestConfiguration)
+{
+    ASSERT_FALSE(clock_type::has_epoch());
+    ASSERT_NO_THROW(clock_type::set_epoch(epoch_calendar));
+    ASSERT_TRUE(clock_type::has_epoch());
+    ASSERT_THROW(clock_type::set_epoch(epoch_calendar), std::logic_error);
+    EXPECT_EQ(clock_type::epoch(), epoch_calendar);
+}
+
+TEST(SimulationClockTest, TestConversion)
```
PhilMiller, line 20: It looks like this test assumes `TestConfiguration` has already run before it, setting the `simulation_clock`'s epoch. With tests run in parallel, is that dependency still honored? I don't know how concurrent gtest can get, and whether our usage of it otherwise allows for that concurrency.

In general, it's not great to have inter-test dependencies. In this case, the easy solution may be simply to merge `TestConfiguration` and `TestConversion` into a single test. Most of `TestTicking` is fine, until the very last `EXPECT_EQ` that references `epoch_calendar. So, maybe merge them all into a single sequence?

include/simulation_time/simulation_clock.hpp
```diff
@@ -0,0 +1,219 @@
...
+
+namespace ngen {
+
+//! Simulation Clock
+//!
+//! Represents simulation ticks monotonically increasing from zero, and
+//! provides conversion utilities between simulation ticks and
+//! calendar time.
+//!
+//! Implements named requirements: Clock (https://en.cppreference.com/w/cpp/named_req/Clock)
```
PhilMiller, line 16: `Clock` or `TrivialClock`? This disagrees with the PR description.

include/simulation_time/simulation_clock.hpp
```diff
@@ -0,0 +1,219 @@
...
+{
+    static calendar_time_point epoch = simulation_clock::uninitialized_epoch();
+    return epoch;
+}
+
+inline simulation_clock::time_point& simulation_clock::internal_now()
+{
+    static simulation_time_point now{}; // Default initialize to zero value
+    return now;
+}
```
PhilMiller, line 143: Does this do the right thing with linking multiple objects that `#include <simulation_clock.hpp>`, or do we need to put this and `internal_epoch` (and presumably the rest of the implementations) in a separate `.cpp` file?

Is there some benefit to inlining all of this that's worth more than not having to think about the above question?

include/simulation_time/simulation_clock.hpp
```diff
@@ -0,0 +1,219 @@
...
+//! provides conversion utilities between simulation ticks and
+//! calendar time.
+//!
+//! Implements named requirements: Clock (https://en.cppreference.com/w/cpp/named_req/Clock)
+//!
+struct simulation_clock final
+{
+    /* Clock named requirements */
+
+    using rep = std::int64_t; // TODO: double vs std::int64_t
```
PhilMiller, line 22: I don't think we're likely to run into sub-second time precision, so `int64_t` is probably fine.

include/simulation_time/simulation_clock.hpp
```diff
@@ -0,0 +1,219 @@
...
+    using period = std::ratio<1>;
+    using duration = std::chrono::duration<rep, period>;
+    using time_point = std::chrono::time_point<simulation_clock, duration>;
+    static constexpr bool is_steady = false;
+
+    //! Current Simulation Time
+    //! \returns The current point in time the simulation is at.
```
PhilMiller, line 29: This is almost philosophical, but in a sense, the description of the returned value is premised on the expected use by callers of `now()` and `tick()` and the epoch routines.

To be really strict, it's really more like "The accumulation of argument values from preceding calls to `tick()`"

include/simulation_time/simulation_clock.hpp
```diff
@@ -0,0 +1,219 @@
...
+    //! \returns true when the epoch is set, and false otherwise.
+    static bool has_epoch() noexcept;
+
+    //! Forward the simulation by N ticks.
+    //! \param steps Number of steps to tick.
+    //! \note Ticking does *not* require that simulation_clock::set_epoch has been called.
+    //! \see simulation_clock::period
+    //! \pre `steps` must be non-negative.
+    //! \throws std::domain_error Throws when the precondition is violated.
+    static void tick(rep steps = 1);
```
PhilMiller, line 106: Do we want this overload `tick(rep)` at all? When do we want calls that aren't in terms of a `duration` type?

If we do keep it, I think the parameter name `steps` is wrong. It's unit `seconds`, while the model time step is typically going to be something else. Is this maybe a hold-over from an earlier version you drafted with a `time_step` member variable that would have been set? Given the inherent `static` nature of `Clock`, I'd lean against bringing that back.

include/simulation_time/simulation_clock.hpp
```diff
@@ -0,0 +1,219 @@
...
+    static simulation_time_point now{}; // Default initialize to zero value
+    return now;
+}
+
+inline simulation_clock::time_point simulation_clock::now() noexcept
+{
+    return simulation_clock::internal_now();
+}
+
+inline simulation_clock::time_point simulation_clock::from_calendar(const calendar_time_point& calendar_time)
```
PhilMiller, line 150:
```suggestion
inline simulation_clock::simulation_time_point simulation_clock::from_calendar(const calendar_time_point& calendar_time)
```

include/simulation_time/simulation_clock.hpp
program--: Hmmm that's a good point, I don't think we need this. I added this prior to really thinking about how duration casting could better represent arbitrary ticking

include/simulation_time/simulation_clock.hpp
program--: Oh good catch, yeah this might be incorrect. No actual reason for it being header-only, haha

include/simulation_time/simulation_clock.hpp
program--: Both, though the intention is to implement `Clock` -- it just happens to also implement `TrivialClock`.

PhilMiller: I noticed after posting this question that `TrivialClock` is actually the more demanding of the two concepts. I think it's fine to just reference `Clock` in the code, to avoid confusion, though maybe mention it in the PR description.

PR #906, opened 2025-08-11
Fix bad uses of `sizeof(char[])/sizeof(char*)`

PhilMiller:
In both cases, we're ultimately getting the length `6` of the literal string `"{{id}}"`, without `'\0'` terminator

In the case where the literal string is passed to `sizeof`, it's obtusely correct, because that evaluates the literal `char[]`, which takes serious language-lawyering to recognize. It's also very fragile to array-to-pointer-decay if someone attempts to refactor the code in some way, which would leave it over by one.

In the case where `std::string pattern` is a function parameter and the code says `sizeof(pattern.c_string()) - 2`, it's only managed to operate correctly because all of the current call sites pass an argument of `"{{id}}"`, and it's only tested on 64-bit systems.

## Changes

- Replace two instances of `sizeof(<string literal>)` with `std::string::size()` to be more durably correct

## Todos

- @robertbartel offered to add tests for this as appropriate

## Checklist

### DIFF
--- a/include/realizations/catchment/Catchment_Formulation.hpp
+++ b/include/realizations/catchment/Catchment_Formulation.hpp
@@ -53,7 +53,7 @@ namespace realization {
 
             if (id_index != std::string::npos) {
                 do {
-                    value = value.replace(id_index, sizeof(pattern.c_str()) - 2, replacement);
+                    value = value.replace(id_index, pattern.size(), replacement);
                     id_index = value.find(pattern);
                 } while (id_index != std::string::npos);
 
--- a/include/realizations/catchment/Catchment_Formulation.hpp
+++ b/include/realizations/catchment/Catchment_Formulation.hpp
@@ -53,7 +53,7 @@ namespace realization {
 
             if (id_index != std::string::npos) {
                 do {
-                    value = value.replace(id_index, pattern.size(), replacement);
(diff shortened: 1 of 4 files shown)


robertbartel:
Good find @PhilMiller.  

I tend to think "yes" regarding @christophertubbs's questions 1 and 2 above.  Regarding question 3, eh, maybe.  I do think we should (more fully) abstract this logic out to a single reusable utility function somewhere.  I feel like I was probably starting to do this by adding the `CatchmentFormulation.config_pattern_substitution` function, but didn't really position it so it could be plugged in everywhere it was needed (and clearly I wasn't thinking hard enough about things when I added it).  If/after we do that, it may be more apparent whether and where to define strings constants like `{{id}}`.

Also, even if the above isn't done, clearly some additional tests are needed, at minimum to ensure `CatchmentFormulation.config_pattern_substitution` behaves robustly enough.  I will work on contributing those.

PhilMiller:
I've addressed Chris's points 1 and 2. I think 3 is out of scope of this change, whose goal is to replace code that happens to work, but isn't really 'correct'.

PhilMiller:
This'll be rebased on #955, after finding that the impacted functionality was not covered by tests, and was in fact broken.

PR #909, opened 2025-09-17
Implementing a BMI mass balance checking protocol

hellkite500:
This feature allows the model engine to attempt dynamic mass balance introspection of BMI modules via a specified `protocol` which dictates the BMI variables names representing `cummulative` mass balance variables which can be inspected at any point after `bmi_intialize()` is called, and it is assumed that these variables balance mass, e.g.

```
balance = mass_in - mass_out - mass_stored - mass_leaked
```

When `balance` is less than a configurable threshold (provided as a formulation parameter in the realization file) then this either prints a warning or throws an error, based on the configuration.

## Additions

Configurable mass balance protocol implementation.  The follow object definition in a formulations `params` object is used to configure the mass balance implementation.

```json
"mass_balance":  {
    "fatal": true, 
    "tolerance": 1.0e-17,
    "check": true,
    "frequency": -1
}
```
Where
- fatal, bool
    When true, a fatal exception is thrown when mass balance tolerance is violated.
    When false, a mass balance warning capturing details of the error is printed, but the simulation continues.
- tolerance, float
    How close to 0 the mass balance can get before it is considered an error.
- check, bool
    Toggle the mass balance checking on or off
- frequency, int
    How often, in number of timesteps, to check the mass balance in the model engine.  If frequency is set to -1, then it will only be checked at the *last* time step.

## Removals

-

## Changes

- 

## Testing

1. Tests added for BMI_C formulation and BMI_Multi formulation

## Notes

- THIS IS AN OPTIONAL PROTOCOL.  If a BMI model doesn't implement the protocol, it is detected upon initialization and a a message provided, but mass balance checks will not be attempted.

## Todos

- Some consideration of unit conversion could be added, right now a sanity check exists which causes an integration error
to occur if the mass balance units aren't all the same, but no conversion is attempted.

## Checklist

- [x] Linux
- [x] MacOS


src/utilities/bmi/mass_balance.cpp
```diff
@@ -0,0 +1,137 @@
...
+        units.reserve(4);
+        try{
+            for(const auto& name : 
+                {INPUT_MASS_NAME, OUTPUT_MASS_NAME, STORED_MASS_NAME, LEAKED_MASS_NAME}
+            ){
+                model->GetValue(name, &mass_var);
+                units.push_back( model->GetVarUnits(name) );
+            }
+            //Compare all other units to the first one (+1)
+            if( std::equal( units.begin()+1, units.end(), units.begin() ) ) {
```
robertbartel, line 80: Unless I've missed something (entirely possible; if so, sorry), there are no explicit units defined anywhere for `tolerance`, and they are simply taken to be whatever units that particular model is using.  This seems like an easy place for the wrong assumption to be made.  We might want to make tolerance units an explicit part of the config and protocol, and then check those here as well (though perhaps slightly separately so a distinctive error message can be used for that).

Not unrelated, a small configuration example and some discussion within _doc/BMI_MODELS.md_ would probably be helpful to include as part of the overall changes.

include/utilities/bmi/protocol.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+  protected:
+    
+    /**
+     * @brief Constructor for subclasses to create NgenBmiProtocol objects
+     * 
+     * @param model A shared pointer to a Bmi_Adapter object which should be
+     * initialized before being passed to this constructor.
+     */
+    NgenBmiProtocol(std::shared_ptr<models::bmi::Bmi_Adapter> model): model(model){}
```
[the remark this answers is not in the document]
hellkite500: Definitely worth discussing.  I started down that path a little bit initially but eventually landed here, especially when trying to think about the lifecycle and semantics of effectively "optional" mechanics and how/when those should be reported.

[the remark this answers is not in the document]
aaraney: Yeah lifecycle semantics could be a little tricky to handle if there is a single interface where we don't differentiate between different variants. Just thinking out loud, if we could differentiate between `NgenBmiProtocol` lifecycle variants we could just use the `NgenBmiProtocol`'s constructor to handle initialization.

include/realizations/catchment/Bmi_Module_Formulation.hpp
```diff
@@ -419,6 +426,7 @@ namespace realization {
         int next_time_step_index = 0;
 
     private:
+        models::bmi::protocols::NgenBmiProtocols bmi_protocols;
```
aaraney, line 429: It should be easy to change in the future. But, I can't get it out of my head wanting to associate a single instance of a `NgenBmiProtocol` with multiple `Formulation` instances. I think expanding on what you have with `models::bmi::protocols::NgenBmiProtocols` will make that possible. Im just leaving my thoughts along the way :).

include/realizations/catchment/Bmi_Multi_Formulation.hpp
```diff
@@ -42,6 +42,15 @@ namespace realization {
 
         virtual ~Bmi_Multi_Formulation() {};
 
+        virtual void check_mass_balance(const int& iteration, const int& total_steps, const std::string& timestamp) const final {
+            for( const auto &module : modules ) {
+                // TODO may need to check on outputs form each module indepdently???
+                // Right now, the assumption is that if each component is mass balanced
+                // then the entire formulation is mass balanced
+                module->check_mass_balance(iteration, total_steps, timestamp);
```
aaraney, line 50: My vote would be not to support the "check on outputs form each module indepdently" case. You should be able to configure a mass balance check for each inner module in a multi-bmi to accomplish that, yeah?

include/realizations/catchment/Bmi_Module_Formulation.hpp
```diff
@@ -250,6 +251,12 @@ namespace realization {
         const std::vector<std::string> get_bmi_input_variables() const override;
         const std::vector<std::string> get_bmi_output_variables() const override;
 
+        virtual void check_mass_balance(const int& iteration, const int& total_steps, const std::string& timestamp) const override {
```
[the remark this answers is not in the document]
hellkite500: This sounds like a `version 1` functionality 😄

aaraney:
Just adding to the above description. I'm probably off a little here, so please correct me where i'm wrong :).

For a bmi module to comply with the mass balance protocol a module must expose the following "variables" to `ngen`.

```
# "name" -- type
"ngen::mass_in" -- double
"ngen::mass_out" -- double
"ngen::mass_stored" -- double
"ngen::mass_leaked" -- double
```

These variables must be available to `ngen` over the following bmi interfaces. It is **not required** for these "variables" to be exposed over `get_output_var_names`.

```
get_value
get_value_ptr
get_var_units
get_var_type
```

`ngen` will query the module for the following names via `get_value` then `get_var_units` and use the output return value (`BMI_SUCCESS` / `BMI_FAILURE`) to determine if the module implements the protocol.

src/utilities/bmi/mass_balance.cpp
```diff
@@ -0,0 +1,137 @@
...
+    //Ensure the model is capable of mass balance using the protocol
+    check_support();
+
+    //now check if the user has requested to use mass balance
+    auto protocol_it = properties.find(CONFIGURATION_KEY);
+    if ( is_supported && protocol_it != properties.end() ) {
+        geojson::PropertyMap mass_bal = protocol_it->second.get_values();
+
+        auto _it = mass_bal.find(TOLERANCE_KEY);
+        if( _it != mass_bal.end() ) tolerance = _it->second.as_real_number();
```
aaraney, line 112: Should we warn on the cases where this is `NaN` or `0` since that would effectively disable the protocol?

PhilMiller: I think 0 is a perfectly fine tolerance to configure, if one expects the subject model to be numerically exact. It's hard to build that way, but not impossible.

include/realizations/catchment/Bmi_Module_Formulation.hpp
aaraney: Ah I remember my thought. This relates to what we talked about below when differentiating between "stateless" and "stateful" protocols instances. My thinking is there is no reason for multiple instances of a "stateless" protocol. We should be able to reuse that instance for all formulation instances. I think we can shove this down into the implementation of `NgenBmiProtocols` if we ever go that route.

include/utilities/bmi/protocol.hpp
```diff
@@ -0,0 +1,208 @@
...
+
+enum class Error{
+    UNITIALIZED_MODEL,
+    UNSUPPORTED_PROTOCOL,
+    INTEGRATION_ERROR,
+    PROTOCOL_ERROR,
+    PROTOCOL_WARNING
+};
+
+class ProtocolError: public std::exception {
```
aaraney, line 46: @PhilMiller, can you take a closer look at this class? Specifically, object construction / destruction and the `what()` method. This is getting into c++ territory I would like to phone an expert to look at :).

include/utilities/bmi/protocol.hpp
```diff
@@ -0,0 +1,88 @@
...
+     * @brief The Bmi_Adapter object used by the protocol
+     * 
+     */
+    std::shared_ptr<models::bmi::Bmi_Adapter> model;
+
+    /**
+     * @brief Whether the protocol is supported by the model
+     * 
+     */
+    bool is_supported = false;
```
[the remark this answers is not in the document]
aaraney: I should have provided a little more context. My thinking was by including `is_supported` here removes a subtype's control over its memory layout and alignment. For example if a subtype adds members larger than 7 bytes total, lets say a pointer, the cost of `is_supported` is 8 bytes b.c. padding. Just wasn't sure if what we gain here was worth the price?

src/utilities/bmi/mass_balance.cpp
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+    }
+    return {};
+}
+
+auto NgenMassBalance::check_support(const ModelPtr& model) -> expected<void, ProtocolError> {
+    if (model->is_model_initialized()) {
+        double mass_var;
+        std::vector<std::string> units;
```
aaraney, line 97: Not required, but I think you could drop the collection and do the unit check below as we iterate. Alternatively, what about just switching to a `std::array` instead?

src/utilities/bmi/mass_balance.cpp
aaraney: thoughts here, @hellkite500?

hellkite500: After some discussion, I think we do need to check/handle NaN here.

src/utilities/bmi/mass_balance.cpp
hellkite500: This could simplify to two strings, `prev` and `curr` and a comparison for each iteration that the current unit is the same as the previous, and error when not.  I think that would even allow something like checking (via associative properties) if units are convertible when they don't match.  That was one case I had in mind when structuring this way, and trying to leave things in a state which allows further and/or different validation in the future.  I'm not sure its super impactful in this case, and may actually be ultimately more efficient to hold all 4 strings and iterate them as opposed to overwriting `prev`, `current` each iteration.

aaraney (dismissed):

aaraney (approved):

.github/workflows/test_and_validate.yml
```diff
@@ -272,12 +273,43 @@ jobs:
       - name: Run Unit Tests
         run: |
           . .venv/bin/activate
+          export ASAN_OPTIONS=${ASAN_OPTIONS}:detect_odr_violation=0
```
aaraney: See 08d8379db5dae1d362ef7547b6de94718525ccbb for explanation.

src/utilities/bmi/mass_balance.cpp
```diff
@@ -0,0 +1,194 @@
...
+            )
+        );
+    }
+    return {};
+}
+
+auto NgenMassBalance::initialize(const ModelPtr& model, const Properties& properties) -> expected<void, ProtocolError>
+{
+    //Ensure the model is capable of mass balance using the protocol
+    check_support(model).or_else( error_or_warning );
```
aaraney: Coming back to this, I think we should check model support only after reading the configuration. Otherwise, this will raise an exception for models that don't support the mass balance checker _and_ aren't configured.

PR #913, opened 2025-10-27
Ci updates

hellkite500:
Pulled from #909 into its own PR in order to better address some address sanitizer failures on the updated macos-15 runners. I'll rebase once that PR lands.

### DIFF
--- a/.github/actions/ngen-build/action.yaml
+++ b/.github/actions/ngen-build/action.yaml
@@ -196,6 +196,12 @@ runs:
             export BOOST_ROOT="$(pwd)/boost_1_86_0"
             export CFLAGS="-fsanitize=address -O1 -g -fno-omit-frame-pointer -Werror"
             export CXXFLAGS="-fsanitize=address -O1 -g -fno-omit-frame-pointer -pedantic-errors -Werror -Wpessimizing-move -Wparentheses -Wrange-loop-construct -Wsuggest-override"
+            if [ ${{ runner.os }} == 'macOS' ]
+            then
+              echo "fun:PyType_FromMetaclass" > /tmp/asan_ignore.txt
+              export CFLAGS="$CFLAGS -fsanitize-ignorelist=/tmp/asan_ignore.txt"
+              export CXXFLAGS="$CXXFLAGS -fsanitize-ignorelist=/tmp/asan_ignore.txt"
+            fi
             . .venv/bin/activate
             [ ! -d "$BOOST_ROOT" ] && echo "Error: no Boost root found at $BOOST_ROOT" && exit 1
             echo "Cmake Version:"


PhilMiller:
In my experience, ASan is almost certainly correct that there is an error, and that it is occurring during execution of the statement described by the stack trace. Thus, it's probably inappropriate to suppress it, and digging into the ASan implementation is likely a misguided rabbit hole.

I'd recommend focusing on trying to find out the name of the BMI variable being queried for purposes of setting, such as by printing it to standard error like ASan writes its notification. Alternately, just audit all of the variables that you expect to go through that code path, and make sure the BMI metadata functions and the tables they're relying on are correct for them.

hellkite500:
> In my experience, ASan is almost certainly correct that there is an error, and that it is occurring during execution of the statement described by the stack trace. Thus, it's probably inappropriate to suppress it, and digging into the ASan implementation is likely a misguided rabbit hole.
> 
> I'd recommend focusing on trying to find out the name of the BMI variable being queried for purposes of setting, such as by printing it to standard error like ASan writes its notification. Alternately, just audit all of the variables that you expect to go through that code path, and make sure the BMI metadata functions and the tables they're relying on are correct for them.

I've been going through these steps as time permits and trying to understand what could be happening here.  It is somewhat odd that the gcc sanitizer doesn't see this issue, and the ASan info produced in this given error is a bit puzzling as I trace through the test code and the code paths to the offending line.

I'll try to write up my thoughts along the way here.

PhilMiller:
OK, looking at this more closely, it starts to look like some sort of defect in ASan, or something managing to corrupt its shadow memory, which would involve truly impressive stack corruption.

As it tells you, the offending memory access on address `0x000101300160` is exactly the first entry of two in `output_var_names`
```
0x000101300160 is located 0 bytes inside of global variable 'output_var_names' defined in '/Users/runner/work/ngen/ngen/extern/test_bmi_c/src/bmi_test_bmi_c.c' (0x000101300160) of size 16
```
The shadow memory values covering `output_var_names` and `output_var_types` and a lot of space around them are all `f9`, "Global redzone".

My intuition here is that something is going wrong with either a mis-matched build between the `ngen` binary and the `test_bmi_c` dynamic library, or with how we're using `dlopen` and its interaction with ASan.

I'm going to take a look at the build logs to see if anything jumps out at me.

PhilMiller:
With what I've seen so far, I'm willing to entertain the notion that ASan is actually broken here. I've seen it simply fail to load a program before in weird enough circumstances, but this is the first time I've seen an apparent false positive error report.

In grad school and previous jobs, I would have been able to relish the opportunity to dig in to the root cause, report it, and maybe actually fix this. Alas, not what we're paid for here.

robertbartel:
> https://groups.google.com/g/address-sanitizer/c/XJXOrSvN8vg?pli=1 makes me mildly concerned that maybe the test binary has not been compiled and/or linked with the sanitizer flags

While not definitive analysis, I did confirm in my own development environment that if the CFLAGS and CXXFLAGS are set for using ASan, both the test executable (`test_bmi_c`) and the test BMI module shared library use `@rpath/libclang_rt.asan_osx_dynamic.dylib` and have plenty of `__asan_` stuff in their symbol tables.  I suppose it's still possible something subtle in the runner environment is causing this not to work exactly the same, though I would have expected the test binary to be fine and the test BMI module to not be compiled/linked properly.

> I'll note that this error seems to be occurring on the very first library-global-variable memory access in the very first code called in the dynamically-loaded BMI module. That would also potentially point to a mis-matched compilation/linkage concern.

It is ... and it isn't.

The ASan error happens in first test of the _Bmi_C_Formulation_Test_ class.  I didn't notice or think about this earlier, but prior to those BMI C formulation tests, _all_ the _Bmi_C_Adapter_Test_ tests are run and pass.   And _Bmi_C_Adapter_Test_ does exercise the module's `Get_var_type` function (at minimum, in `Bmi_C_Adapter_Test.GetVarType_0_a`).  

Curiously, the problem test actually has the ngen adapter class calling the test BMI module's `Get_var_itemsize` function, which in turn calls `Get_var_type`, where the potential buffer overflow code happens.   _Bmi_C_Adapter_Test_ does not appear to exercise `Get_var_itemsize`.  On the surface, I wouldn't expect that to matter, but we're getting into some gnarly territory here.

robertbartel:
> Ok, I've managed to find some changes that seem to address at least some of issues.

I take it back ... whether the error occurs does not appear to be directly related to the changes.

Consider these Action runs:
1. [Run 276](https://github.com/robertbartel/ngen/actions/runs/19476122915/job/55736233127) (commit [9bbb9a9](https://github.com/robertbartel/ngen/commit/9bbb9a9af23444120bba7ee626db88afe502417d))
  i. Passes everything
2. [Run 279](https://github.com/robertbartel/ngen/actions/runs/19477932017) (commit [24cbe26](https://github.com/robertbartel/ngen/commit/24cbe26a4d9895b22a0fa9287aeeea920a4e962d))
  i. Fails just _test_bmi_multi_
3. [Run 280](https://github.com/robertbartel/ngen/actions/runs/19479918104/job/55749191969) (commit [991d100](https://github.com/robertbartel/ngen/commit/991d1001f7e67624b73f4a4047ce0c64ffa96762)
  i. Fails _test_bmi_multi_, and now the earlier issue in _test_bmi_c_

Commits from 1 and 3 are no different, and changes between 1/3 and 2 don't seem to be meaningfully different in a way that could cause problems.

robertbartel:
I ran another similar set of experiments in my Mac dev environment, compiling with `-O0` and running `test_bmi_c` and `test_bmi_cpp` 1000 times each.  In both cases, there were no ASan errors.  

At this point, I suggest we adjust the Actions so that `-O0` is used for Mac runners.  We also open up a separate issue to investigate these ASan errors further, though not necessarily right now.  

As far as I can tell, `-O1` or higher is recommended for use with ASan because `-O0` is slow, but using the latter should be perfectly valid.  I've not seen anything suggesting `-O1` is better for error detection because, e.g., with `-O1` you'll catch errors that you wouldn't with `-O0` (if anything, it seems the reverse would be more likely).  Obviously something subtle and contrary to expectations is going on here, which eventually someone will need to figured out, but I think we should prioritize getting the CI checks working.   

@hellkite500, @aaraney, @PhilMiller:  thoughts?  We can assemble and discuss in more detail if needed.

hellkite500:
Closed in favor of #921

PR #915, opened 2025-10-29
Fix type handling of Fortran BMI adapter

robertbartel:
The existing BMI adapter for Fortran modules will not behave correctly in situations when the size of Fortran types is modified from the "standard" expected size.  This is because the adapter does not ask the module directly for variable item sizes when deciding what corresponding type to use for values inside the framework.  

For example, consider this method from _Bmi_Fortran_Adapter_:

``` C++
            inline void inner_get_value(const std::string& name, void *dest) {
                std::string varType = inner_get_var_type(name);
                //Can use the C type or the fortran type, e.g. int or integer
                if (varType == "int" || varType == "integer") {
                    inner_get_value_int(name, (int *)dest);
                }
                else if (varType == "float" || varType == "real") {
                    inner_get_value_float(name, (float *)dest);
                }
                else if (varType == "double" || varType == "double precision") {
                    inner_get_value_double(name, (double *)dest);
                }
                else {
                    throw ::external::ExternalIntegrationException(
                            "Can't get model " + model_name + " variable " + name + " of type '" + varType + ".");
                }
            }
```

The adapter only asks for a variable's type and _assumes_ the appropriate item size strictly based on the type name string:  e.g., that an advertised `integer` will be the same size as a C++ `int`.  This is not guaranteed to be correct, as type resizing can be done at compile time, using flags like _gfortran_'s `-fdefault-real-8`.

## Changes

- Correct `get_analogous_cxx_type` to get the C++ type corresponding to not just the category (i.e., _integer_, _float_) but also the memory size of a variable
- Correct adapter BMI getter/setter functions similarly, having them now defer to `get_analogous_cxx_type` for determining how types should be handled 
- Moving the first level of getter/setter logic from "inner" methods to directly within the main BMI `GetValue`/`SetValue` implementations, since updated implementations now violated the intended premise of "inner" functions for the class.

## Testing

1.  ~Existing unit tests pass as they did previously~ (actually, they are ... there are a few failing (though seemingly unrelated) tests ... I'm investigating)

## Notes

- I removed some "inner" functions (see the header file for details), because doing so fell within the scope.  The necessary updates violated the stated premise for "inner" functions, thus imposing a need to move where the logic was placed (and thereby making redundant the functions that were removed).  The class would likely benefit from refactoring to remove other "inner" functions also, and just have logic be directly within the implemented BMI functions.  However, that would be beyond the scope for this.

## Todos

- Before taking out of draft status, I need to confirm the functionality by re-compiling the Fortran components locally with the adjusted precision, and then running a few tests.

## Checklist

1.

- [ ] Linux

### DIFF
--- a/include/bmi/Bmi_Fortran_Adapter.hpp
+++ b/include/bmi/Bmi_Fortran_Adapter.hpp
@@ -535,8 +535,8 @@ namespace models {
             /**
              * Internal implementation of logic used for @see GetInputItemCount.
              *
-             * "Inner" functions such as this should not contain nested function calls to any other member functions for
-             * the type.
+             * "Inner" functions such as this should not contain nested function calls to any non-inner member functions
+             * for the type.
              *
              * @return The count of input BMI variables.
              * @see GetInputItemCount
@@ -552,8 +552,8 @@ namespace models {
             /**
              * Internal implementation of logic used for @see GetOutputItemCount.
              *
-             * "Inner" functions such as this should not contain nested function calls to any other member functions for
-             * the type.
+             * "Inner" functions such as this should not contain nested function calls to any non-inner member functions
+             * for the type.
              *
              * @return The count of output BMI variables.
              * @see GetOutputItemCount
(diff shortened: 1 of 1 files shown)


PhilMiller:
This sounds like a nice change. Did the case(s) in question come up with one of the BMI models we have at hand?

Issue #920, opened 2025-12-03
Add optional alternative nexus output format


robertbartel:
Add an optional, alternative format for nexus outputs that writes nexus data to per-formulation files, instead of writing data to per-nexus files.  This will facilitate ingest of the nexus data by t-route.  It also will help to avoid or mitigate problems with large domains and too many open files, as some compute environments (especially shared research environments) may have strict limits on the number of available file descriptors.

### Requirements
- [ ] Configurable option (almost certainly via realization config)
- [ ] When active, alternate format nexus data file(s) are written, and default per-nexus data files are not opened
- [ ] Alternate format nexus data files will be NetCDF
- [ ] Alternate format nexus data files will be per-formulation, with time and nexus id components added to the data
- [ ] Implementation will need to support writing multiple per-formulation files when the `formulations` section in the realization config has multiple separate formulations configured
  - If ngen does not already fully support such a configuration (needs to be confirmed), then this will need to ensure it is compatible with such functionality if/when available, and that may be addressed at the same time to support ensembling
- [ ] When alternatve format is active, enforce and inform that only global formulation configs are supported
  - Without this, catchments (and nexuses) would have separate formulations, and we'd effectively be back at per-nexus files
  - If this starts to appear to be problematic or unsatisfactory, another issue can be opened to track improving in the future
- [ ] (Implied) During parallel execution, all ranks should write to the same file(s)
  - This might have ramifications that either need to be addressed immediately or noted for later, and in either case clearly documented

PR #922, opened 2025-12-10
Support optional per-formulation NetCDF nexus outputs

robertbartel:
Changes made to address #920.

## Additions

- New classes for managing and writing Nexus output files
  - Abstraction
  - Implementation for previous per-nexus CSV file setup
  - Implementation for new per-formulation NetCDF file setup

## Changes

- Have formulation manager read and expose new config value for setting whether to use new per-formulation NetCDF format for nexus outputs
  - And updating realization config doc to reflect this and note new capability
- Setup nexus output manager class in main function and pass to created _SurfaceLayer_ instead of map of CSV files
- Use nexus output manager class in _SurfaceLayer_ to write outputs

## Testing

1.  New automated unit tests for class handling the per-formulation NetCDF file writing and management
2. Manual testing of ngen executions exercising the new functionality, both serially and executing with `mpirun`

## Checklist

1.

- [ ] Linux

### DIFF
--- a/test/utils/PerFormulationNexusOutputMgr_Test.cpp
+++ b/test/utils/PerFormulationNexusOutputMgr_Test.cpp
@@ -462,6 +462,22 @@ TEST_F(PerFormulationNexusOutputMgr_Test, construct_0_c) {
     ASSERT_EQ(size_vector[1], 1);
 }
 
+/** Test that example 0 gets constructed and is immediately not marked as closed. */
+TEST_F(PerFormulationNexusOutputMgr_Test, construct_0_d)
+{
+    std::string form_name = ex_0_form_0_nexus_ids[0];
+
+    utils::PerFormulationNexusOutputMgr mgr(ex_0_form_0_nexus_ids, ex_0_form_names, output_root, 2);
+
+    // Make sure we know what files to clean up
+    std::shared_ptr<std::vector<std::string>> filenames = mgr.get_filenames();
+    for (const std::string& f : *filenames) {
+        files_to_cleanup.push_back(f);
+    }
+
+    ASSERT_FALSE(mgr.is_closed());
+}
+
 /** Test correct chunking setup for example 3 (single instance). */
 TEST_F(PerFormulationNexusOutputMgr_Test, construct_3_c) {
 
@@ -1022,6 +1038,80 @@ TEST_F(PerFormulationNexusOutputMgr_Test, commit_writes_3_b) {
         }
     }
 
+}
+
+/** Test that close works automatically after all time steps for example 0. */
+TEST_F(PerFormulationNexusOutputMgr_Test, is_closed_0_a) {
+
+    std::string form_name = ex_0_form_names->at(0);
+
+    utils::PerFormulationNexusOutputMgr mgr(ex_0_form_0_nexus_ids, ex_0_form_names, output_root, 2);
+
+    // Make sure we know what files to clean up
+    std::shared_ptr<std::vector<std::string>> filenames = mgr.get_filenames();
+    for (const std::string& f : *filenames) {
+        files_to_cleanup.push_back(f);
+    }
+
+    for (size_t t = 0; t < ex_0_timestamps.size(); ++t) {
+        for (size_t n = 0; n < ex_0_form_0_nexus_ids.size(); ++n) {
+            mgr.receive_data_entry(form_name,
+                                   ex_0_form_0_nexus_ids[n],
+                                   utils::time_marker(t, ex_0_timestamps_seconds[t], ex_0_timestamps[t]),
+                                   ex_0_data[t][n]);
+        }
+        ASSERT_FALSE(mgr.is_closed());
+        mgr.commit_writes();
+    }
+    ASSERT_TRUE(mgr.is_closed());
+}
+
+/** Test that close does not work automatically if all time steps not gone through for example 0. */
+TEST_F(PerFormulationNexusOutputMgr_Test, is_closed_0_b) {
+
+    std::string form_name = ex_0_form_names->at(0);
+
+    utils::PerFormulationNexusOutputMgr mgr(ex_0_form_0_nexus_ids, ex_0_form_names, output_root, 2);
+
+    // Make sure we know what files to clean up
+    std::shared_ptr<std::vector<std::string>> filenames = mgr.get_filenames();
+    for (const std::string& f : *filenames) {
+        files_to_cleanup.push_back(f);
+    }
+
+    for (size_t n = 0; n < ex_0_form_0_nexus_ids.size(); ++n) {
+        mgr.receive_data_entry(form_name,
+                               ex_0_form_0_nexus_ids[n],
+                               utils::time_marker(0, ex_0_timestamps_seconds[0], ex_0_timestamps[0]),
+                               ex_0_data[0][n]);
+    }
+    ASSERT_FALSE(mgr.is_closed());
+    mgr.commit_writes();
+    ASSERT_FALSE(mgr.is_closed());
+}
+
+/** Test that close works if called explicitly for example 0. */
+TEST_F(PerFormulationNexusOutputMgr_Test, is_closed_0_c) {
+
+    std::string form_name = ex_0_form_names->at(0);
+
+    utils::PerFormulationNexusOutputMgr mgr(ex_0_form_0_nexus_ids, ex_0_form_names, output_root, 2);
+
+    // Make sure we know what files to clean up
+    std::shared_ptr<std::vector<std::string>> filenames = mgr.get_filenames();
+    for (const std::string& f : *filenames) {
+        files_to_cleanup.push_back(f);
+    }
+
+    for (size_t n = 0; n < ex_0_form_0_nexus_ids.size(); ++n) {
+        mgr.receive_data_entry(form_name,
+                               ex_0_form_0_nexus_ids[n],
+                               utils::time_marker(0, ex_0_timestamps_seconds[0], ex_0_timestamps[0]),
+                               ex_0_data[0][n]);
+    }
+    ASSERT_FALSE(mgr.is_closed());
+    mgr.close();
+    ASSERT_TRUE(mgr.is_closed());
 }
 
 /** Test that example 0 works with write_nexus_ids_once and has nexus id var values written to NetCDF file. */


include/utilities/NexusOutputsMgr.hpp
```diff
@@ -0,0 +1,93 @@
...
+         * @return The value for the default formulation id.
+         */
+        const std::string get_default_formulation_id() {
+            return "default";
+        }
+
+    };
+} // utils
+
+#endif //NGEN_NEXUSOUTPUTSMGR_HPP
```
hellkite500: posix newline, henceforth referred to as cowbell

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,209 @@
...
+         * ``local_nexus_ids[0]``, etc.
+         */
+        size_t local_offset;
+        /** Map of formulation ids to nexus data file paths (as string) */
+        std::unordered_map<std::string, std::string> nexus_outfiles;
+
+    };
+} // utils
+
+#endif //NGEN_PERFORMULATIONNEXUSOUTPUTMGR_HPP
```
hellkite500: more cowbell!

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -55,6 +55,13 @@ namespace realization {
                 //TODO seperate the parsing of configuration options like time
                 //and routing and other non feature specific tasks from this main function
                 //which has to iterate the entire hydrofabric.
+
+                // TODO: (later) consider whether this really belongs inside the global formulation
```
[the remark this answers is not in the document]
robertbartel: At the moment I have.  The `per_formulation_nexus_files` element is an optional sibling of, e.g., `time` and `global` in the realization config.  It is not a child of `global`.  This TODO was a reminder to reexamining in the future whether it _should_ be a child of the formulation spec, in case I failed to consider something while trying to implement this quickly.  

If you are confident this is the way to go, I can remove the TODO comment?

robertbartel:
FYI, this is ready for a full review.  At least one of those check failures I believe is a false positive, which I'd hoped would be fixed by #921.  The other I'm not sure about; it may be an issue with the check being incompletely set (and thus not having netcdf available), or it may be an issue with this code not sufficiently handling situations when the option to use netcdf is turned off.

include/utilities/NexusOutputsMgr.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+    protected:
+        ~NexusOutputsMgr() = default;
+
+        /**
+         * Get some a formulation id value.
+         *
+         * @return The value for the default formulation id.
+         */
+        const std::string get_default_formulation_id() {
```
hellkite500: Is this intended to be overridden by derived managers?  If so this should also be virtual

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,308 @@
...
+
+    /**
+     * Subtype that write/manages per-formulation nexus data files stored in NetCDF format.
+     *
+     * Because it works with NetCDF files, this file does not write received data entries until ``commit_writes`` is
+     * called.
+     */
+    class PerFormulationNexusOutputMgr : public NexusOutputsMgr
+    {
+    private:
```
hellkite500: Putting the friend class declaration under the private specifier doesn't actually do anything.  And private is the default specifier anyways, so not sure this is really needed nor helpful since it only contains the friend declaration.

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,308 @@
...
+                }
+                data[i] = data_cache[nexus_ids[i]];
+            }
+
+            const std::string filename = nexus_outfiles[current_formulation_id];
+
+            const netCDF::NcFile ncf(filename, netCDF::NcFile::write, netCDF::NcFile::nc4);
+
+            // On the first write, also write the nexus id variable values
+            if (current_time_index == 0) {
```
hellkite500: This seems worth putting into a private function to test by itself.  Also makes changing when in the lifecycle this happens if we want to in the future.

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,308 @@
...
+                std::vector<unsigned int> numeric_nex_ids(this->nexus_ids.size());
+                char delimiter = '-';
+
+                for (size_t i = 0; i < this->nexus_ids.size(); ++i) {
+                    std::string::size_type pos = this->nexus_ids[i].find(delimiter);
+                    if (pos == std::string::npos) {
+                        throw std::runtime_error("Invalid nexus id '" + this->nexus_ids[i] + "' (no delimiter '"
+                                                 + std::string(1, delimiter) + "').");
+                    }
+                    numeric_nex_ids[i] = std::stoi(this->nexus_ids[i].substr(pos + 1));
```
hellkite500: I'm mildly curious about handling conflicting numeric ids, since we loose context about nexus types which may have overlapping numeric ids depending on the source/hydrofabric/ect

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,308 @@
...
+                }
+                const netCDF::NcVar var_nexus = ncf.getVar(nc_nex_id_dim_name);
+                var_nexus.putVar({this->local_offset}, {numeric_nex_ids.size()}, numeric_nex_ids.data());
+            }
+
+            const netCDF::NcVar flow = ncf.getVar(nc_flow_var_name);
+
+            // Assume base on how constructor was set up (imply for conciseness)
```
hellkite500: This comment is somewhat confusing, not sure what its trying convey

robertbartel:
I'll also note that the aforementioned errors seems to be stemming from a call to `HY_PointHydroNexus.get_downstream_flow`, made within `SurfaceLayer.update_models`.  I did significantly update `SurfaceLayer.update_models` in these changes, but not really how the call to `HY_PointHydroNexus.get_downstream_flow` (or the code to determine its args) works.  

One thing I did do was alter `SurfaceLayer.update_models` to only get the current timestamp once, outside the loop for handling output for the nexuses instead of inside that loop.  That calls `Simulation_Time.get_timestamp`, which _can_ have side effects in certain circumstances, but I am pretty sure it effectively shouldn't here.  And even if it did, I don't see why that would cause this type of error.  Still, I could use a sanity check on that @hellkite500.  See the [diff for that file](https://github.com/NOAA-OWP/ngen/pull/922/files#diff-dc88df1d0165ca575757a1864dc91829bf00c3c2880bd889e370d94bdd323f2c).

include/utilities/PerFormulationNexusOutputMgr.hpp
robertbartel: I was initially writing these as strings but ran into a snag, and before I could get it resolved noticed that the example file @shorvath-noaa had sent me with dummy data used an integer variable for the feature ids and only used the numeric component.  So I switched this to only use the numeric ids also.  If we need to revisit this, I don't object, but it likely will need to be holistically.

include/utilities/NexusOutputsMgr.hpp
hellkite500: Then we should probably pick where that default is implemented and mark that as `final`

include/utilities/FileChecker.h
```diff
@@ -15,6 +19,21 @@ namespace utils {
...
     public:
 
+        /**
+         * Check whether a directory already exists.
+         *
+         * @param path The path of interest.
+         * @return Whether there is a directory existing at the given path.
+         */
+        static bool directory_exists(const std::string& path) {
+            #ifdef _WIN32 // This should catch both 32 and 64 bit Windows
```
aaraney: ditto

include/utilities/NexusOutputsMgr.hpp
```diff
@@ -0,0 +1,84 @@
...
+#define NGEN_NEXUSOUTPUTSMGR_HPP
+
+#include <string>
+
+namespace utils
+{
+    /** Simple helper struct for tracking meaningful points in time for output writing. */
+    struct time_marker
+    {
+        const long sim_time_index;
```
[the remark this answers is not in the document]
hellkite500: I have thought often about type cleanup across the board.  Not opposed to starting here =).  We use long for time steps/indicies in many places, we may want to scope that change seperately and make a complete update to time types across ngen wholistically.

include/utilities/NexusOutputsMgr.hpp
```diff
@@ -0,0 +1,84 @@
...
+         */
+        virtual void commit_writes() = 0;
+
+        /**
+         * Get whether this instance manages writing data for the given nexus to a managed data file.
+         *
+         * @param nexus_id The id for the Nexus of interest.
+         * @return Whether this instance manage a file storing data for the given nexus.
+         */
+        virtual bool is_nexus_managed(const std::string &nexus_id) = 0;
```
aaraney: I don't see where this is used. Can we drop this?

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,316 @@
...
+namespace utils
+{
+
+    /**
+     * Subtype that write/manages per-formulation nexus data files stored in NetCDF format.
+     *
+     * Because it works with NetCDF files, this file does not write received data entries until ``commit_writes`` is
+     * called.
+     */
+    class PerFormulationNexusOutputMgr : public NexusOutputsMgr
```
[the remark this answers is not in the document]
robertbartel: I'm on the fence about this, but leaning toward leaving it alone for now.  Doing it properly would be much more complicated than just this suggestion.  It's certainly a more precise name, but also longer (and was already on the long side).  Given that and the complexity, I'd rather leave it be at the moment.  Let me know if there's a strong rationale I'm not considering for why we should make this change up front (we can always refactor it later if we _need_ to).

src/NGen.cpp
```diff
@@ -423,30 +426,59 @@ int main(int argc, char *argv[]) {
         partition_one.generate_partition(catchment_collection);
         local_data = std::move(partition_one.partition_data);
     }
-    hy_features::HY_Features_MPI features = hy_features::HY_Features_MPI(local_data, nexus_collection, manager, mpi_rank, mpi_num_procs);
+    std::shared_ptr<hy_features::HY_Features_MPI> features = std::make_shared<hy_features::HY_Features_MPI>(local_data, nexus_collection, manager, mpi_rank, mpi_num_procs);
```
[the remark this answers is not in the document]
robertbartel: I will undo the changes to make the `features` variable a pointer.  That was originally to support another design decision which made use of the possibility of `null` (I forget the exact details at this point), but has since been moved away from.

The other shared pointer I added is for the `NexusOutputsMgr`.  I suppose that can be a unique pointer instead, and then have the layer class accept a reference as you suggested in a separate comment.

robertbartel:
Hmm, after some testing by Jason, I realize that I haven't fully accounted for supporting an environment that uses MPI but does not have parallel NetCDF installed (or even documented the parallel NetCDF dependency).  I'll need to address those things.

robertbartel:
Things should be updated now.  I've added to the documentation some more, including details on the optional parallel NetCDF dependency.  I've also configured builds so that this new feature will require parallel NetCDF if you try to use it with a parallel ngen build, though it will only require NetCDF if used with just a serial build.

include/utilities/NexusOutputsMgr.hpp
hellkite500: Sorry, too many threads of thought in the previous comments haha.

If there really should be a single default string across all managers, then maybe we should consider that as a namespace constexpr instead of a function?.  It wasn't clear initially if this was intended to be a virtual function where sublcasses could override the default, which is where my original comment came from.  I was just thinking through the needs of a "default" and whether it needed strict consistency or not.  I'm not sure it really matters if the goal is to have "some default".

include/core/SurfaceLayer.hpp
```diff
@@ -33,7 +34,7 @@ namespace ngen
         private:
 
         std::vector<std::string> nexus_ids;
-        std::unordered_map<std::string, std::ofstream>& nexus_outfiles;
+        std::shared_ptr<utils::NexusOutputsMgr> nexus_outputs_mgr;
```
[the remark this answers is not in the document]
PhilMiller: If I were coming at this earlier, I would voice a preference for simulation classes like `SurfaceLayer` not having any reference to the data outputs. Rather, they should just present interfaces for the output logic to grab the relevant data.

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,778 @@
...
+                                     const size_t total_timesteps,
+                                     const int rank,
+                                     const std::vector<int>& nexuses_per_rank)
+            :   nexus_ids(nexus_ids),
+                rank(rank),
+                nexuses_per_rank(nexuses_per_rank),
+                total_timesteps(total_timesteps)
+        {
+            // TODO: look at potentially making this configurable rather than static
+            data_flush_interval = 100;
```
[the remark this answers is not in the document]
aaraney: Just back of the envelope, hdf5's cache is usually 16MiB (2^24 bytes). Im not sure if there is a separate cache per open file or if its shared? Just wondering b.c. we'd also be reading forcing data... Anyways, if we have lets say 1024 (2^10) nexuses in an mpi rank and we are writing out doubles 8 bytes (2^3) that's 2^13 bytes for a single timestep. That leaves us (24-13=11) 2024 (2^11) timesteps that we could in theory put in the cache that i'm treating like a buffer (need to confirm this).

I guess more what i'm getting at, is if we can determine the size of the buffer netcdf is using to buffer writes we can flush to disk at a rate that matches when we would fill up that buffer. So, parametrize this using the size of `nexus_ids`  and the netcdf (hdf5) cache.

include/realizations/catchment/Formulation_Manager.hpp
```diff
@@ -137,9 +153,17 @@ namespace realization {
 
                 /**
                  * Read catchment configurations from configuration file
-                 */      
+                 */
                 auto possible_catchment_configs = tree.get_child_optional("catchments");
 
+                // For now at least, this isn't allowed
```
hellkite500, line 159: @robertbartel is there a reason you had in mind for why this is the case?  Nexus features are reasonably independent of the catchment/formulation itself, and I'm not seeing off the top of my head why this feature wouldn't "just work" regardless of the form of formulation.

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,786 @@
...
+                if (this->rank != 0)
+                    throw std::runtime_error("Must supply nexuses_per_rank values when using multiple MPI processes.");
+                this->nexuses_per_rank.push_back(this->nexus_ids.size());
+            }
+
+            if (this->nexuses_per_rank.size() <= this->rank) {
+                throw std::runtime_error("To few values in nexuses_per_rank value ("
+                    + std::to_string(this->nexuses_per_rank.size()) + ") for rank "
+                    + std::to_string(rank) + ".");
+            }
```
hellkite500: I'm not understanding this condition here, and its not clear what its evaluating based on the previous lines building the `nexuses_per_rank`

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,786 @@
...
+        /**
+         * Receive a data entry for this nexus, specifying details including the formulation id.
+         *
+         * @param formulation_id The id of the formulation involved in producing this data.
+         * @param nexus_id The id for the nexus to which this data applies.
+         * @param data_time_marker A marker for the current simulation time for the data.
+         * @param flow_data_at_t The nexus flow contribution at this time index (the main data to write).
+         */
+        void receive_data_entry(const std::string &formulation_id, const std::string &nexus_id,
+                                const time_marker &data_time_marker, const double flow_data_at_t) override
```
hellkite500: `flow_data_at_t` *could* be a const reference as well -- the assignment into the map container will make a copy at that point.

test/utils/PerFormulationNexusOutputMgr_Test.cpp
```diff
@@ -0,0 +1,1096 @@
...
+    }
+
+    ASSERT_THROW(mgr.receive_data_entry(form_name,
+                                ex_0_form_0_nexus_ids[0],
+                                        utils::time_marker(1, ex_0_timestamps_seconds[1], ex_0_timestamps[1]),
+                                        ex_0_data[1][0]),
+                 std::runtime_error);
+}
+
+// TODO: figure out what is going on with this one
```
hellkite500: Did you ever figure this out?

test/utils/PerFormulationNexusOutputMgr_Test.cpp
```diff
@@ -0,0 +1,1096 @@
...
+
+    std::vector<std::string> nex_id_strs(nex_id_numeric.size());
+    for (size_t i = 0; i < nex_id_strs.size(); ++i) {
+        nex_id_strs[i] = "nex-" + std::to_string(nex_id_numeric[i]);
+    }
+
+    ASSERT_EQ(nex_id_strs, ex_1_form_0_all_nexus_id);
+}
+
+#endif // #if NGEN_WITH_MPI
```
hellkite500: Can none of these tests apply in the serial use case?

test/gtest_mpi_main.cpp
```diff
@@ -0,0 +1,34 @@
...
+    MPI_Finalize();
+    return result;
+
+    #else
+
+    std::cerr << "MPI not available, skipping MPI tests." << std::endl;
+    return -1;
+
+    #endif
+}
```
hellkite500: newline -- cause `posix`?

include/realizations/catchment/Formulation_Manager.hpp
robertbartel: We're going to start getting into some semantic issues distinguishing exactly what a "formulation" is, both by itself across different contexts (e.g., configuration versus running simulation) and compared to, say, a "catchment realization" or "ensemble member."

If I'm remembering correctly, this largely came out of the details of the ensemble work.  We wanted to produce a single nexus output file for everything for an ensemble member.  It made sense that an ensemble member must be configured globally for a simulation.   So adopting this restriction helped to simplify the scope of the problem.

I see your point, but that really takes things beyond "per-formulation". We probably could do "per-simulation," but we'd anticipated in the future that there would actually be multiple formulations configured at the global level and run for all catchments, which we'd want to treat as individual ensemble members that had separated output.  So we really do want the "per-formulation" option.  We could support mutually exclusive options for both, but all this is widening the scope.

include/utilities/PerFormulationNexusOutputMgr.hpp
robertbartel: The value at each index `r` within `nexuses_per_rank` corresponds to the number of nexuses partitioned to rank `r`, and thus to the output manager object with rank `r`.  So, by line 96, the size of `nexuses_per_rank` _must_  be large than the rank value, for any possible rank value in the simulation, or there'll be a problem with the indices and correspondence.

Regarding the previous lines, by line 96 `nexuses_per_rank` must not be empty.  It will often have values set from the external source that provided it to the constructor.  In cases when there aren't multiple ranks (e.g., serial runs),  it's also possible to infer things, so it supports that.

include/utilities/PerFormulationNexusOutputMgr.hpp
robertbartel: Well, for one, `flow_data_at_t` actually gets assigned as a vector element.  The map is used to look up the vector index.  But pretty sure there's still a copy.

My impression is that the complier can optimize this kind of trivial copy sufficiently, so using a reference isn't really needed.  But if that's wrong I'm interested in hearing more.

test/utils/PerFormulationNexusOutputMgr_Test.cpp
robertbartel: Yeah, array dimensions issue.  Will push a fix in a moment.

test/utils/PerFormulationNexusOutputMgr_Test.cpp
robertbartel: Yes, those all run in the serial case.  Technically that's the `#endif` for the `#else` portion of `#if NGEN_MPI_UNIT_TESTS` now.  I'll change the comment a bit to hopefully help to clarify.

test/gtest_mpi_main.cpp
robertbartel: Oh fine ...

include/utilities/PerFormulationNexusOutputMgr.hpp
hellkite500: I interpret this as the size of `nexuses_per_rank` to be the number of ranks, with the value being, as you said, the number of nexus features managed by the rank at a given index, but that clearly isn't correct, and it's just challenging my brain to reason about this section haha.

doc/DEPENDENCIES.md
```diff
@@ -227,3 +228,28 @@ source venv/bin/activate
...
 
+### Parallel NetCDF
+
+In order to make use of the per-formulation NetCDF nexus output files capabilities [noted here](REALIZATION_CONFIGURATION.md#per_formulation_nexus_files) within installations of ngen that support parallel MPI execution, NetCDF-4 and its dependencies must have been built/installed with parallel I/O support.  [See here](https://docs.unidata.ucar.edu/netcdf-c/4.9.3/parallel_io.html) for further details in the NetCDF documentation.  
+
+Without NetCDF Parallel I/O support, parallel builds of ngen will be limited to individual, per-nexus output files for nexus data in CSV format, and an error will occur if a realization config attempting to use per-formulation outputs is supplied.
+
+You can see if the NetCDF dependency found by CMake for ngen builds includes parallel support by looking for this part of the CMake output when you generate/regenerate the build system directory.  Note specifically the `Parallel` sub-item at the end of the NetCDF details.
+
+```aiignore
```
PhilMiller: Out of curiosity, what does this annotation do here? I can only find sources mentioning files `.aiignore` that are like `.gitignore` for AI agents

include/core/SurfaceLayer.hpp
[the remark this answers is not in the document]
PhilMiller: On the reference vs `shared_ptr` front, the shared pointer does jumble ownership and lifetime concerns a little bit, but it guarantees that we're not dealing with a dangling reference sometime later. On this point, I'm in favor of the code as it appears right now.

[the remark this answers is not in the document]
PhilMiller: Regarding the necessity of distinguishing `SurfaceLayer` as opposed to `Layer`, it's almost unnecessary, but not entirely. The calls to `get_downstream_flow` have side effects inside the `HY_PointHydroNexus` instances

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,786 @@
...
+            if (rank == 0) {
+                create_netcdf_file();
+                setup_netcdf_metadata();
+            }
+            // If rank > 0, reg open
+            else {
+                open_netcdf_file();
+                lookup_netcdf_metadata();
+            }
+            #endif
```
PhilMiller: In the 2+ ranks case, there are some missing `MPI_Barrier()` calls in here or these functions' implementation

PhilMiller: To elaborate, the process creating and setting up the file needs to be done doing so, with an appropriate `flush`-like call returned, before any other process looks at it.

The `MPI_Barrier` call would then be after all that in the process doing the setup, and before anything is done in all the other processes

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,786 @@
...
+                throw std::runtime_error("PerFormulationNexusOutputMgr currently cannot have more than one formulation id.");
+            }
+            else {
+                this->formulation_id = (*formulation_ids)[0];
+            }
+
+            this->local_offset = 0;
+
+            for (size_t r = 0; r < rank; ++r) {
+                this->local_offset += nexuses_per_rank[r];
```
PhilMiller: There's a few small scalability bugs here, in that every process will have an `O(P)`-size array, and this loop will take `O(P)` time for the last rank. 

Looking at how the values in `nexuses_per_rank` are actually used, I think the `MPI_Bcast` in `main()` could actually be an ~`MPI_Scan`~ `MPI_Exscan` to calculate this `local_offset` value, and maybe an `MPI_Reduce` for rank 0 to have the total as calculated up at line 85, `flow_var_chunk_size_per_dim[0] += nexuses_per_rank[r];`. Both of those happen in `O(log P)` time, and don't require any process to carry more than its own single value.

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,786 @@
...
+         * Function will create and store the variable ids for the nexus_id, time, and flow/runoff variables.  To
+         * support those variables, it will also create the nexus_id and time dimensions, though not store those ids.
+         */
+        void setup_netcdf_metadata() {
+            int nc_nex_id_dim_id, nc_time_dim_id;
+
+            size_t total_nexuses = 0;
+            for (const int n : nexuses_per_rank) {
+                total_nexuses += n;
+            }
```
PhilMiller: This is actually redundant with a calculation performed in the constructor. It's also the only use of the member variable outside the constructor. Turn it into a function parameter, skip the recalculation, and drop the member variable entirely?

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,786 @@
...
+                {"calendar", "gregorian"},
+                {"long_name", "Time"}
+            };
+            add_variable(nc_time_dim_name, NC_UINT, {nc_time_dim_id}, time_var_attrs, &time_nc_var_id);
+
+            std::map<std::string, std::string> flow_var_attrs = {
+                {"units", "m3 s-1"},
+                {"long_name", "Simulated Surface Runoff"}
+            };
+            add_variable(nc_flow_var_name, NC_DOUBLE, {nc_nex_id_dim_id, nc_time_dim_id}, flow_var_attrs,
```
PhilMiller: I just want to make sure that the dimension order here is such that all of the nexus values for a time step are going to end up contiguous, rather than all of the time steps for each nexus. The NetCDF docs for `nc_def_var` don't spell that out.

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,786 @@
...
+        size_t data_flush_interval;
+
+        /** The total number of timesteps that will be written to the managed file. */
+        size_t total_timesteps;
+
+        /** File id member variable for opening and writing to NetCDF file via the C API. */
+        int netcdf_file_id = -1;
+
+        /** The rank (MPI rank if using MPI) for this instance, which will be 0 if parallel execution is not in use. */
+        int rank;
```
PhilMiller: Is this being used for isolated testing, or is it duplicative of the global `int mpi_rank;` in `NGen.cpp`?

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,786 @@
...
+
+        /** Variable id for the ``nexus_id`` NetCDF variable. */
+        int nexus_nc_var_id;
+
+        /** Variable id for the ``time`` NetCDF variable. */
+        int time_nc_var_id;
+
+        /** Variable id for the ``flow`` NetCDF variable. */
+        int flow_nc_var_id;
```
PhilMiller: I'd suggest renaming these to be `nc_var_id_foo` to make their commonality more apparent.

include/utilities/PerFormulationNexusOutputMgr.hpp
PhilMiller: Yeah, don't make references of primitive types. If it's a primitive type, it can pass in a register. A reference can stick the underlying pointer in a register, but the actual value still has to be stored on the stack and dereferenced. That can't be optimized away, because it's part of the ABI, unless the function call gets inlined.

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,786 @@
...
+            default:
+                return "unrecognized error code '" + std::to_string(nc_status) + "'";
+            }
+        }
+
+        const std::string nc_nex_id_dim_name = "feature_id";
+        const std::string nc_time_dim_name = "time";
+        const std::string nc_flow_var_name = "runoff_rate";
+
+        const double flow_var_fill_value = -9999.0;
```
PhilMiller: I know it's not the common practice, but could this please be made `NaN` to make any contaminated calculations blow up, rather than silently calculating wrong values?

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,786 @@
...
+            current_formulation_id.clear();
+
+            if (current_time_index == total_timesteps) {
+                nc_status = nc_close(netcdf_file_id);
+                if (nc_status != NC_NOERR) {
+                    throw std::runtime_error("Error closing nexus file '" + nexus_outfile + "' after write for "
+                        + "FINAL time step " + std::to_string(current_time_index) + ": "
+                        + parse_netcdf_return_code(nc_status));
+                }
+            }
```
PhilMiller: I think this functionality should be extracted to a separate function, rather than solely appearing here. In a run that doesn't go all the way to `total_timesteps` for whatever reason (e.g. state save), we still need the `nc_close()` call to happen to a final flush and clean shutdown.

include/utilities/PerNexusCsvOutputMgr.hpp
```diff
@@ -0,0 +1,60 @@
...
+         * @param nexus_id The id for the nexus to which this data applies.
+         * @param data_time_marker A marker for the current simulation time for the data.
+         * @param flow_data_at_t The nexus flow contribution at this time index (the main data to write).
+         */
+        void receive_data_entry(const std::string &formulation_id, const std::string &nexus_id,
+                                const time_marker &data_time_marker, const double flow_data_at_t) override {
+            if (formulation_id != get_default_formulation_id()) {
+                throw std::runtime_error("Cannot write data entry for non-default formulation " + formulation_id + " for nexus " + nexus_id + " when per-nexus CSV output is enabled.");
+            }
+            nexus_outfiles[nexus_id] << data_time_marker.sim_time_index << ", " << data_time_marker.time_stamp << ", " << flow_data_at_t << std::endl;
```
PhilMiller: Per the recently merged #945, 
```suggestion
            nexus_outfiles[nexus_id] << data_time_marker.sim_time_index << ", " << data_time_marker.time_stamp << ", " << flow_data_at_t << '\n';
```

include/utilities/PerNexusCsvOutputMgr.hpp
```diff
@@ -0,0 +1,60 @@
...
+        PerNexusCsvOutputMgr(const std::vector<std::string>& nexus_ids, const std::string &output_root) {
+            for(const auto& id : nexus_ids) {
+                nexus_outfiles[id].open(output_root + id + "_output.csv", std::ios::trunc);
+            }
+        }
+
+        /**
+         * No-op, since this type writes entries as they are received.
+         */
+        void commit_writes() override { }
```
PhilMiller: Given the below suggestion, this could call `for (auto &f : nexus_outfiles) { f.flush(); }` to have everything in the filesystem's hands, and not a process-internal buffer.

src/NGen.cpp
```diff
@@ -442,18 +445,54 @@ int main(int argc, char *argv[]) {
...
-                nexus_outfiles[id].open(manager->get_output_root() + id + "_output.csv", std::ios::trunc);
-            }
-        } else {
-          nexus_outfiles[id].open(manager->get_output_root() + id + "_output.csv", std::ios::trunc);
+        std::vector<int> nexuses_per_rank(mpi_num_procs, 0);
+        nexuses_per_rank[mpi_rank] += nexus_ids.size();
+        for (int i = 0; i < mpi_num_procs; i++) {
+            MPI_Bcast(nexuses_per_rank.data() + i, 1, MPI_INT, i, MPI_COMM_WORLD);
+            MPI_Barrier(MPI_COMM_WORLD);
         }
```
PhilMiller:
```suggestion
        MPI_Allgather(MPI_IN_PLACE, 1, MPI_INT, nexuses_per_rank.data(), 1, MPI_INT, MPI_COMM_WORLD);
}
```
Looped MPI collectives are a parallel code smell.

Rather than literally applying this change, though, please see my note about using ~~`MPI_Scan`~~ [edit:] `MPI_Exscan` and `MPI_Reduce` to avoid every process getting the whole array.

PhilMiller:
```suggestion
        int local_nexus_count = nexus_ids.size();
        int local_nexus_write_offset;
        int total_nexus_count;
        MPI_Exscan(&local_nexus_count, &local_nexus_write_offset, 1, MPI_INT, MPI_SUM, MPI_COMM_WORLD);
        if (rank = 0) MPI_Recv(&total_nexus_count, 1, MPI_INT, mpi_num_procs - 1, 0, MPI_COMM_WORLD, MPI_STATUS_NULL);
        if (rank = mpi_num_procs - 1) {
            total_nexus_count = local_nexus_write_offset + local_nexus_count;
            MPI_Send(&total_nexus_count, 1, MPI_INT, mpi_num_procs - 1, MPI_COMM_WORLD);
        }
```

PhilMiller: Or, rather, just see my suggested replacement code below

PhilMiller: Sorry, edited to call `MPI_Exscan` rather than `MPI_Scan`, since we want (e.g.) rank 0 to start at an offset of 0, not its count.

include/utilities/PerFormulationNexusOutputMgr.hpp
PhilMiller: See suggested implementation in comments on `NGen.cpp`

PhilMiller:
I've only really skimmed this as a whole. I've focused on the parallel issues and usage of the NetCDF library, since that's where I likely have the most to contribute. Everything else I've seen looks more or less fine.

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,786 @@
...
+                nc_status = nc_put_vara_long(netcdf_file_id, time_nc_var_id, &start_t, &count_t, &epoch_minutes);
+                if (nc_status != NC_NOERR) {
+                    throw std::runtime_error("Error writing time value to nexus file '" + nexus_outfile + "' ("
+                        + parse_netcdf_return_code(nc_status) + ") at time index " + std::to_string(current_time_index) + ".");
+                }
+            }
+
+            // Assume base on how constructor was set up (imply for conciseness)
+            //size_t nexus_dim_index = 0;
+            //size_t time_dim_index = 1;
```
PhilMiller: Actually, please make these named `constexpr` constants and use them in all the applicable call sites, rather than literal 0/1, so that those values are not confused with (e.g.) literal offsets or counts.

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,786 @@
...
+         *
+         * In any successful execution, this function will have the side effect of setting the @ref netcdf_file_id
+         * member variable.
+         *
+         * Note this function is expected to only be called at most once by an instance, during the constructor (though
+         * after the @ref nexus_outfile and @ref rank member variables are properly set).  An exception will be thrown
+         * if the @ref netcdf_file_id member variable is already set (to something other than ``-1``) when this function
+         * is called.
+         */
+        void create_netcdf_file_parallel() {
```
PhilMiller:
```suggestion
        void create_netcdf_file_parallel(MPI_Comm comm) {
```

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,786 @@
...
+            // Bail with error if this has already been run
+            if (netcdf_file_id != -1) {
+                throw std::runtime_error("Cannot (parallel) create netCDF file after already created and have nc_id set.");
+            }
+            std::cout << "Creating nexus NetCDF file '" << nexus_outfile << "' for parallel access." << std::endl;
+            int nc_status = nc_create_par(nexus_outfile.c_str(), NC_NETCDF4 | NC_NOCLOBBER, MPI_COMM_WORLD, MPI_INFO_NULL, &netcdf_file_id);
```
PhilMiller:
```suggestion
            int nc_status = nc_create_par(nexus_outfile.c_str(), NC_NETCDF4 | NC_NOCLOBBER, comm, MPI_INFO_NULL, &netcdf_file_id);
```

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,786 @@
...
+            }
+
+            nexus_outfile = output_root + "/formulation_" + this->formulation_id + "_nexuses.nc";
+
+            // To support unit testing when MPI and parallel netcdf compiled in (which won't always be running through
+            // MPI), we have to detect things.
+            #if NGEN_WITH_MPI && NGEN_WITH_PARALLEL_NETCDF
+            // If 2+ ranks yes MPI init, all ranks: para create + setup dims/vars
+            if (nexuses_per_rank.size() > 1 && isMpiInitialized()) {
+                create_netcdf_file_parallel();
```
PhilMiller:
```suggestion
                create_netcdf_file_parallel(MPI_COMM_WORLD);
```

doc/DEPENDENCIES.md
robertbartel: This seems to have been an auto-completed addition.  I don't actually remember, but I suspect I was interested in the fenced code block and just didn't notice the extra `aiiignore` that CLion (a JetBrains produce) will give you now after typing _```_ and then hitting _Return_.  

I've removed this, pending a push.

include/utilities/PerFormulationNexusOutputMgr.hpp
robertbartel: I think I assumed that because parallel NetCDF file metadata operations - like the `nc_create_par` call within `create_netcdf_file` - are collective, there was a barrier within their implementation.  But perhaps that's a mistake, and even if it was correct, a comment is in order to clarify.  I'll dig into this a bit.

robertbartel: Your second comment above didn't load for me initially for some reason @PhilMiller.  To expound on my above comment a bit ...

> To elaborate, the process creating and setting up the file needs to be done doing so, with an appropriate flush-like call returned, before any other process looks at it.

From places in the docs like [here](https://docs.unidata.ucar.edu/netcdf-c/4.9.3/parallel_io.html) and [here](https://docs.unidata.ucar.edu/netcdf-c/4.9.3/group__datasets.html#ga5e03371d14dfac165734b3d0d2c3adc6), it's made clear that functions for file creation - and as I read things, the rest of file metadata setup - need to be called by all the MPI processes.  So I think NetCDF is doing what you suggest behind the scenes.  

I thought for a moment after reading your first comment that we might also need a barrier after the metadata setup that I'd overlooked, but examining more closely now, I don't think we do.  I.e., I'm pretty sure the collective nature of those NetCDF parallel file functions handles everything properly for the parallel setup.  But if you still think I'm missing something, let me know.

PhilMiller: In the cases I'm concerned about here, `create_netcdf_file` and the contained call to `nc_create` (not `nc_create_par`, like in `create_netcdf_file_parallel`) are **not called collectively**. Rank 0 calls `create_netcdf_file`, and ranks not-0 call `open_netcdf_file`, which calls `nc_open`.

robertbartel: Ok, I've inadvertently tricked you here (or myself) ... there is no case where both `open_netcdf_file` is called and MPI is in use.   If MPI is in use, and there is more than one process, `create_netcdf_file_parallel` will be called in the constructor by all instances (not if there is _only_ one process, which is somewhat counterintuitive but not strictly prohibited AFAIK).

What is possible - though rare and not the typical intended use case - is for there to be multiple instances of this type of manager object in one process.  In that case, they still each get their subset of nexuses they handle, but they execute serially.   To distinguish themselves, they each still get a rank assignment.

That scenario probably will only ever apply in automated testing, where it is currently exercised.  That was to support other testing:  multiple instances in one process was easier to do than to try to get automated tests to execute through MPI.  I was working on the latter but didn't finish something that was ready to also be committed.

PhilMiller: Ok, those subtleties need to be way more explicit in either the code structure, or the adjacent comments. In particular, the thing about `rank` referring to distinct instances with serial execution in the testing setting.

include/utilities/PerFormulationNexusOutputMgr.hpp
robertbartel: I've cleaned this up a good bit based on the suggestion.

I will note that I'm a little unsatisfied with requiring `local_offset` to be determined outside the class.  Feels like there should be a better way to encapsulate that.  Maybe not without reducing performance, though.

src/NGen.cpp
robertbartel: Basically adopted this, with some slight tweaks since the total nexus count is needed by all ranks.

src/NGen.cpp
```diff
@@ -442,18 +445,61 @@ int main(int argc, char *argv[]) {
...
+        int local_nexus_write_offset, total_nexus_count;
+        int local_nexus_count = nexus_ids.size();
+
+        // Get offset as sum of nexus counts for all ranks before this one
+        MPI_Exscan(&local_nexus_count, &local_nexus_write_offset, 1, MPI_INT, MPI_SUM, MPI_COMM_WORLD);
+        // Above not defined for 0, so manually set
+        if (mpi_rank == 0)
+            local_nexus_write_offset = 0;
+
+        // Calc total count in the last rank based on it's (max) offset, then broadcast to the others
```
PhilMiller:
```suggestion
        // Calc total count in the last rank based on its (max) offset, then broadcast to the others
```

include/utilities/PerFormulationNexusOutputMgr.hpp
robertbartel: It's used beyond just MPI execution for the moment.  I'll rename and document the relationship to MPI rank.

include/utilities/PerFormulationNexusOutputMgr.hpp
robertbartel: I think so.  

The dimensions for the _flow_ NetCDF variable are `{nexus_id_dim, time_dim}`.

And I have the `flow_var_chunk_size_per_dim` array set  as  `{total_nexus_count , 1}`.  

So, when defining chunk sizes, I expect  passing `flow_var_chunk_size_per_dim` for `chunksizesp`  will cause each chunk to have `total_nexus_count` data points, with these having different placement in the nexus id dimension but the same placement in the time dimension.

FWIW, to sanity check, I tried reversing the elements of `flow_var_chunk_size_per_dim` and got a `NC_EBADCHUNK` (bad chunk size) error.

include/utilities/PerFormulationNexusOutputMgr.hpp
robertbartel: I renamed `rank` and tried to make the comments a little more clear (may still be pending a push).

robertbartel:
I think I've updated everything @PhilMiller pointed out, as well as addressed the merge conflict.

test/CMakeLists.txt
```diff
@@ -265,6 +269,39 @@ ngen_add_test(
...
+########################## Specialized MPI Tests
+if (NGEN_WITH_MPI)
+    find_package(MPI REQUIRED)
+    add_executable(test_mpi_unit
+            gtest_mpi_main.cpp                          # Special test main() for use with MPI
+            utils/PerFormulationNexusOutputMgr_Test.cpp # Involved test files
+    )
+    # TODO: this might need more
+    target_include_directories(test_mpi_unit PUBLIC include/utils ${MPI_INCLUDE_PATH})
+    target_link_libraries(test_mpi_unit PUBLIC gtest NGen::core NGen::core_nexus NetCDF ${MPI_LIBRARIES})
```
PhilMiller:
```suggestion
    target_include_directories(test_mpi_unit PUBLIC include/utils)
    target_link_libraries(test_mpi_unit PUBLIC gtest NGen::core NGen::core_nexus NetCDF MPI::MPI_C)
```
Preferable to use the CMake targets rather than explicit variables.

src/NGen.cpp
```diff
@@ -442,18 +445,61 @@ int main(int argc, char *argv[]) {
...
+
+        #endif
+    }
+    else {
+        #if NGEN_WITH_MPI
+        nexus_outputs_mgr = std::make_shared<utils::PerNexusCsvOutputMgr>(nexus_ids, manager->get_output_root());
         #else
-        nexus_outfiles[id].open(manager->get_output_root() + id + "_output.csv", std::ios::trunc);
+        nexus_outputs_mgr = std::make_shared<utils::PerNexusCsvOutputMgr>(nexus_ids, manager->get_output_root());
         #endif
```
PhilMiller:
```suggestion
        nexus_outputs_mgr = std::make_shared<utils::PerNexusCsvOutputMgr>(nexus_ids, manager->get_output_root());
```
Both arms of the #if seem to be identical.

include/utilities/PerFormulationNexusOutputMgr.hpp
PhilMiller, line 1: It would be nice to have the bulk of the code moved out of a header, and over to a stand-alone source file. Cut down on compile-time dependencies and compilation time.

test/CMakeLists.txt
robertbartel: Yeah, don't remember exactly why I did that, though I never got this working how I wanted, so I know I resorted to some trial and error for sanity checking.

src/NGen.cpp
robertbartel: Ah, thanks.  Overlooked this when I did some previous tweaking.

include/utilities/PerFormulationNexusOutputMgr.hpp
robertbartel: I thought about this too after things got to a certain level of complexity.  I'd held off mostly for PR scope reasons, but I can start working on that.

PhilMiller:
I asked Claude to give its own take on this PR, and there's at least one hit I agree with. Will post the comments individually

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,797 @@
...
+         *
+         */
+        void open_netcdf_file() {
+            // Bail with error if this has already been run
+            if (netcdf_file_id != -1) {
+                throw std::runtime_error("Cannot open netCDF file after already have nc_id set.");
+            }
+            // While I thought about retries, because things are serial in this scenario, if it doesn't work the
+            // first time, it still won't for the second, etc.
+            int nc_status = nc_open(nexus_outfile.c_str(), NC_NETCDF4 | NC_NOCLOBBER, &netcdf_file_id);
```
PhilMiller: These flags are the wrong ones for `nc_open`. I think it should just be `NC_WRITE`
https://docs.unidata.ucar.edu/netcdf-c/current/group__datasets.html#gaccbdb128a1640e204831256dbbc24d3e

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,797 @@
...
+
+            // For just obj_id 0, write the time value
+            if (obj_id == 0) {
+                long epoch_minutes = current_epoch_time / 60;
+                const size_t start_t = static_cast<size_t>(current_time_index);
+                const size_t count_t = 1;
+                // TODO: (later) consider if we need to sanity check that times are consistent across obj_ids (we were
+                // TODO:        effectively assuming this to be the case when not explicitly writing times).
+
+                nc_status = nc_put_vara_long(netcdf_file_id, nc_var_id_time, &start_t, &count_t, &epoch_minutes);
```
PhilMiller: Claude noted an inconsistency in that `nc_var_id_time` is added as `NC_UINT`, but written as `long`. The NetCDF library may do the necessary type coercion, but we shouldn't rely on that. Please make them match.

include/utilities/PerFormulationNexusOutputMgr.hpp
robertbartel: Ah, yup, those are create flags.  Fixing.

PhilMiller:
This one's from Claude, without my full consideration, since I didn't dig into the config logic:

```
 3. per_formulation_nexus_files: false throws an error on non-NetCDF/non-parallel builds

  if (per_formulation_setting) {
      #if !NGEN_WITH_NETCDF
      throw std::runtime_error("per_formulation_nexus_files is set to true, but...");
      #elif NGEN_WITH_MPI and !NGEN_WITH_PARALLEL_NETCDF
      throw std::runtime_error("...parallel NGEN was built without parallel NetCDF...");
      #else
      use_per_formulation_nexus_files = per_formulation_setting->get_value<bool>();
      #endif
  }
  The compile-time guard fires before the boolean value is read. A config with "per_formulation_nexus_files": false on a non-NetCDF build throws an
  error, even though false should be a no-op. The value should be read first, and the error only thrown if it's true.
```

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,797 @@
...
+            }
+            else if (current_epoch_time != data_time_marker.epoch_time) {
+                throw std::runtime_error(
+                    "Cannot receive data for formulation " + formulation_id + " for nexus " + nexus_id +
+                    ": expected '" + std::to_string(current_epoch_time) + "' for epoch time at current time " +
+                    "index " + std::to_string(current_time_index) + " but got '"
+                    + std::to_string(data_time_marker.epoch_time) + "'.");
+            }
+
+            current_nexus_data[nexus_data_indices[nexus_id]] = flow_data_at_t;
```
PhilMiller: Per Claude:
```
  4. Silent data corruption in receive_data_entry() on unknown nexus ID

  current_nexus_data[nexus_data_indices[nexus_id]] = flow_data_at_t;
  std::unordered_map::operator[] default-constructs (inserts 0) for missing keys. An unrecognized nexus_id silently overwrites current_nexus_data[0] —
  corrupting data for the first nexus with no error or warning. Should use .at() or a .find() guard.
```
Given all the other careful error checking, this seems apt.

PhilMiller:
```
  5. Same silent-insert problem in PerNexusCsvOutputMgr::receive_data_entry()

  nexus_outfiles[nexus_id] << data_time_marker.sim_time_index << "...";
  If nexus_id wasn't registered in the constructor, operator[] creates a default-constructed (closed) ofstream. Writing to it fails silently — data is
  lost without any error.
```
We maybe don't care to bother there, since that's not doing much if any error checking.

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,797 @@
...
+         * if the @ref netcdf_file_id member variable is already set (to something other than ``-1``) when this function
+         * is called.
+         */
+        void create_netcdf_file() {
+            // Bail with error if this has already been run
+            if (netcdf_file_id != -1) {
+                throw std::runtime_error("Cannot create netCDF file after already created and have nc_id set.");
+            }
+            std::cout << "Creating nexus NetCDF file '" << nexus_outfile << "' for regular access." << std::endl;
+            int nc_status = nc_create(nexus_outfile.c_str(), NC_NETCDF4 | NC_NOCLOBBER, &netcdf_file_id);
```
PhilMiller: Claude noted that the use of `NOCLOBBER` interferes with re-runs, and will error out here. That will be an issue for hotstarts and restarts from checkpoints as well.

Maybe `create_netcdf_file[_parallel]` will need a flag argument to accept `NC_EEXIST`, or a different exception should be thrown in that specific case, so that the caller can behave appropriately.

PhilMiller: In particular, if the file already exists, and we want to extend it, then the call to `setup_netcdf_metadata` needs to be skipped, or optionally replaced by code to check that the metadata match what we expect. Besides variables and dimensions matching, we would potentially want to check that the nexus IDs all match, and where we're writing in the time dimension, and whether that aligns nicely with the data already present (probably, e.g. immediately following)

include/utilities/PerFormulationNexusOutputMgr.hpp
PhilMiller: Claude called this out too.

include/utilities/PerFormulationNexusOutputMgr.hpp
```diff
@@ -0,0 +1,797 @@
...
+                nc_var_id = &nc_var_id_time;
+            }
+            else if (var_name == nc_var_name_flow) {
+                nc_var_id = &nc_var_id_flow;
+            }
+            else {
+                throw std::runtime_error("Can't set NC_COLLECTIVE for invalid variable name '" + var_name + "'.");
+            }
+            set_nc_var_parallel_collective(*nc_var_id);
+        }
```
PhilMiller: This overload is dead code.

I'd vaguely thought that to be the case, and didn't care to look. Claude said as much, and I confirmed.

include/utilities/PerFormulationNexusOutputMgr.hpp
robertbartel: So, probably a good catch, but this gets a little complicated (although I could be overthinking it).  

NetCDF variable types' sizes are apparently platform independent, but NetCDF only has variable setting functions for types with sizes that are platform dependent (e.g., `nc_put_vara_long`).

I'm going to do this:
* change the time NetCDF variable to be of `NC_INT64` type
* change the referenced code to use `nc_put_vara_longlong`
* throw an exception during the constructor if `sizeof(long long) != 8`

@PhilMiller, any thoughts or objections?

PhilMiller: That sounds good to me

include/utilities/PerFormulationNexusOutputMgr.hpp
robertbartel: It's trivial to just switch both of these to `at()` instead of `[]`, so I'll go ahead and do that.

include/utilities/PerFormulationNexusOutputMgr.hpp
robertbartel: I really think this is out of scope for the initial PR.  This might require a more nuanced discussion, which maybe we do in a different forum than the comments.  For now, I'll just note this seems too larger and not needed enough (right now) to try to do here.  

I'll ping @hellkite500 and @aaraney on this too, just in case they want to give feedback.  If they or you @PhilMiller disagree, I'll set up some time for several of us to discuss.

include/utilities/PerFormulationNexusOutputMgr.hpp
robertbartel: Yeah, it wasn't at one point, but things evolved, and I left it in since it was already written.  But cleaning things up makes sense.

include/utilities/PerFormulationNexusOutputMgr.hpp
hellkite500: I'm in favor of opening an issue and handling this in a future PR.  We will need to decide on how we want this to work, and as I think about it, we probably need a couple different possible strategies, depending on the semantic of the specific restart.

PhilMiller: Maybe I take it on as part of the AWI work...

hellkite500:
I ran a scale test for 1 month, conus (v2.2) using 2 nodes (384 cores).  Performance wise, the netcdf nexus writing is now on par with the previous one file per nexus (organized under per rank subdirectories on lustre...)

I also ran this through the vtune profiler and it shows some interesting behavior:
```
Top Hotspots
Function                   Module              CPU Time  % of CPU Time(%)
-------------------------  ----------------  ----------  ----------------
PMPI_File_set_view         libmpi.so.12      51202.549s             35.5%
PMPI_Allreduce             libmpi.so.12      44231.477s             30.7%
PMPI_File_write_at_all     libmpi.so.12      11070.648s              7.7%
```
These are now the top 3 hotspots from the 1 month run!

Since it doesn't seem like we are degrading performance, and we have a lot to gain in other areas, I think we can explore these hotspots independently, but I wanted to mention them here.

PhilMiller:
In my estimation, that profile observation would be most consistent with a moderate degree of load imbalance. I'm very confident that's what's going on with the `Allreduce` call, and reasonably confident there's similar synchronization happening in the `File_` calls for it to apply there, too.

I'm excited to note that once we land support for state serialization, we're most of the way to being able to implement dynamic load balancing, too. As Nels knows, this is very much my happy place.

PhilMiller:
So, I just checked all three of those calls are indeed collective. The writing necessarily takes some actual time to push data out to the filesystem, but it's the smallest of those. The other two, accounting for 66% of CPU time, ought to be able to push way down to near-zero, implying the possibility of a 3x walltime speedup. I'm actually somewhat curious where the `MPI_File_set_view` call is happening, since that amount of time implies it's probably every timestep, and that seems excessive.

As a big-hammer alternative to addressing the causes of "synchronized parallel call looks expensive because the CPU spends a lot of time there", we could move to a single-process-writes model, with all of the nexus flow values `MPI_Igather`'d to one rank, with that rank opening and writing to the file without any of the parallel overheads. That may actually work out a lot better, since we're only talking a few megabytes being written per timestep anyway. If we `MPI_Wait` on the gather and do the write one timestep lagged from the calculation, then we can hide the latency of the gather collective, and potentially cut the time spent blocked on this by (in my estimate) ~95%, which basically would net us that 3x speedup.

PhilMiller:
Just as a check on my arithmetic, Nels, that run you're reporting took something like 13-15 minutes wall time?

hellkite500:
> Just as a check on my arithmetic, Nels, that run you're reporting took something like 13-15 minutes wall time?

Roughly, yes...under profile.

The same run without instrumentation was 6m38s

PhilMiller:
Whoa. What profiling tool were you using that doubled execution time? With over 2x overhead, I'd be reluctant to plan optimization work based on its output.

hellkite500:
Intel vtune, running a hotspot profile. I'll double check the exact invocation.

edit:
Confirmed this came from a hostpot collection

```sh
mpirun -n ${SLURM_NTASKS} vtune -collect hotspots -data-limit=5000 -result-dir "${PROFILE_DIR}"
```
The report from both nodes is pretty similar.

Additional context
```
$ vtune --version
Intel(R) VTune(TM) Profiler 2025.4.0 (build 630424) Command Line Tool
Copyright (C) 2009 Intel Corporation. All rights reserved.
$ mpirun --version
Intel(R) MPI Library for Linux* OS, Version 2021.13 Build 20240701 (id: 179630a)
Copyright 2003-2024, Intel Corporation.
```

PR #927, opened 2026-01-13
ci: don't cache python venv

aaraney:
Don't cache python virtual environment in `ngen-build` composite action.

Add support for `uv` python tooling. If `uv` is present in the action environment, it will be used. Otherwise, `python3 -m venv` and `pip` will be used.

The cache has caused flakey test results in repos using this composite action. https://github.com/NOAA-OWP/evapotranspiration/pull/49 is an example.


hellkite500 (dismissed):
Based on the struggle we have had in the past with maintaining a consistent venv across runners/actions, I think this is a good way to go -- especially if we can leverage the performance of uv in building the venv everytime we need it.

PR #931, opened 2026-01-26
Handle hydrofabric case when nexus has no contributing catchments

robertbartel:
While on some level counterintuitive, it is possible (and has been seen in practice somewhat frequently) for a hydrofabric dataset to contain nexuses that have no contributing catchments.  This appears to be a valid state for the nexus and a hydrofabric as a whole.  However, previously ngen did not account for this scenario properly.  There was no recorded upstream flow for such nexuses at any given time step, and a lack of flow data was treated as an error by the nexus class.  

This change updates the nexus class to infer and return a flow data value of `0.0` for the no-contributing-catchments case, reflecting that nothing contributed any flow.

Since the case is a bit counterintuitive though, this change also adds some warning messaging for any nexuses in such a state.  These will only be printed once at the first timestep.

### DIFF
--- a/src/core/SurfaceLayer.cpp
+++ b/src/core/SurfaceLayer.cpp
@@ -15,10 +15,11 @@ void ngen::SurfaceLayer::update_models()
         for(const auto& id : features.nexuses()) {
             #if NGEN_WITH_MPI
             // When running with MPI, only be concerned with the local nexuses
-            if (!features.is_remote_sender_nexus(id) && features.nexus_at(id)->get_contributing_catchments().size() == 0) {
+            if (!features.is_remote_sender_nexus(id) && features.nexus_at(id)->get_contributing_catchments().size() == 0)
             #else
-            if (features.nexus_at(id)->get_contributing_catchments().size() == 0) {
+            if (features.nexus_at(id)->get_contributing_catchments().size() == 0)
             #endif
+            {
                 // Likely this means a flow value of 0.0, but that's dependent on the nexus class implementation
                 std::cout << "WARNING: Nexus "<< id << " has no contributing catchments for flow values!" << std::endl;
             }
--- a/src/core/SurfaceLayer.cpp
+++ b/src/core/SurfaceLayer.cpp
@@ -10,6 +10,21 @@ void ngen::SurfaceLayer::update_models()
     
     Layer::update_models();
 
+    // One the first time step, check all the nexuses and warn user about ones have no contributing catchments
+    if (current_time_index == 0) {
+        for(const auto& id : features.nexuses()) {
+            #if NGEN_WITH_MPI
+            // When running with MPI, only be concerned with the local nexuses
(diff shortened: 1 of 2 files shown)


src/core/SurfaceLayer.cpp
```diff
@@ -10,6 +10,21 @@ void ngen::SurfaceLayer::update_models()
     
     Layer::update_models();
 
+    // One the first time step, check all the nexuses and warn user about ones have no contributing catchments
+    if (current_time_index == 0) {
```
hellkite500: We should be able to push this validation warning into the feature initialization itself, perhaps in the base constructor here?

https://github.com/NOAA-OWP/ngen/blob/a29e64b0c36519a731657f75ed2c82ebe3378921/src/core/nexus/HY_HydroNexus.cpp#L21

and avoid the loop conditional

src/core/nexus/HY_PointHydroNexus.cpp
```diff
@@ -58,11 +58,21 @@ double HY_PointHydroNexus::get_downstream_flow(std::string catchment_id, time_st
...
+        // there are no contributing catchments so there is no flow to release
+        return 0.0;
+    }
+
     else if ( s1 == upstream_flows.end() )
     {
         // there are no recorded flows for this time.
         // throw exception
 
+        std::cout << "No recorded flows for time step " << t << std::endl;
```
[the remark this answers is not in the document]
robertbartel: Done, pending push.

src/core/SurfaceLayer.cpp
```diff
@@ -10,6 +10,21 @@ void ngen::SurfaceLayer::update_models()
...
+        for(const auto& id : features.nexuses()) {
+            #if NGEN_WITH_MPI
+            // When running with MPI, only be concerned with the local nexuses
+            if (!features.is_remote_sender_nexus(id) && features.nexus_at(id)->get_contributing_catchments().size() == 0) {
+            #else
+            if (features.nexus_at(id)->get_contributing_catchments().size() == 0) {
+            #endif
```
PhilMiller:
```suggestion
            if (!features.is_remote_sender_nexus(id) && features.nexus_at(id)->get_contributing_catchments().size() == 0)
            #else
            if (features.nexus_at(id)->get_contributing_catchments().size() == 0)
            #endif
            {
```
So that editors can match up the brace regardless of the #if compilation, and they're not unbalanced.

src/core/SurfaceLayer.cpp
hellkite500: I see your point here, and I think it may be time to introduce some debug macros we build with to include this kind of verification, but can turn it off in production builds that have validated hydrofabric and other invariants that we don't want to check.  This can definitely follow on here, and I think we can accept this bug fix as is.

hellkite500 (approved):

PR #942, opened 2026-03-10
Optimize setting BMI input variables prior to each time step update.

robertbartel:
Optimizing `Bmi_Module_Formulation.set_model_inputs_prior_to_update` by storing certain reused details rather than re-calculating or re-retrieving on each time step.

## Additions

- new `Bmi_Var_Details` struct type to hold some reusable information on BMI variables 
- new member variable vector `Bmi_Module_Formulation.bmi_input_var_details` with info on module BMI input variables, stored in aforementioned struct instances

## Changes

- modifications to `Bmi_Module_Formulation.set_model_inputs_prior_to_update` to lazily init the contents of `Bmi_Module_Formulation.bmi_input_var_details` and then use that to provide info necessary prior to BMI setting operations, rather than re-querying/re-processing that info on every time step.
- fix param docstring and tweak param names for `Bmi_Module_Formulation.set_model_inputs_prior_to_update`

## Testing

1.  Existing test `Formulation_Manager_Test.basic_run_3` does cover usage of this function.  I made temporary local modifications introducing incorrect logic in a few different places in `set_model_inputs_prior_to_update`, and this did induce failures of that test.

## Notes

- For tests of the v2.2 hydrofabric of the 10U VPU (~84,000 catchments/formulation instances) for 30 days, across 24 processors, with a formulation made from 5 individual BMI modules, this reduced `NGen::simulation` time by ~100 seconds.

## Checklist

1.

- [ ] Linux

### DIFF
--- a/doc/BMI_MODELS.md
+++ b/doc/BMI_MODELS.md
@@ -183,6 +183,19 @@ There are some special BMI formulation config parameters which are required in c
 * `fixed_time_step`
   * boolean value to indicate whether this model has a fixed time step size
   * implied to be `true` by default
+* `store_input_variable_metadata`
+  * boolean value, `false` by default
+  * indicates whether BMI input variable metadata should be stored on first look-up and reused, thus optimizing execution
+    * BMI input variables have values set at each time step
+    * certain metadata is required when performing those set operations:
+      * name and mapped alias (if set)
+      * units
+      * memory size and type of individual items (e.g., `int`, `float`, `double`, etc.)
+      * whether the variable is an array, and if so, the number of individual items
+    * for BMI modules that will not change applicable metadata throughout a simulation, this can be stored and reused
+  * the BMI specification does not expressly guarantee metadata values will not change, so configuration must consider the particular BMI module in use
+  * also, these optimizations do result in additional memory usage
+    * while relatively small, can scale significantly for larger simulations, so this may not be usable for situations when memory is constrained
   
 ## BMI Models Written in C
 
--- a/doc/BMI_MODELS.md
+++ b/doc/BMI_MODELS.md
@@ -183,16 +183,16 @@ There are some special BMI formulation config parameters which are required in c
 * `fixed_time_step`
   * boolean value to indicate whether this model has a fixed time step size
   * implied to be `true` by default
-* `store_input_variable_metadata`
+* `cache_input_variable_metadata`
   * boolean value, `false` by default
-  * indicates whether BMI input variable metadata should be stored on first look-up and reused, thus optimizing execution
+  * indicates whether BMI input variable metadata should be saved in memory on first look-up and reused, thus optimizing execution
     * BMI input variables have values set at each time step
     * certain metadata is required when performing those set operations:
       * name and mapped alias (if set)
       * units
       * memory size and type of individual items (e.g., `int`, `float`, `double`, etc.)
       * whether the variable is an array, and if so, the number of individual items
-    * for BMI modules that will not change applicable metadata throughout a simulation, this can be stored and reused
+    * for BMI modules that will not change applicable metadata throughout a simulation, this can be saved in memory and reused
   * the BMI specification does not expressly guarantee metadata values will not change, so configuration must consider the particular BMI module in use
   * also, these optimizations do result in additional memory usage
     * while relatively small, can scale significantly for larger simulations, so this may not be usable for situations when memory is constrained
--- a/doc/BMI_MODELS.md
+++ b/doc/BMI_MODELS.md
@@ -196,6 +196,7 @@ There are some special BMI formulation config parameters which are required in c
   * the BMI specification does not expressly guarantee metadata values will not change, so configuration must consider the particular BMI module in use
   * also, these optimizations do result in additional memory usage
     * while relatively small, can scale significantly for larger simulations, so this may not be usable for situations when memory is constrained
+  * for a multi-BMI formulation, this option is configured on the individual nested modules (i.e., within each nested module's `params`) rather than at the top level of the multi-BMI formulation, so it can be enabled or disabled per nested module
   
 ## BMI Models Written in C
 
--- a/include/realizations/catchment/Bmi_Module_Formulation.hpp
+++ b/include/realizations/catchment/Bmi_Module_Formulation.hpp
@@ -28,14 +28,14 @@ namespace realization {
         std::string name;
         /** The framework's configured alias for the variable. */
         std::string mapped_alias;
+        /** String for the C++ type corresponding to this variable's type. */
+        std::string cpp_type;
         /** Reference to applicable data provided that is source for input values. */
         std::shared_ptr<data_access::GenericDataProvider> provider;
         /** The size of individual items for this variable. */
         int item_size;
         /** The number of items for this variable. */
         int num_items;
-        /** String for the C++ type corresponding to this variable's type. */
-        std::string cpp_type;
 
         Bmi_Var_Details() : Bmi_Var_Details("", "", nullptr, -1, -1, "") { }
 
--- a/include/realizations/catchment/Bmi_Module_Formulation.hpp
+++ b/include/realizations/catchment/Bmi_Module_Formulation.hpp
@@ -32,12 +32,12 @@ namespace realization {
         Bmi_Var_Details(const std::string& name, const std::string& alias, const int item_size, const int num_items, const std::string& cpp_type, const std::string& units)
             : name(name), mapped_alias(alias), cpp_type(cpp_type), units(units), item_size(item_size), num_items(num_items) { }
 
-        Bmi_Var_Details(const Bmi_Var_Details& source) : Bmi_Var_Details(source.name, source.mapped_alias, source.item_size, source.num_items, source.cpp_type, source.units) { }
+        Bmi_Var_Details(const Bmi_Var_Details& source) = default;
 
         friend bool operator<(const Bmi_Var_Details& lhs, const Bmi_Var_Details& rhs) {
-            return std::tie(lhs.name, lhs.mapped_alias, lhs.cpp_type, lhs.units, lhs.item_size,
-                            lhs.num_items) < std::tie(rhs.name, rhs.mapped_alias, rhs.cpp_type, rhs.units,
-                                                      rhs.item_size, rhs.num_items);
+            return std::tie(lhs.name, lhs.mapped_alias, lhs.cpp_type, lhs.units, lhs.item_size, lhs.num_items)
+                   <
+                   std::tie(rhs.name, rhs.mapped_alias, rhs.cpp_type, rhs.units, rhs.item_size, rhs.num_items);
         }
 
         const std::string& get_name() const {
--- a/include/realizations/catchment/Bmi_Module_Formulation.hpp
+++ b/include/realizations/catchment/Bmi_Module_Formulation.hpp
@@ -72,7 +72,7 @@ namespace realization {
         /** String for the C++ type corresponding to this variable's type. */
         const std::string cpp_type;
         /** String for variable's units. */
-        const std::string units;;
+        const std::string units;
         /** The size of individual items for this variable. */
         int item_size;
         /** The number of items for this variable. */
--- a/include/realizations/catchment/Bmi_Module_Formulation.hpp
+++ b/include/realizations/catchment/Bmi_Module_Formulation.hpp
@@ -22,7 +22,7 @@ class Bmi_Cpp_Multi_Array_Test;
 
 namespace realization {
 
-    /** Type to hold some certain details about a BMI module variable that the framework will need used repeatedly. */
+    /** Type to hold some certain details about a BMI module variable that the framework will need to use repeatedly. */
     class Bmi_Var_Details {
 
     public:
--- a/include/realizations/catchment/Bmi_Module_Formulation.hpp
+++ b/include/realizations/catchment/Bmi_Module_Formulation.hpp
@@ -554,7 +554,7 @@ namespace realization {
          *
          * These should be pointers to @ref Bmi_Var_Details instances in @ref known_bmi_input_vars.
          */
-        std::unique_ptr<std::vector<const Bmi_Var_Details*>> bmi_input_var_details;
+        std::unique_ptr<std::vector<Bmi_Var_Details*>> bmi_input_var_details;
 
         /**
          * Vector of data providers for BMI input vars, with the provider at an index corresponding to the var in
@@ -563,7 +563,7 @@ namespace realization {
          * As with @ref bmi_input_var_details, these should be populated during the first call to
          * @ref do_bmi_sets_from_stored_metadata via a nested call to @ref initialize_bmi_input_var_metadata.
          */
-        std::unique_ptr<std::vector<const std::shared_ptr<data_access::GenericDataProvider>>> bmi_input_providers;
+        std::unique_ptr<std::vector<std::shared_ptr<data_access::GenericDataProvider>>> bmi_input_providers;
 
         models::bmi::protocols::NgenBmiProtocols bmi_protocols;
         /**
--- a/include/realizations/catchment/Bmi_Module_Formulation.hpp
+++ b/include/realizations/catchment/Bmi_Module_Formulation.hpp
@@ -23,36 +23,60 @@ class Bmi_Cpp_Multi_Array_Test;
 namespace realization {
 
     /** Type to hold some certain details about a BMI module variable that the framework will need used repeatedly. */
-    struct Bmi_Var_Details {
+    class Bmi_Var_Details {
+
+    public:
+
+        //Bmi_Var_Details() : Bmi_Var_Details("", "", nullptr, -1, -1, "", "") { }
+
+        Bmi_Var_Details(const std::string& name, const std::string& alias, const int item_size, const int num_items, const std::string& cpp_type, const std::string& units)
+            : name(name), mapped_alias(alias), cpp_type(cpp_type), units(units), item_size(item_size), num_items(num_items) { }
+
+        Bmi_Var_Details(const Bmi_Var_Details& source) : Bmi_Var_Details(source.name, source.mapped_alias, source.item_size, source.num_items, source.cpp_type, source.units) { }
+
+        friend bool operator<(const Bmi_Var_Details& lhs, const Bmi_Var_Details& rhs) {
+            return std::tie(lhs.name, lhs.mapped_alias, lhs.cpp_type, lhs.units, lhs.item_size,
+                            lhs.num_items) < std::tie(rhs.name, rhs.mapped_alias, rhs.cpp_type, rhs.units,
+                                                      rhs.item_size, rhs.num_items);
+        }
+
+        const std::string& get_name() const {
+            return name;
+        }
+
+        const std::string& get_mapped_alias() const {
+            return mapped_alias;
+        }
+
+        const std::string& get_cpp_type() const {
+            return cpp_type;
+        }
+
+        const std::string& get_units() const {
+            return units;
+        }
+
+        int get_item_size() const {
+            return item_size;
+        }
+
+        int get_num_items() const {
+            return num_items;
+        }
+
+    private:
         /** The module's publicized name for this variable. */
-        std::string name;
+        const std::string name;
         /** The framework's configured alias for the variable. */
-        std::string mapped_alias;
+        const std::string mapped_alias;
         /** String for the C++ type corresponding to this variable's type. */
-        std::string cpp_type;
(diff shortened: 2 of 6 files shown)


robertbartel:
Refactored and added config support to optionally use these optimization, with the default being to re-fetch things every time as had been done previously.

robertbartel:
> I think all the bones are here to get us what we want. I think we can minimally tweak the design and avoid the majority of the duplicate data problem we were talking about above.

I'd like to avoid getting too much further into the weeds on things in this PR.  Part of the motivation for it originally was that it was a relatively simple change with clear benefits.  There was a memory tradeoff, yes, but one that at a glance appeared acceptable.  It turned out that it also wasn't guaranteed to be correct for all BMI modules, so a little more work to make it optional was necessary.  That had the additional benefit of protecting against the possibility of any unanticipated negative effects on performance that could emerge in different situations.

Any more, though, and I worry this stops being the low-hanging fruit that I was originally targeting.  @aaraney do you feel strongly that the current memory implications compel us to not adopt this in it's current form?  Otherwise, we can always build more on this in the future.

robertbartel:
With NOAA-OWP/nwm-qa#380, it looks like there is a more concrete reason why compute optimizations are desirable right now.  So there is now good reason for getting into the weeds if we need to in order to make this work.  I'll start working on @aaraney's suggestions.

src/realizations/catchment/Bmi_Module_Formulation.cpp
```diff
@@ -625,62 +630,112 @@ namespace realization {
...
+        }
+
+        void Bmi_Module_Formulation::perform_set(const time_t &src_data_start, const time_step_t& t_delta,
+                                                 const int num_items, data_access::GenericDataProvider* provider, const std::string& mapped_alias,
+                                                 const std::string& var_name, const std::string& cpp_type, const std::string& units) const {
+            std::shared_ptr<void> value_ptr;
+            if (num_items != 1) {
+                //more than a single value needed for var_name
+                std::vector<data_type> values = provider->get_values(
+                    CatchmentAggrDataSelector(get_catchment_id(), mapped_alias, src_data_start, t_delta,units)
```
PhilMiller:
```suggestion
                    CatchmentAggrDataSelector(get_catchment_id(), mapped_alias, src_data_start, t_delta, units)
```

src/realizations/catchment/Bmi_Module_Formulation.cpp
```diff
@@ -625,62 +630,112 @@ namespace realization {
...
+                        + std::to_string(num_items) + ", provider returned "
+                        + std::to_string(values.size()) + " items\n");
                 }
-                get_bmi_model()->SetValue(var_name, value_ptr.get());
+                value_ptr = get_values_as_type(cpp_type, values.begin(), values.end() );
+
+            }
+            else {
+                //scalar value
+                double value = provider->get_value(CatchmentAggrDataSelector(this->get_catchment_id(),mapped_alias, src_data_start, t_delta, units));
```
PhilMiller:
```suggestion
                double value = provider->get_value(CatchmentAggrDataSelector(this->get_catchment_id(), mapped_alias, src_data_start, t_delta, units));
```

src/realizations/catchment/Bmi_Module_Formulation.cpp
```diff
@@ -625,62 +630,112 @@ namespace realization {
...
+        }
 
-                } else {
-                    //scalar value
-                    double value = provider->get_value(CatchmentAggrDataSelector(this->get_catchment_id(),var_map_alias, model_epoch_time, t_delta,
-                                                   get_bmi_model()->GetVarUnits(var_name)));
-                    value_ptr = get_value_as_type(type, value);
+        std::shared_ptr<data_access::GenericDataProvider>& Bmi_Module_Formulation::get_provider_for_input_var(const std::string& var_name, const std::string& mapped_alias) {
+            if (input_forcing_providers.find(mapped_alias) != input_forcing_providers.end())
+                return input_forcing_providers[mapped_alias];
```
PhilMiller:
```suggestion
            auto alias_iter = input_forcing_providers.find(mapped_alias);
            if (mapped_iter != input_forcing_providers.end())
                return mapped_iter->second;
```
And ditto below, to avoid searching in the map twice, back to back, for the same key.

PhilMiller (commented):
_

PhilMiller:
Regarding enablement being configurable, I think all of our current BMI models would allow caching their input expectations.

src/realizations/catchment/Bmi_Module_Formulation.cpp
robertbartel: Fixed.

src/realizations/catchment/Bmi_Module_Formulation.cpp
robertbartel: Fixed.

src/realizations/catchment/Bmi_Module_Formulation.cpp
robertbartel: Fixed.

include/realizations/catchment/Bmi_Module_Formulation.hpp
```diff
@@ -464,6 +563,85 @@ namespace realization {
...
+        /**
+         * Initialize the metadata as needed for @ref do_bmi_sets_from_stored_metadata.
+         *
+         * This will populate the @ref bmi_input_var_details member.  As metadata is gather into @ref Bmi_Var_Details
+         * objects, these will be inserted into to the @ref known_bmi_input_vars static member.  Pointers to the values
+         * in that set are then added to @ref bmi_input_var_details.
+         */
+        void initialize_bmi_input_var_metadata();
+
+        void perform_set(const time_t &src_data_start, const time_step_t &t_delta, const int num_items, data_access::GenericDataProvider *provider, const std::string &mapped_alias, const std::string &var_name, const std::string &cpp_type, const std::string &units) const;
```
hellkite500: I'm trying to weigh the tradeoff of having 8 args to this function vs passing a cache struct and in the case of refetch, just creating the struct each time if needed.

doc/BMI_MODELS.md
```diff
@@ -183,6 +183,19 @@ There are some special BMI formulation config parameters which are required in c
 * `fixed_time_step`
   * boolean value to indicate whether this model has a fixed time step size
   * implied to be `true` by default
+* `store_input_variable_metadata`
```
hellkite500: As much of a pain as it may be to adjust throughout this PR, I think this needs a name change to something like `cache_input_variable_metadata`, which more clearly indicates to users and developers (now and in the future) that this really is an *in memory cache*, and not some other kind of storage.

include/realizations/catchment/Bmi_Module_Formulation.hpp
robertbartel: I'll address this in a pending commit.

doc/BMI_MODELS.md
robertbartel: Fair enough.  Done.

robertbartel:
Making note of this, but not sure we need to do something about it here.

We still seem to be have some sporadic issues with the _test_bmi_c_ check on macOS.   While the last retry was good, before that it was failing and giving `AddressSanitizer:DEADLYSIGNAL` errors that look like this:

```
[ DISABLED ] Bmi_C_Adapter_Test.DISABLED_GetGridSize_0_a
[----------] 33 tests from Bmi_C_Adapter_Test (1956 ms total)

[----------] 35 tests from Bmi_C_Formulation_Test
[ RUN      ] Bmi_C_Formulation_Test.Initialize_0_a
=================================================================
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer: nested bug in the same thread, aborting.
```

I've never been able to reproduce something like this locally on my Mac.   And, after a few retries (four this time) and without any further changes to the PR branch, the error doesn't happen and the check passes (FWIW, this last retry, I turned on the extra debugging ... unclear if that made any difference, and I don't remember if I did that in any previous clearing retries in #922 that I discuss below).   

I saw similar  `AddressSanitizer:DEADLYSIGNAL` errors every now and then in check attempts throughout iterating on #922.  I'm not entirely sure they occurred at exactly the same place, and it's tricky to go back and find.  But I'm sure they were `AddressSanitizer:DEADLYSIGNAL` errors on _test_bmi_c_ and macOS, and that they "resolved" themselves just by re-running the check.

include/realizations/catchment/Bmi_Module_Formulation.hpp
```diff
@@ -382,11 +439,26 @@ namespace realization {
...
+        /**
+         * Set member variable indicating whether @ref set_model_inputs_prior_to_update should store and reuse metadata.
+         *
+         * Set the @ref cache_input_variable_metadata member variable, which indicates whether
+         * @ref set_model_inputs_prior_to_update should store and reuse metadata, as opposed to refreshing such data
+         * each time @ref set_model_inputs_prior_to_update is called.
+         *
+         * @param store_input_var_metadata Whether @ref set_model_inputs_prior_to_update should store and reuse metadata
+         */
+        void set_store_input_var_metadata(bool store_input_var_metadata);
```
hellkite500: The config name change is definitely useful, but we should consider whether all the functions and docstrings for __\*store\*__ functions should be renamed.  This is definitely more *internal* now though, and may not be needed, at least upfront.

robertbartel: Yeah, I missed those.  I've gone ahead and done those as well.  I _think_ I've got everything now.

test/realizations/Formulation_Manager_Test.cpp
```diff
@@ -902,6 +1076,84 @@ TEST_F(Formulation_Manager_Test, basic_run_3) {
...
+
+    std::ostream* raw_pointer = &std::cout;
+    std::shared_ptr<std::ostream> s_ptr(raw_pointer, [](void*) {});
+    utils::StreamHandler catchment_output(s_ptr);
+
+    realization::Formulation_Manager manager = realization::Formulation_Manager(stream);
+
+    this->add_feature("cat-52");
+    this->add_feature("cat-67");
+    manager.read(this->fabric, catchment_output);
```
PhilMiller: The added example realization config has specific formulations for these two catchments, which should make the global config in which caching is enabled not applicable.

include/realizations/catchment/Bmi_Module_Formulation.hpp
```diff
@@ -22,6 +22,63 @@ class Bmi_Cpp_Multi_Array_Test;
 
 namespace realization {
 
+    /** Type to hold some certain details about a BMI module variable that the framework will need used repeatedly. */
```
PhilMiller:
```suggestion
    /** Type to hold some certain details about a BMI module variable that the framework will need to use repeatedly. */
```

include/realizations/catchment/Bmi_Module_Formulation.hpp
```diff
@@ -22,6 +22,63 @@ class Bmi_Cpp_Multi_Array_Test;
...
+    class Bmi_Var_Details {
+
+    public:
+
+        //Bmi_Var_Details() : Bmi_Var_Details("", "", nullptr, -1, -1, "", "") { }
+
+        Bmi_Var_Details(const std::string& name, const std::string& alias, const int item_size, const int num_items, const std::string& cpp_type, const std::string& units)
+            : name(name), mapped_alias(alias), cpp_type(cpp_type), units(units), item_size(item_size), num_items(num_items) { }
```
PhilMiller: Could we follow a member naming convention of (e.g.) trailing underscores, and not have members and constructor arguments shadowing each other?

include/realizations/catchment/Bmi_Module_Formulation.hpp
```diff
@@ -22,6 +22,63 @@ class Bmi_Cpp_Multi_Array_Test;
...
+
+        Bmi_Var_Details(const Bmi_Var_Details& source) : Bmi_Var_Details(source.name, source.mapped_alias, source.item_size, source.num_items, source.cpp_type, source.units) { }
```
PhilMiller: Why is this not just

```suggestion
        Bmi_Var_Details(const Bmi_Var_Details& source) = default;
```

include/realizations/catchment/Bmi_Module_Formulation.hpp
```diff
@@ -22,6 +22,63 @@ class Bmi_Cpp_Multi_Array_Test;
...
+
+        friend bool operator<(const Bmi_Var_Details& lhs, const Bmi_Var_Details& rhs) {
+            return std::tie(lhs.name, lhs.mapped_alias, lhs.cpp_type, lhs.units, lhs.item_size,
+                            lhs.num_items) < std::tie(rhs.name, rhs.mapped_alias, rhs.cpp_type, rhs.units,
+                                                      rhs.item_size, rhs.num_items);
```
PhilMiller: Could this be formatted to put the corresponding members parallel to each other, so that it's more visually obvious?

include/realizations/catchment/Bmi_Module_Formulation.hpp
```diff
@@ -464,6 +563,98 @@ namespace realization {
...
+         * @param var_name The BMI variable name as retrievable directly via BMI.
+         * @param mapped_alias The framework's internal mapped alias for this variable.
+         * @return The appropriate data provider
+         */
+        std::shared_ptr<data_access::GenericDataProvider>& get_provider_for_input_var(const std::string& var_name, const std::string& mapped_alias);
+
+        /**
+         * Initialize the metadata as needed for @ref do_bmi_sets_from_stored_metadata.
+         *
+         * This will populate the @ref bmi_input_var_details member.  As metadata is gather into @ref Bmi_Var_Details
```
PhilMiller:
```suggestion
         * This will populate the @ref bmi_input_var_details member.  As metadata is gathered into @ref Bmi_Var_Details
```

test/realizations/Formulation_Manager_Test.cpp
PhilMiller: This is contrary to the comment describing the test

robertbartel: I'll adjust and push shortly.

include/realizations/catchment/Bmi_Module_Formulation.hpp
robertbartel: Also pending push.

include/realizations/catchment/Bmi_Module_Formulation.hpp
robertbartel: I don't think it makes sense to try to start that in just this PR.  It would be beneficial to formalize a code style and have tools in place to enforce it.  That really needs to be it's own task.

include/realizations/catchment/Bmi_Module_Formulation.hpp
robertbartel: Probably no good reason.  Likely things evolved a few times and I just didn't think about simplifying when it got to this form, but honestly I don't remember.  I'll change it now.

include/realizations/catchment/Bmi_Module_Formulation.hpp
robertbartel: Sure.

include/realizations/catchment/Bmi_Module_Formulation.hpp
robertbartel: Will fix momentarily.

include/realizations/catchment/Bmi_Module_Formulation.hpp
```diff
@@ -22,6 +22,63 @@ class Bmi_Cpp_Multi_Array_Test;
...
 
+    /** Type to hold some certain details about a BMI module variable that the framework will need to use repeatedly. */
+    class Bmi_Var_Details {
+
+    public:
+
+        //Bmi_Var_Details() : Bmi_Var_Details("", "", nullptr, -1, -1, "", "") { }
+
+        Bmi_Var_Details(const std::string& name, const std::string& alias, const int item_size, const int num_items, const std::string& cpp_type, const std::string& units)
+            : name(name), mapped_alias(alias), cpp_type(cpp_type), units(units), item_size(item_size), num_items(num_items) { }
```
hellkite500:
```suggestion
        Bmi_Var_Details(std::string name, std::string alias, const int item_size, const int num_items, std::string cpp_type, std::string units)
            : name(std::move(name)), mapped_alias(std::move(alias)), cpp_type(std::move(cpp_type)), units(std::move(units)), item_size(item_size), num_items(num_items) { }
```

Given how this used in the non-caching path, a Bmi_Var_Details struct gets created each time step, and 4 string copies are created by the current constructor.  Using a by-value constructor and `move`, the units and type arguments are actually zero copied since the temporary returned by the functions producing those can be `move`d, leaving only two copies instead of 4.  The other two could be `moved` at the call site if we wanted to really reduce the additional overhead of string processing per step in code that is supposed to reduce it.

robertbartel: I don't think we could use `move` at the call site because of how the variables are reused, but as is I think I follow and this looks good.  I'm going to spend a little more time convincing myself of how this works, but then I'll incorporate.

include/realizations/catchment/Bmi_Module_Formulation.hpp
robertbartel: @PhilMiller, any further discussion needed in this scope, or are you comfortable deferring this suggestion until later?

test/realizations/Formulation_Manager_Test.cpp
```diff
@@ -999,14 +1178,91 @@ TEST_F(Formulation_Manager_Test, basic_run_3) {
...
+        double diff = actual > expected ? actual - expected : expected - actual;
+        ASSERT_LE(diff, error_margin);
+    }
+}
+
+/**
+ * Testing config the same as EX 1 (like in basic_run_1) but with `cache_input_variable_metadata` true for global
+ * formulation config (which is not all formulations for that configuration, as there are two independently specified).
+ */
+TEST_F(Formulation_Manager_Test, basic_run_9) {
```
hellkite500, line 1379: This test mirrors basic_run_1, but both really only show that the code runs to completion, they don't really differentiate what *should* be different about these setups.  Perhaps this should have an additional assertion such as
```cpp
for (const std::string& id : {std::string("cat-52"), std::string("cat-67")}) {
    auto mod = std::dynamic_pointer_cast<realization::Bmi_Module_Formulation>(manager.get_formulation(id));
    ASSERT_NE(mod, nullptr) << id << " should be a BMI module formulation";
    ASSERT_FALSE(mod->is_input_variable_metadata_cached())
        << id << " must not inherit the global cache_input_variable_metadata=true";
}
```
This requires access to the internal cache flag, which interestingly has a public setter (`set_cache_input_var_metadata`) but no getter — it's a private member read directly only via a test friend. Adding a public getter (e.g. `is_input_variable_metadata_cached()`) would let this assertion work and make these tests more meaningful.

include/realizations/catchment/Bmi_Module_Formulation.hpp
PhilMiller: That's fine.

test/realizations/Formulation_Manager_Test.cpp
robertbartel: > ... which interestingly has a public setter ...

For the benefit of future readers:  technically that setter is protected, not public.

It's a fair point though.  I think I'll probably leave the `basic_run_*` tests alone - there is some value I think in just seeing if things run - but I'll add some other tests that cover this.

**Update:** new tests have now been added to cover this in a few scenarios.

robertbartel:
~Testing this after the last rebase, I'm running into unexpected errors at the end of my small test scenario.  Setting as draft for now while I figure those out.~

**Update**: the errors I was seeing appear to have actually only been due to running out of disk space on my test machine ( :man_facepalming: ), so taking back out of _Draft_ status.

robertbartel:
FWIW, retesting a bit today over a handful of runs, I'm seeing between 45 and 50 seconds reduced runtime with these optimizations active, on a 720-time-step simulation that posts `NGen:simulation` values (i.e., excluding init time) between 480 and 490 seconds when the optimizations are not active.  Not quite as much as discussed in the summary, but  close to 10%.

robertbartel:
I've fixed another set of merge conflicts after other PRs got merged.  Would be good to get this wrapped up unless there are still concerns, or this will keep having conflicts as other stuff gets merged ahead of it.

hellkite500 (approved):

hellkite500:
As far as I'm concerned, this is good to merge.  As noted, there are some refinements we can continue to make, but having an explicit configurable switch on/off makes it much easier to integrate where it is useful, and refine in future efforts as needed.

PR #945, opened 2026-03-18
fix: nexus csv output buffering

aaraney:
@hellkite500 and I found this has an undesirable impact on performance.

### DIFF
--- a/src/core/SurfaceLayer.cpp
+++ b/src/core/SurfaceLayer.cpp
@@ -38,7 +38,7 @@ void ngen::SurfaceLayer::update_models()
         double contribution_at_t = features.nexus_at(id)->get_downstream_flow(cat_id, current_time_index, 100.0);
         
         if(nexus_outfiles[id].is_open()) {
-        nexus_outfiles[id] << current_time_index << ", " << current_timestamp << ", " << contribution_at_t << std::endl;
+        nexus_outfiles[id] << current_time_index << ", " << current_timestamp << ", " << contribution_at_t << "\n";
         }
 
         #if NGEN_WITH_MPI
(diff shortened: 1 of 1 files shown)


src/core/SurfaceLayer.cpp
```diff
@@ -38,7 +38,7 @@ void ngen::SurfaceLayer::update_models()
         double contribution_at_t = features.nexus_at(id)->get_downstream_flow(cat_id, current_time_index, 100.0);
         
         if(nexus_outfiles[id].is_open()) {
-        nexus_outfiles[id] << current_time_index << ", " << current_timestamp << ", " << contribution_at_t << std::endl;
+        nexus_outfiles[id] << current_time_index << ", " << current_timestamp << ", " << contribution_at_t << "\n";
```
PhilMiller:
```suggestion
            nexus_outfiles[id] << current_time_index << ", " << current_timestamp << ", " << contribution_at_t << "\n";
```
May as well fix the single line's indentation while you're at it.

aaraney: Done! Thanks for the feedback, @PhilMiller!

PhilMiller:
This pretty clearly moves things in the right direction.

The PR title might be a bit off. The change is to *allow* line buffering, by not forcing a flush at the end of each line.

aaraney:
> This pretty clearly moves things in the right direction.
> 
> The PR title might be a bit off. The change is to _allow_ line buffering, by not forcing a flush at the end of each line.

Fair enough. I took "don't line buffer nexus outputs" to mean outputs are flushed per line, don't do that. I changed the title to better reflect the change set.

PhilMiller (approved):

PR #946, opened 2026-03-18
Suppress spurious macOS 15 AddressSanitizer error reports during BMI module loading

PhilMiller:
The CI failures seen for `test_bmi_c` on macOS were basically AddressSanitizer tripping over its own feet. Suppress error reports from within the stumbling-block function.

## Additions

- AddressSanitizer suppression file in the test suite, naming the particular macOS-specific function involved in the implementation of AddressSanitizer-instrumented `dlopen()`

## Changes

- Reference the ASan suppression file in the particular CI test that consistently fails

## Checklist

1. [x] Passes macOS 15 CI
2. [x] Passes Linux CI

## Notes

I originally tried various in-code changes based on investigatory effort in interaction with Claude. It inadvertently offered an approach that would only work on Linux, and was then able to explain why it didn't work on macOS once that became apparent.

Another alternative approach would be to not compile the BMI C test model with AddressSanitizer. As noted in comments, that would involve a lot more complicated CMake modifications, so I chose to use a suppression file instead.

### DIFF
--- a/.github/workflows/test_and_validate.yml
+++ b/.github/workflows/test_and_validate.yml
@@ -172,6 +172,7 @@ jobs:
 
       - name: run_bmi_c_tests
         run: |
+          export ASAN_OPTIONS=${ASAN_OPTIONS}:detect_odr_violation=0,suppressions=$(pwd)/test/asan_suppressions.txt
           cd ./cmake_build/test/
           ./test_bmi_c
           cd ../../
--- a/src/bmi/AbstractCLibBmiAdapter.cpp
+++ b/src/bmi/AbstractCLibBmiAdapter.cpp
@@ -6,6 +6,22 @@
 
 #include <dlfcn.h>
 
+// Suppress false positive from ASan's own global registration
+// when dlopen()-ing instrumented shared libraries on macOS.
+// See: ASan runtime's AsanApplyToGlobals() calls getsectiondata()
+// in libmacho.dylib, which uses strcmp on Mach-O section names.
+// ASan's intercepted strcmp then reads shadow memory that has
+// already been poisoned (0xf9 = global redzone) by the same
+// initialization sequence, triggering a spurious report.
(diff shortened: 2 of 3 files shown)


PhilMiller:
First amendment was to just fix a typo in the commit message. Second amendment was to fix my mistake of defining an `extern "C"` function inside of `namespace`, such that it wouldn't actually be defined as the unmangled symbol ASan is going to look for.

PhilMiller:
This didn't work, ostensibly because 

> Reason 1: __asan_default_suppressions doesn't override on macOS
> 
> This is the primary reason your __asan_default_suppressions definition had no effect.
On macOS, the ASan runtime is a dynamic library (libclang_rt.asan_osx_dynamic.dylib). macOS uses two-level namespace for dynamic libraries by default — each dylib's symbol references are bound at link time to a specific image. When the ASan runtime internally calls __asan_default_suppressions() in InitializeSuppressions(), it resolves to its own definition, not yours. Your strong definition in the main executable lives in a different image and is never consulted.
> 
> This is different from Linux/ELF, where all shared objects participate in a flat global symbol table and a strong definition in the main executable overrides a weak one in any .so. On macOS, SANITIZER_INTERFACE_WEAK_DEF on Apple platforms does not produce a truly overridable weak symbol — it produces a regular symbol that the runtime always binds to internally.
> 
> MaskRay's blog post on sanitizer interceptors confirms that on Apple platforms, the sanitizer runtime uses dyld's __DATA,__interpose mechanism for function interception rather than symbol override, and the symbol resolution rules are fundamentally different from ELF.

I'll try a file-based suppression for CI. That will be less helpful for us as developers, and for users, since we'll have to set the same if we run into the same problem. The other suggested workaround is to not compile the BMI test library with AddressSanitizer enabled, but I feel like that would involve an unpleasant amount of mucking with CMake files.

aaraney (approved):
Thanks for the investigation and fix, @PhilMiller!

PR #947, opened 2026-03-25
Close nexus csv outputs before routing

aaraney:
#### Problem

Nexus buffers are not flushed before running routing.
Nexus output files are incomplete because of this causing t-route to fail.
This only occurs when `ngen` runs routing.

This issue was introduced in #945.

### DIFF
--- a/src/NGen.cpp
+++ b/src/NGen.cpp
@@ -564,6 +564,11 @@ int main(int argc, char *argv[]) {
 
     } //done time
 
+    // flush and close nexus
+    for (auto& output_file: nexus_outfiles) {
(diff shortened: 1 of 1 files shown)


aaraney:
This PR is just a solution to the problem. If you've got better ways to solve this that are more maintainable, i'm all ears!

PhilMiller (approved):
I think this solution is appropriate. It may actually solve more than you expect, in that holding those streams open might lead t-route to bump up against limits that could be avoided by closing them here.

hellkite500 (approved):

PR #951, opened 2026-04-15
throws when `variables_names_map` is empty

aaraney:
@seyounger discovered that an empty `variables_names_map` causes `ngen` to throw the following:

```
C++ exception with description "variables_names_map is a String, not an object and cannot be referenced as one." thrown in the test body.
```

This PR is just a test to showcase the failure.

I would expect that this would be a no-op.

### DIFF
--- a/test/realizations/catchments/Bmi_C_Formulation_Test.cpp
+++ b/test/realizations/catchments/Bmi_C_Formulation_Test.cpp
@@ -623,6 +623,37 @@ TEST_F(Bmi_C_Formulation_Test, check_mass_balance_frequency_1) {
     ASSERT_THROW(formulation.check_mass_balance(2, 2, "t2"), ProtocolError);
 }
 
+TEST_F(Bmi_C_Formulation_Test, Initialize_0_with_empty_variables_names_map){
+    int i = 0;
+
+    Bmi_C_Formulation formulation(catchment_ids[i], std::make_shared<CsvPerFeatureForcingProvider>(*forcing_params_examples[i]), utils::StreamHandler());
+    std::string realization_config = "{"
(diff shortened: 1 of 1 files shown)


robertbartel:
My $0.02 here ...

Absent a problem I'm not seeing with the workaround, I don't think this behavior should be modified (or at least considered) in isolation (and so these tests should not be added currently).  I'm pretty sure at least _model_params_ behaves the same way.  I do strongly suspect other elements are indeed no-ops if empty, though.

That's also not necessarily to say all these things should be consistent ... that seems reasonable, but it too should be more deeply examined.  And honestly, I suspect the backend for configuration management could use some work.

christophertubbs:
@robertbartel I can't tell if you support this change. Thumbs up or down?

PhilMiller (changes requested):
Austin's own commit message says "DONT MERGE" [sic].

robertbartel:
I think my preference was to not unless/until a more wholistic and thorough analysis is performed, on this and other similar config elements. 

That seemed like it was probably what Austin was thinking, but I wasn't sure at the time (it was marked Draft; on the other hand, there was no obvious path to doing something more, so it wasn't clear whether maybe this was all that was in mind for the PR).

christophertubbs:
I marked it as a draft. I may roll through and mark PRs not intended for merging as drafts. We have a backlog of PRs to merge and I want to separate the "Probably good to go"s from the "Not ready at all"s.

robertbartel:
> I marked it as a draft.

Ah, ok ... yeah, I didn't notice it hadn't until recently been switched to _Draft_.  That does explain a bit more why I felt like I needed to add my comment.

PR #344, opened 2021-11-24
Increase and allow configurability of precision in output from BMI formulations

robertbartel:
This change primarily makes adjustments to the `get_output_line_for_timestep` functions of the BMI formulation types to have controllable precision with respect to the number of decimal places included in the output.  It also sets the default precision to 9 decimal places if not explicitly set.  Previously, the effective precision was 6 decimal places.

PR #959, opened 2026-04-28
Improve unit conversion error reporting

PhilMiller:
Improve how errors in unit conversion are tracked and reported. Each particular mis-match between distinct models and their variables will be logged exactly once. The reports have comprehensive detail allowing for straightforward correction of the underlying modules or configuration.

## Additions

- Define and use a `unit_conversion_exception` type ("UCE") when UDUNITS fails to apply a conversion, storing the provider and variable details
- Catch UCEs at the requester call site, and log the corresponding requester details along with those of the provider
- Keep a set of the already-reported errors so that they don't get repeated

## Changes

- Move `UnitsHelper` out of `core/`, to avoid ODR violation
- Report UCEs from BMI modules, CSV forcings, and NetCDF forcings
- Tests and test models have units corrected to match up
- Separate model `update()` from the existing `get_response()`, which is now specifically used as getting the effective depth of surface runoff
- Normalization of how units of dimensionless quantities are spelled
- Explicitly test CSV forcing reader's handling of when it can't parse units from the field headers
- Lots of preprocessor include cleanups resulting from refactoring

## Testing

1. All integrated tests pass, include some that have been revised and expanded

## Notes

- This has been cherry-picked from the NGWPC codebase with subsequent conflict fixups, and then substantial revisions to address reviewer comments

## Checklist


include/forcing/DataProvider.hpp
```diff
@@ -101,6 +103,33 @@ namespace data_access
...
+                == std::tie(rhs.requester_name, rhs.requester_variable, rhs.provider_name, rhs.provider_variable, rhs.failure_message);
+        }
+    };
+
+    extern std::set<unit_error_log_key> unit_errors_reported;
+
+    struct unit_conversion_exception : public std::runtime_error {
+        unit_conversion_exception(std::string message) : std::runtime_error(message) {}
+        std::string provider_model_name;
+        std::string provider_bmi_var_name;
```
hellkite500: This generic exception assumes a "bmi" relationship that isn't strictly required.  A more generic "provided_var_name" may be appropriate if this exception is applicable to any provider

include/realizations/catchment/Bmi_Multi_Formulation.hpp
```diff
@@ -570,10 +571,19 @@ namespace realization {
             try {
                 auto const& nested_module = data_provider_iter->second;
                 long nested_module_time = nested_module->get_data_start_time() + ( this->get_model_current_time() - this->get_model_start_time() );
-                auto selector = CatchmentAggrDataSelector(this->get_catchment_id(),var_name,nested_module_time,this->record_duration(),"1");
+                auto selector = CatchmentAggrDataSelector(this->get_catchment_id(),var_name,nested_module_time,this->record_duration(),"");
                 //TODO: After merge PR#405, try re-adding support for index
                 return nested_module->get_value(selector);
             }
+            catch (data_access::unit_conversion_exception &uce) {
+                // We asked for it as a dimensionless quantity, "1", just above
```
hellkite500: This comment and logic is counter intuitive -- in fact the "1" has been changed to ""...
and the "no_" convention in this flag just reads backwards to me...

src/realizations/catchment/Bmi_Module_Formulation.cpp
```diff
@@ -151,16 +157,27 @@ namespace realization {
 
                 // Convert units
                 std::string native_units = get_bmi_model()->GetVarUnits(bmi_var_name);
-                try {
-                    UnitsHelper::convert_values(native_units, values.data(), output_units, values.data(), values.size());
+
+                // --- minimal addition: normalize "none"/"" to "1" and skip conversion if equal ---
```
hellkite500: Why repeat this normalization, when `convert_values` has been updated to do the same thing, and with a larger scope of "dimensionless" quantities...

src/realizations/catchment/Bmi_Module_Formulation.cpp
```diff
@@ -151,16 +157,27 @@ namespace realization {
...
-                    #ifndef UDUNITS_QUIET
-                    logging::warning((std::string("WARN: Unit conversion unsuccessful - Returning unconverted value! (\"")+e.what()+"\")\n").c_str());
-                    #endif
+                // -------------------------------------------------------------------------------
+
+                try {
+                    UnitsHelper::convert_values(in_units_norm, values.data(), out_units_norm, values.data(), values.size());
                     return values;
                 }
+                catch (const std::runtime_error& e) {
```
hellkite500: I'm still thinking about how to provide the proper reporting/context under this scenario, so maybe that is the answer...but would it be possible to have the UnitsHelper and convert* functions throw the unit_conversion_exception instead of a `std::runtime_error`, and either:
fill in context in this catch
or
provide more context to the unit helper...

src/core/mediator/UnitsHelper.cpp
```diff
@@ -70,21 +70,39 @@ double UnitsHelper::get_converted_value(const std::string &in_units, const doubl
 
 double* UnitsHelper::convert_values(const std::string &in_units, double* in_values, const std::string &out_units, double* out_values, const size_t& count)
 {
-    if(in_units == out_units){
-        // Early-out optimization
-        if(in_values == out_values){
+    auto is_noneish = [](const std::string& u)->bool {
```
hellkite500: make this a function and use it in both convert_values and get_converted_value -- this would provide consistency and avoid having this logic duplicated in several places.

test/forcing/CsvPerFeatureForcingProvider_Test.cpp
```diff
@@ -188,9 +188,14 @@ TEST_F(CsvPerFeatureForcingProviderTest, TestForcingUnitHeaderParsing)
         {"U2D", "m s-1", "cm s-1"},
         {"V2D", "m/s", "cm s-1"},
         {"TEST", "kg", "g"},
+        // Disable the cases where units won't be parsed and
+        // conversion was previously not expected. The non-conversion
+        // is now being treated as an error, per NGWPC-7604
+#if 0
```
hellkite500: This is a fundamental change of behavior that is counter to the premise of "don't prevent a modeler from doing something".  Being obnoxious with warnings and logs is definitely a good idea, but changing to error conditions in all possible conversion failures should be an opt in mechanic.

hellkite500 (commented):
While I support the general idea behind this, I have some design and implementation concerns that may require a bit of dialogue and/or refactoring to get through.

PhilMiller:
All useful feedback. I'll try to address it shortly.

robertbartel (changes requested):
FWIW, I'm not fully done with review - still trying to make sure I have my head wrapped around how this executes, especially any differences with scalar versus vector values.  But I did want to go ahead and make a few initial comments.

The only clear thing at this point that I can say needs changing is that the _BMIconventions.md_ document needs to be updated to reflect these behavior changes, especially related to unitless and none-ish conventions.

A couple other early impressions that aren’t fully formed yet, but might stimulate some useful discussion …

* This feels a little brittle.  In fairness, that’s probably got a lot to due with the larger design of things around it, although this doesn’t necessarily disqualify the observation from relevance.
* Throwing an(other) exception on every variable read with units issues seems like it could get a little computationally expensive.  Cheaper than the IO overhead of `cerr`-ing every time a warning condition is encountered, but that could be turned off.  It doesn’t look like this exception-based approach can be.

PhilMiller:
My experience with these changes on the NGWPC side is that they ultimately pushed the overall system toward an overall more robust situation, but there was a lot of reconciliation of these warning messages along the way. There's a follow-on PR coming that also optionally specifies the desired units of output variables named in the realization config, which would avoid modeling misinterpretations and a current inconsistency between mosaiced formulations for different catchments that output the same variables but in different units. Maybe I'll put that up as a PR against this branch, so we can see and evaluate it in context.

PhilMiller:
Regarding the exceptions concern, the actual exception `throw` is generally not expected to be much more expensive than a `return`. We are generating a bunch of strings in that path, though, which I can see being more of a concern. I think that can mostly be squeezed out, which I'll try to do.

PhilMiller:
I haven't directly addressed the review comments yet, but I did push an additional commit that better applies and motivates these changes. That change could potentially be made on its own, but it kinda comes with the rest of what's here, so that's how I'm presenting it at the moment. If asked, I could propose that in more isolation instead or in advance.

PhilMiller:
On the string passing overheads in the error path, I think I'm going to push back a little bit on trying to change that right away. I agree it could become a burden. However, a 'healthy' simulation run that's not incurring unit mismatches will never incur that burden. For pre-production (benchmarking, calibration, regionalization, retrospective, etc) and operational use cases, where we're most concerned about performance, we should have hammered out those issues before running. For a scientifically meaningful run, I really don't think there's an excuse for accepting unit mismatches. If we need some targeted adapters for (e.g.) `precip_rate` mixing up mass vs volume, we should build those and push forward.

More broadly, we need to profile and see if/where passing around so many strings actually does cost ngen performance, and start grinding those away.

src/realizations/catchment/Bmi_Module_Formulation.cpp
PhilMiller: Done, and I think it was a good call. We throw a nicely-typed exception from the bottom of the stack, and enrich and re-`throw`n it until it's caught, logged, and execution moves on. No more `std::runtime_error`

include/realizations/catchment/Bmi_Multi_Formulation.hpp
PhilMiller: Comment removed as redundant

The error case here has substantial changes in a follow-on PR to add exactly the output unit specifications anyway, so I'm going to leave this alone.

test/forcing/CsvPerFeatureForcingProvider_Test.cpp
PhilMiller: I've broken the test into two parts, confirming that the expected exception gets thrown, and that we still get unconverted values back.

This modeling need is handled at higher levels, in `Bmi_Module_Formulation::set_model_inputs_prior_to_update` - it will still consume unconverted inputs, it will just complain there with better-explained detail of what seems to be wrong.

src/realizations/catchment/Bmi_Module_Formulation.cpp
```diff
@@ -201,11 +231,10 @@ namespace realization {
...
                     return UnitsHelper::get_converted_value(native_units, value, output_units);
                 }
-                catch (const std::runtime_error& e){
-                    #ifndef UDUNITS_QUIET
-                    std::cerr<<"WARN: Unit conversion unsuccessful - Returning unconverted value! (\""<<e.what()<<"\")"<<std::endl;
-                    #endif
-                    return value;
+                catch (UnitsHelper::unit_conversion_exception& uce){
+                    uce.provider_model_name = get_bmi_model()->get_model_name();
+                    uce.provider_var_name = bmi_var_name;
```
PhilMiller, line 236: I notice I'm missing `unconverted_value` assignment here

PhilMiller: Not actually; the preceding call to `UnitsHelper` filled it in. That could maybe be made more consistent between scalar and array paths if we wanted.

src/realizations/catchment/Bmi_Multi_Formulation.cpp
```diff
@@ -298,31 +299,22 @@ std::string Bmi_Multi_Formulation::get_output_line_for_timestep(int timestep, st
...
-
-        // Do the rest with a leading comma
-        for (int i = 1; i < output_var_names.size(); ++i) {
-            *output_text_stream << delimiter << get_var_value_as_double(0, output_var_names[i]);
+        std::string output_str;
+        for (const std::string& name : get_output_variable_names()) {
+            // Placeholder to request no conversion
+            std::string output_units = "";
+            double value = get_value(CatchmentAggrDataSelector(this->get_catchment_id(), name, 0, 0, output_units), MEAN);
+            output_str += (output_str.empty() ? "" : ",") + std::to_string(value);
```
PhilMiller, line 308: That technically needs to be `delimiter` rather than literal `","`, but it's going away in a subsequent PR soon, so I'm not going to fix it.

src/realizations/catchment/Bmi_Module_Formulation.cpp
hellkite500: This is one of two major issues I was trying to work through with consolidation of this mechanic.  Subtle but mechanical differences in *how* the exception's are handled, and how the reporting is generated and presented.  These are potentially divergent paths which attempt consistency which can be hard to reason about and easy to trip up, especially as the number of location where unit conversion and error reporting may grow (any new provider, output unit conversion, ect).

include/forcing/CsvPerFeatureForcingProvider.hpp
```diff
@@ -163,11 +163,10 @@ class CsvPerFeatureForcingProvider : public data_access::GenericDataProvider
         try {
             return UnitsHelper::get_converted_value(available_forcings_units[output_name], value, output_units);
         }
-        catch (const std::runtime_error& e){
-            #ifndef UDUNITS_QUIET
-            std::cerr<<"WARN: Unit conversion unsuccessful - Returning unconverted value! (\""<<e.what()<<"\")"<<std::endl;
-            #endif
-            return value;
+        catch (UnitsHelper::unit_conversion_exception& uce) {
+            uce.provider_model_name = "CsvPerFeatureProvider " + std::to_string(catchment_id);
```
PhilMiller: Should I drop the catchment ID here, to avoid an error being reported per-catchment?

src/realizations/catchment/Bmi_Module_Formulation.cpp
```diff
@@ -22,15 +22,24 @@ namespace realization {
             if (timestep != (next_time_step_index - 1)) {
                 throw std::invalid_argument("Only current time step valid when getting output for BMI C++ formulation");
             }
-            std::string output_str;
 
+            static bool no_conversion_message_logged = false;
+            if (!no_conversion_message_logged) {
+                no_conversion_message_logged = true;
+                logging::warning("Output variables do not have unit conversion. Capability not yet implemented in ngen.");
+            }
```
PhilMiller, line 30: @hellkite500 Your changes will presumably make this go away.

src/realizations/catchment/Bmi_Module_Formulation.cpp
hellkite500, line 689: This can come later, but if we want consistent unit conversion error handling, then handling the error conversions on array inputs should happen as well.

I would be inclined to put this in an open issue if we punt it here.

include/realizations/catchment/Bmi_Multi_Formulation.hpp
```diff
@@ -397,28 +399,20 @@ namespace realization {
...
             std::string output_name = selector.get_variable_name();
-            time_t init_time = selector.get_init_time();
-            long duration_s = selector.get_duration_secs();
-            std::string output_units = selector.get_output_units();
-            
             // If not found ...
             if (availableData.empty() || availableData.find(output_name) == availableData.end()) {
                 throw std::runtime_error(get_formulation_type() + " cannot get output value for unknown " + output_name + SOURCE_LOC);
             }
-            return availableData[output_name]->get_value(CatchmentAggrDataSelector(this->get_catchment_id(),output_name, init_time, duration_s, output_units), m);
```
robertbartel, line 408: This feels like a stupid question, but:  what assurances are there that `selector`'s  catchment id value will be the same as `this->get_catchment_id()`?  

I'm not necessarily opposed to demanding callers be responsible for ensuring this consistency and then blindly trusting that, but there was a reason it was like this previously.

src/mediator/UnitsHelper.cpp
```diff
@@ -0,0 +1,184 @@
...
+        } else {
+            std::memcpy(out_values, in_values, sizeof(double)*count);
+            return out_values;
+        }
+    }
+
+    std::call_once(unit_system_inited, init_unit_system);
+
+    // Don't catch the UCE here to fill in uce.unconverted_values,
+    // because the caller may be able to more efficiently std::move it
```
hellkite500, line 168: If the assumption is that error conditions aren't on a typical hot path, then this complicates the handling for no significant reason?  If this did catch
```cpp
try {
        auto converter = get_converter(in_norm, out_norm);
        cv_convert_doubles(converter.get(), in_values, count, out_values);
        return out_values;
    } catch (unit_conversion_exception& uce) {
        uce.unconverted_values.assign(in_values, in_values + count);
        throw;
    }
}
```
then the logging/handling reduces simply to a function
```cpp
void UnitsHelper::report_conversion_error(unit_conversion_exception const& uce, std::string const& ...)
```
which every provider/formulation catch block calls and forwards relevant info to.

There's a lot of churn here, so I'll leave that direction as optional refactoring to follow on if we feel like it doesn't immediately benefit.

include/realizations/catchment/Bmi_Multi_Formulation.hpp
PhilMiller: Honestly, I had taken it for granted, without much thought. I'm curious to learn the reason why that's a bad assumption.

src/mediator/UnitsHelper.cpp
PhilMiller: We don't want the error conditions to be on the hot path, but I feel like (again, supporting research or future-us users), they could be, and copying an array in its entirety might be a bridge too far for some use cases. I'm thinking if someone gets into running gridded models, and so is copying something sized to CONUS or an entire coastal domain mesh. Even the pilot run to debug unit issues could be unreasonably slowed.

I'm open to just going along and changing it, though, if that's really what you'd prefer.

src/realizations/catchment/Bmi_Multi_Formulation.cpp
```diff
@@ -298,31 +299,23 @@ std::string Bmi_Multi_Formulation::get_output_line_for_timestep(int timestep, st
...
-        // Do the rest with a leading comma
-        for (int i = 1; i < output_var_names.size(); ++i) {
-            *output_text_stream << delimiter << get_var_value_as_double(0, output_var_names[i]);
+        std::string output_str;
+        time_t model_time = convert_model_time(get_model_current_time()) + get_bmi_model_start_time_forcing_offset_s();
+        for (const std::string& name : get_output_variable_names()) {
+            // Placeholder to request no conversion
+            std::string output_units = "";
+            double value = get_value(CatchmentAggrDataSelector(this->get_catchment_id(), name, model_time, 3600, output_units), MEAN);
+            output_str += (output_str.empty() ? "" : ",") + std::to_string(value);
```
robertbartel, line 308: Using `std::to_string(value)` here instead of the `output_text_stream` will break some previous functionality.  The `output_precision` config field is now being ignored, so output precision can't be set.  

I saw the commit message noting that value formatting is going away - and so this regression is inconsequential - but I'm not sure I follow the rationale for that.  Certainly I don't in the just context of this PR.  That's not to say we _can't_ remove it, and maybe we should (it seems to have already regressed in other, albeit less used places).  But not like this.

And for context, this was a feature specifically requested at one point (wish I had the request recorded somewhere, but can't find that: just #344).  So it's plausible it's getting used.

robertbartel (changes requested):
There are a few minor things - some of the new stuff doesn't have docstring or unit tests - but that's not critical.  There was one more important thing those, which involves a regression of output precision control capabilities.  We may still want to remove that, but I don't think that belongs here.  We can discuss further, though.

src/realizations/catchment/Bmi_Multi_Formulation.cpp
PhilMiller: I'm thinking of this as a transient regression, to be addressed by Nels's #998. That PR deletes `get_output_line_for_timestep` entirely, and does the text formatting in the CSV writer, where it belongs.

include/forcing/CsvPerFeatureForcingProvider.hpp
hellkite500: Ugh, my comment didn't stick correctly, so here it is again.

I think this will clutter logs at scale.  I'm curious if we can just drop the `failure_message` from the set key completely, since the request meta data that created the message should be unique?  This would also prevent two exceptions per error that I think will come out now: the original exception and then one " (repeated) " exception.

If its possible to keep at least one catchment id in the reporting, I think that is helpful, but if it is reporting effectively the same error repeated for all catchments in the domain running *the same model*, then in theory tracing and fixing 1 should fix them all...

src/realizations/catchment/Bmi_Multi_Formulation.cpp
robertbartel: Ug ... rather than hold this up and create a can't-get-there-from-here, I'll adjust things.  I may go open a bug tracking this to be resolved by #998.

robertbartel (dismissed):

include/forcing/CsvPerFeatureForcingProvider.hpp
PhilMiller: Dropped. The actual prints still mentioned a catchment ID, so finding the file shouldn't be too hard

include/realizations/catchment/Bmi_Multi_Formulation.hpp
robertbartel: At the time `get_value` was updated to support the `DataProvider` interface and accept a `CatchmentAggrDataSelector` arg, it was written to use that arg as the basis for a creating a new `CatchmentAggrDataSelector` and then used the former.  And originally, the former got a different id value: `""`.

At the moment, I cannot tell why this was done.  And, it seems to have caused a bug that was later fixed.  But it would have been so easy to just use the param directly, so I have to strongly suspect this was intentional.

robertbartel (approved):
I have a couple of concerns, but they are being tracked separately and do not appear to be imminently problematic.  Otherwise, this looks good and is taking things in the right direction.

PR #960, opened 2026-04-28
Initial Hydrofabric v4.0 support

robertbartel:
Initial Hydrofabric ~v3.0~ v4.0 support while maintaining v2.2 backward compatibility.  Includes version detection, schema aliasing, and synthesized `divides.toid`.

> [!NOTE]
> Hydrofabric v3.0 (not officially released) and v4.0 are both compatible with these changes.  Original plan was for v3.0 support, until it was learned that v4.0 would be the next official release.

> [!NOTE]
> AI tools were used to assist with developing this code.  This was done in a deliberate and tightly-scoped manner. 

## Additions

- `HydrofabricVersion` enum + `detect_version(sqlite3*)` / `detect_version(columns)` helpers to define and detect versions, and now other version-specific helper functions during data read.
- v3.0 _nexus_ table attribute aliasing: `nexus_id`/`nexus_toid` → `id`/`toid` in feature.cpp.                              
- v3.0 `divides.toid` synthesis via join on _divides_ and _flowpaths_ in _read.cpp_.
- `fp-` flowpath prefix support in `HY_Features_Ids`.
- Updating of _partitionGenerator.cpp_.                                                             
- Synthetic GeoPackage fixtures (_example_v2_2_, _example_v4_0_, _example_v4_0_minimal_, _example_v4_0_dangling_, _example_v4_0_extra_col_) plus Python fixture generators in _test/data/geopackage/_.
- New tests in _test/geopackage/GeoPackage_Test.cpp_ covering version detection, v4.0 aliasing, `toid` synthesis, dangling-flowpath, extra-column tolerance, and minimal-subset link resolution. 
- Test fixture diff tool _test/data/geopackage/diff_fixture.py_ and a README under _test/data/geopackage/_.

## Changes

- Have  _read.cpp_  account for different `HydrofabricVersion` values.
- Move creation of feature data `PropertyMap` from `geopackage::build_feature` to `geopackage::read` to limit awareness/impact of `HydrofabricVersion` to just the latter.
- Throw in a change to .gitignore to have it ignore venv/ (might be out of scope but also trivial; but can be removed).

## Testing

1. `test_unit` passes on several CMake generation variants, including both serial and parallel compilations.
2. New _GeoPackage_Test.cpp_ cases cover v2.2 regression, v4.0 happy path, dangling flowpaths,
  extra-column tolerance, minimal-subset, and bad-schema throw behavior.

## Notes

- The `toid` join for _divides_ is not subset-filtered; while not incorrect, this could lead to some use cases being inefficient in execution.

## Checklist

- [ ] Linux


robertbartel:
~I've marked this back to being _draft_ at the moment while a question gets resolved.~

I tested with a sample v3.0 hydrofabric file that @bolotinl provided, but encountered a "0 downstream connections" error.  Basically, looks like `cat-150698` expects (via _flowpaths_) to connect to `nex-148775`, but `nex-148775` isn't in the nexus table.

The question is:  is this a data problem or a software problem?

With a similar situation in v2.2, I'm pretty sure we'd expect ngen to also fail here, which would suggest this is a subsetter/data problem.  But I'm also not sure where a definitive, formal v3.0 spec is documented, so strictly speaking I don't _know_ that this data scenario is invalid.  If anyone has insight on this before I otherwise figure out the answer, please chime in.  

### Update

Taking this out of _draft_ after a conversation with a few folks, where we all were fairly confident this was in fact a data issue.

hellkite500:
I've looked this over and discussed with @robertbartel.  I like the refactor to isolate version semantics!  I would suggest waiting until we can run a test on an official beta release of NWM v3 hydrofabric (should have a VPU available this week...) before merging.

robertbartel:
> I would suggest waiting until we can run a test on an official beta release of NWM v3 hydrofabric (should have a VPU available this week...) before merging.

I did originally agree with this suggestion, but it's unclear exactly when we are going to receive that new example VPU.  We've had the expectation of "probably later today" a few times this week (and perhaps maybe some last week even), but so far it's still not ready (that I know of).  

So, I created a modified version of the hydrofabric that Lauren provided, which basically fabricated the missing nexus data entries and pointed each's `toid` to `null` (one exception, where data was copied from the previous CONUS v3.0 example dataset, but ngen still ends up handling that nexus the same).  That should be valid:  ngen doesn't care about the geometry and with the nexus rows added can now recognize the spots where the network effectively terminates, and so gracefully stop following it.

With that in hand, I was able to execute a small test run of these changes.  It was only 700 time steps over made-up forcings and BMI configs, but it executed cleanly.  Unless we expect structural/schema changes in what is still being worked on, I'm not sure we gain much by continuing to wait for the VPU example.  

I guess one possible caveat to that might be if there are changes in the way that termination of the network is represented, since this seems to be a property of the data rather than schema.  

Regardless, any thoughts here @hellkite500?

aaraney:
Edit: false alarm,I checkout the first commit from the change set instead of the last...

Word of caution, github pr commits are listed in _chronological order_ unlike the commit history... ugh. Happy Monday!


image



~I'm getting the following error running a serial build w/ a v4 HF.~

<details><summary><code>ngen --info</code> output</summary>

```
ngen --info
NGen version: 0.3.0
Build configuration summary:
  Source directory: /src
  Generator: Unix Makefiles
  Build type: Release
  System: Linux
  C Compiler: /usr/bin/cc
  C Flags:
  CXX Compiler: /usr/bin/c++
  CXX Flags: -I/local/include
  Git Commit ID: a67c526f2afb79ef6ab1bdceb324e156805de6ac
  Flags:
    NGEN_WITH_MPI: OFF
    NGEN_WITH_NETCDF: ON
    NGEN_WITH_PARALLEL_NETCDF: OFF
    NGEN_WITH_HDF5: OFF
    NGEN_WITH_SQLITE: ON
    NGEN_WITH_UDUNITS: ON
    NGEN_WITH_BMI_FORTRAN: ON
    NGEN_WITH_BMI_C: ON
    NGEN_WITH_PYTHON: ON
    NGEN_WITH_ROUTING: ON
    NGEN_WITH_ROUTING_TROUTE_BMI: OFF
    NGEN_WITH_TESTS: ON
    NGEN_WITH_COVERAGE: OFF
    NGEN_QUIET: ON
  Extern Models:
    NGEN_WITH_EXTERN_ALL: OFF
    NGEN_WITH_EXTERN_SLOTH: OFF
    NGEN_WITH_EXTERN_TOPMODEL: OFF
    NGEN_WITH_EXTERN_CFE: OFF
    NGEN_WITH_EXTERN_PET: OFF
    NGEN_WITH_EXTERN_NOAH_OWP_MODULAR: OFF
Environment summary:
  Boost:
    Version: 1.86.0
    Include: /src/boost
  NetCDF:
    Version: 4.9.3
    Library: /local/lib/libnetcdf.so
    Library (CXX): /src/build/extern-libs/netcdf-cxx4/libnetcdf-cxx4.a
    Include: /local/include
    Include (CXX): /src/extern/netcdf-cxx4/netcdf-cxx4/cxx4
    Parallel: FALSE
  SQLite:
    Version: 3.49.1
    Library: /local/lib/libsqlite3.so
    Include: /local/include
  UDUNITS2:
    Library: /local/lib/libudunits2.so
    Include: /local/include
  Fortran:
    BMI_FORTRAN_ISO_C_LIB_PATH:
    BMI_FORTRAN_ISO_C_LIB_NAME: OFF
    BMI_FORTRAN_ISO_C_LIB_DIR: OFF
  Python:
    Version: 3.9.25
    Virtual Env: /.venv
    Executable: /.venv/bin/python3.9
    Interpreter Type: Python
    Site Library: /.venv/lib/python3.9/site-packages
    Include: /local/uv/cpython-3.9-linux-aarch64-gnu/include/python3.9
    Runtime Library: /local/uv/cpython-3.9-linux-aarch64-gnu/lib
    NumPy Version: 1.26.4
    NumPy Include: /.venv/lib/python3.9/site-packages/numpy/core/include
    pybind11 Version:
    pybind11 Include: /src/extern/pybind11/include
Runtime configuration summary:
  Python:
    Version: 3.9.25 (main, Oct 31 2025, 22:59:33)
[Clang 21.1.4 ]
    Virtual Env: <none>
    Executable: /.venv/bin/python3
    Site Library: /.venv/lib/python3.9/site-packages
    Include: /local/uv/cpython-3.9-linux-aarch64-gnu/include/python3.9
    Runtime Library: /local/uv/cpython-3.9-linux-aarch64-gnu/lib/python3.9
    NumPy Version: 1.26.4
    NumPy Include: /.venv/lib/python3.9/site-packages/numpy/core/include

```

</details>

```shell
$ ngen /home/ec2-user/hydrofabric/hydrofabric.gpkg all 
/home/ec2-user/hydrofabric/hydrofabric.gpkg all /home/ec2-user/realizations/realization.json

NGen Framework 0.3.0
Building Nexus collection
terminate called after throwing an instance of 'std::out_of_range'
  what():  column -1 out of range of 5 columns
/run.sh: line 15:    84 Aborted                 (core dumped) ngen /home/ec2-user/hydrofabric/hydrofabric.gpkg all /home/ec2-user/hydrofabric/hydrofabric.gpkg all /home/ec2-user/realizations/realization.json
```

gdb output

```shell
$ gdb ngen core

(gdb) bt
#0  0x0000ffff882ee7b8 in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x0000ffff882a6b3c in raise () from /lib64/libc.so.6
#2  0x0000ffff88293074 in abort () from /lib64/libc.so.6
#3  0x0000ffff88590c60 in __gnu_cxx::__verbose_terminate_handler() () from /lib64/libstdc++.so.6
#4  0x0000ffff8858e5bc in __cxxabiv1::__terminate(void (*)()) () from /lib64/libstdc++.so.6
#5  0x0000ffff8858e620 in std::terminate() () from /lib64/libstdc++.so.6
#6  0x0000ffff8858e904 in __cxa_throw () from /lib64/libstdc++.so.6
#7  0x000000000054e3ac in ngen::sqlite::database::iterator::handle_get_index_(int) const ()
#8  0x000000000054e5c0 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > ngen::sqlite::database::iterator::get<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >(int) const ()
#9  0x0000000000550270 in ngen::geopackage::build_feature(ngen::sqlite::database::iterator const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ()
#10 0x0000000000549ad4 in ngen::geopackage::read(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&) ()
#11 0x0000000000414da4 in main ()

```

robertbartel:
Adding a few more updates.  Main thing was the addition of another HF version: a v4 beta version.  This was the original one these changes were built on, which required synthesizing the link between divides and flowpaths because that was not explicitly in the data.  

However, updates are now anticipated for the structure of v4.  The latest test file actually explicitly has a field to link divided to flowpath.  To make processing more efficient, this field will be used for the v4 format, rather than the synthesis.  But, to ensure support for different files, at least for now, the beta version requiring the synthesis will _also_ be supported (in addition to v2.2).  I'm open to reconsidering this if there is a compelling reason why we shouldn't go the route.

Separately, there was also a little bit of cleanup and fixing of somethings that also have been added.

include/geopackage/HydrofabricVersion.hpp
```diff
@@ -0,0 +1,172 @@
...
+ *   - nexus, v4: "nexus_id"; nexus, v2.2: "id".
+ *   - anything else: "id".
+ *
+ * @param[in] version Hydrofabric schema version detected for this load
+ * @param[in] layer Layer name being read (e.g., "divides", "nexus")
+ * @param[in] db Open GeoPackage database; consulted only for v2.2
+ *               divides introspection
+ * @return Column name to use as the feature id
+ */
+std::string get_layer_id_column(HydrofabricVersion version, const std::string& layer, sqlite::database& db);
```
hellkite500, line 107: the db connection could also be `const`

src/geopackage/HydrofabricVersion.cpp
```diff
@@ -0,0 +1,260 @@
...
+        }
+        if (properties.count("toid") == 0) {
+            throw std::runtime_error(
+                "v4 nexus row missing required 'nexus_toid' column"
+            );
+        }
+        if (id.empty()) {
+            throw std::runtime_error("v4 nexus row has empty 'id' value");
+        }
+    } else if (layer == "divides") {
```
hellkite500, line 222: There is just enough nested "if layer, if version" here that I think it may be worth a couple private helper functions, one to update nexus properties, one to update divide properties.

src/geopackage/read.cpp
```diff
@@ -15,6 +18,9 @@ void check_table_name(const std::string& table)
     }
 }
 
+// Hydrofabric tables touched: nexus and divides always; flowpaths only for
+// v4.0beta1 divides->toid synthesis (never for v4.0). Auxiliary tables
+// (network, flowlines, pois, lakes, attribute tables, etc.) are left alone.
```
hellkite500, line 23: I think we have broken the general utility of this function by forcing it to handle only divide and nexus tables/layers.

We will likely need the ability to read a general table of other names (thinking about parameter tables...) It's not clear to me on the this review pass if this is still strictly possible or not...but at the very least, we will have several bespoke "no-ops" that are layer specific buried in this function.

src/geopackage/read.cpp
```diff
@@ -101,6 +99,10 @@ std::shared_ptr<geojson::FeatureCollection> ngen::geopackage::read(
     query_get_layer_geom_meta.next();
     const std::string layer_geometry_column = query_get_layer_geom_meta.get<std::string>(0);
 
+    // Precomputed once so the per-row loop can attribute "toid" via a
+    // hashtable lookup instead of N queries; empty outside (V4_0_BETA1, "divides").
+    std::unordered_map<std::string, std::string> divide_toid_lookup = build_divide_toid_lookup(version, layer, db);
```
hellkite500, line 104: Here's where I think we need to think about the order of operations here and whether we really want "divide toid" semantics explicitly bundled in a general "geopackage read layer" function...

src/geopackage/read.cpp
```diff
@@ -109,16 +111,41 @@ std::shared_ptr<geojson::FeatureCollection> ngen::geopackage::read(
...
-            id_column,
-            layer_geometry_column
-        );
+        std::string id = query_get_layer.get<std::string>(id_column);
+        geojson::PropertyMap properties = build_properties(query_get_layer, layer_geometry_column);
 
-        features.push_back(feature);
+        // No-op for v2.2; for v4.0, aliases nexus_id/nexus_toid to id/toid and
+        // injects the synthesized "toid" on divides rows from divide_toid_lookup.
+        update_property_map_for_version(properties, version, layer, id, divide_toid_lookup);
```
hellkite500, line 119: Same applies here...

src/geopackage/read.cpp
```diff
@@ -109,16 +111,41 @@ std::shared_ptr<geojson::FeatureCollection> ngen::geopackage::read(
...
+            id,
+            layer_geometry_column,
+            std::move(properties)
+        ));
         query_get_layer.next();
     }
 
+    // Aggregate WARN for divides whose toid could not be synthesized, rather
+    // than one line per row.
+    if (is_v4(version) && layer == "divides") {
```
hellkite500, line 132: And warning here definitely feels awkward

src/partitionGenerator.cpp
```diff
@@ -395,6 +395,34 @@ void read_arguments(int argc, char* argv[],
     if (error) exit(-1);
 }
 
+// Attach a SentinelFeature named "wb-<label>-<feature id>" as the
+// destination of every non-sentinel feature in `collection` whose id
+// prefix satisfies `matches_type` and that has no destination feature
+// yet. New sentinels are added to `collection` only after the scan
+// completes, to avoid iterator invalidation from inserting eagerly.
+void attach_sentinels(const geojson::GeoJSON& collection, const std::string& label,
```
hellkite500, line 403: I see where this is extending existing functionality that was specific to the partitioner.  But now that we see a need to handle this for *both* nexus and divide feature types, I would like to think about bringing this into the network/hyfeatures graph at the lower level and ensure this is consistently applied/tagged anywhere we may need to handle this kind of sentinel.  This isn't required for this PR, though.

test/data/geopackage/make_v2_2_fixture.py
```diff
@@ -0,0 +1,257 @@
+#!/usr/bin/env python3
```
hellkite500, line 1: Can we just commit the fixture script and create/remove the example gpkg during test runs?

test/data/realization/vpu10u_v2_2_regression.json
```diff
@@ -0,0 +1,29 @@
+{
```
hellkite500, line 1: is this a real test artifact required for running tests?

test/data/geopackage/make_v2_2_fixture.py
hellkite500: And since we have sqlite available in any geopackage testable setup, we could move these fixtures to C++ code and maintain the mocks/fixtures which make the creation/management one less step (compared to trying to run the python script and env to produce the mock gpkg for each test...)

hellkite500 (changes requested):

robertbartel:
Regarding @hellkite500's most recent set of requested changes, I've done some work to make sure the geopackage code remains general and things are properly separated for the hydrofabric code.  That ended up requiring a bit more effort than I thought to get the way I wanted, so much so that I haven't initially integrated into this PR's branch.  Instead, it is available in a separate draft PR, #1021.  However, the branch for that is a direct descendent of the branch for this PR.

As such, that PR could be reviewed as if it were the result of the aforementioned requested changes.  If this is done, and there is consensus about approving it, I could just fast-forward things here.  But given scope, I figure we might want to do this as-is first, to make the scope smaller (basically, stacking them).

christophertubbs:
This hasn't been looked at in a while. @hellkite500 Are you satisfied with the changes and, if not, would you be ok with leaving the fixes to another PR?

PR #961, opened 2026-04-28
Migrate project to C++17

robertbartel:
MIgrating project to C++ 17 standard, and performing minimal code modifications to allow this. 

> [!IMPORTANT]
> Before starting this, I did hear from @christophertubbs that he had been informed we were clear to move the project to C++17. 

## Changes

- Change top-level project standard to 17.
- Change inter _test_bmi_cpp_ project standard to 17.
- Adjusting commented out line in forcing CMakeLists.txt that referenced 14 to avoid future issues if that is uncommented.
- Updating documentation to reflect the change.
- Explicitly casting return values to `(void)` for several functions marked as `nodiscard` but whose return values were being ignored

## Testing

1. Builds and test targets succeed for multiple build generation variants

## Notes

- The `(void)` casts were the most convenient way to address this, but I'm not sure that is the best approach.  The involve lines seems to be @hellkite500's and the (relatively new) BMI protocol work, so I'm wondering if he might comment a bit on this specifically. 

## Checklist

- [x] Linux

### DIFF
--- a/src/forcing/CMakeLists.txt
+++ b/src/forcing/CMakeLists.txt
@@ -36,4 +36,4 @@ if(NGEN_WITH_PYTHON)
     target_link_libraries(forcing PUBLIC pybind11::embed NGen::ngen_bmi)
 endif()
 
-#target_compile_options(forcing PUBLIC -std=c++14 -Wall)
+#target_compile_options(forcing PUBLIC -std=c++17 -Wall)
--- a/src/forcing/CMakeLists.txt
+++ b/src/forcing/CMakeLists.txt
@@ -35,5 +35,3 @@ if(NGEN_WITH_PYTHON)
     )
     target_link_libraries(forcing PUBLIC pybind11::embed NGen::ngen_bmi)
 endif()
-
-#target_compile_options(forcing PUBLIC -std=c++17 -Wall)
--- a/src/utilities/bmi/mass_balance.cpp
+++ b/src/utilities/bmi/mass_balance.cpp
@@ -29,7 +29,7 @@ namespace models { namespace bmi { namespace protocols {
 
 NgenMassBalance::NgenMassBalance(const ModelPtr& model, const Properties& properties) :
   check(false), is_fatal(false), tolerance(1.0E-16), frequency(1){
-    initialize(model, properties);
+    (void) initialize(model, properties);
 }
 
 NgenMassBalance::NgenMassBalance() : check(false) {}
@@ -183,7 +183,7 @@ auto NgenMassBalance::initialize(const ModelPtr& model, const Properties& proper
     }
     if ( check ) {
         //Ensure the model is capable of mass balance using the protocol
-        check_support(model).or_else( error_or_warning );
+        (void) check_support(model).or_else( error_or_warning );
     }
     return {}; // important to return for the expected to be properly created!
 }
--- a/src/utilities/bmi/mass_balance.cpp
+++ b/src/utilities/bmi/mass_balance.cpp
@@ -183,7 +183,7 @@ auto NgenMassBalance::initialize(const ModelPtr& model, const Properties& proper
     }
     if ( check ) {
         //Ensure the model is capable of mass balance using the protocol
-        (void) check_support(model).or_else( error_or_warning );
+        return check_support(model).or_else( error_or_warning );
     }
     return {}; // important to return for the expected to be properly created!
 }
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,7 +109,7 @@ endif()
 
 # -----------------------------------------------------------------------------
 # Set global C++ options
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED True)
 set(CMAKE_CXX_VISIBILITY_PRESET hidden)
 
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ As we attempt to apply hydrological modeling at different scales, the traditiona
 This framework includes an encapsulation strategy which focuses on the hydrologic data first, and then builds a functional abstraction of hydrologic behavior.  This abstraction is naturally recursive, and unlocks a higher level of modeling and reasoning using computational modeling for hydrology.  This is done by organizing model components along well-defined flow boundaries, and then implementing strict API’s to define the movement of water amongst these components.  This organization also allows control and orchestration of first-class model components to leverage more sophisticated programming techniques and data structures.
 
 
-  - **Technology stack**: Core Framework using C++ (minimum standard c++14) to provide polymorphic interfaces with reasonable systems integration.
+  - **Technology stack**: Core Framework using C++ (minimum standard c++17) to provide polymorphic interfaces with reasonable systems integration.
   - **Status**:  Version 0.1.0 in initial development including interfaces, logical data model, and framework structure.  See  [CHANGELOG](CHANGELOG.md) for revision details.
 
 ## Structural Diagrams
--- a/doc/DEPENDENCIES.md
+++ b/doc/DEPENDENCIES.md
@@ -49,17 +49,17 @@ If CMake is unable to find a compiler automatically, the CMake `CMAKE_C_COMPILER
 
 ### Version Requirements
 
-Project C++ code needs to be compliant with the C++ 14 standard.  Supported compilers need to be of a recent enough version to be compatible.  
+Project C++ code needs to be compliant with the C++ 17 standard.  Supported compilers need to be of a recent enough version to be compatible.  
 
 Additionally, C++ compilers needs to be compatible (ideally officially *tested* as such) with other project C++ dependencies.
 
 #### GCC
 
-Based on [this page](https://gcc.gnu.org/projects/cxx-status.html#cxx14), the C++ 14 support requirement probably equates to a version of GCC \>= version `5.0.0`.
+Based on [this page](https://gcc.gnu.org/projects/cxx-status.html#cxx17), the C++ 17 support requirement probably equates to some GCC version `7` release or later.
 
 #### Clang
 
-The Clang versioning scheme is a little convoluted.  Using the official scheme, Clang 3.4 and later should support all C++ 14 features.
+The Clang versioning scheme is a little convoluted.  Using the official scheme, Clang 5 and later should support all C++ 17 features.
 
 However, Apple likes to apply their own versioning to Clang and LLVM.  The `Apple LLVM version 10.0.1 (clang-1001.0.46.4)` version released with MacOS 10.14.x should do fine.  Recent, earlier version likely will as well, but YMMV.
 
--- a/extern/test_bmi_cpp/CMakeLists.txt
+++ b/extern/test_bmi_cpp/CMakeLists.txt
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.10)
 # Uncomment this and rebuild artifacts to enable debugging
 set(CMAKE_BUILD_TYPE Debug)
 
-set(CMAKE_CXX_STANDARD 14)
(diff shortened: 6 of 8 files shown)


robertbartel:
The protocol failures track:  I likely need to change something similar to what I started in a build target I didn't think to exercise locally. I'll address shortly.

PhilMiller:
Just seeing the headline and description: HUZZAH!

src/forcing/CMakeLists.txt
```diff
@@ -36,4 +36,4 @@ if(NGEN_WITH_PYTHON)
     target_link_libraries(forcing PUBLIC pybind11::embed NGen::ngen_bmi)
 endif()
 
-#target_compile_options(forcing PUBLIC -std=c++14 -Wall)
+#target_compile_options(forcing PUBLIC -std=c++17 -Wall)
```
PhilMiller: We should just delete this line. This should really not appear in the `CMakeLists.txt` file at all.

christophertubbs:
Would the issue really be:

```c++
    if ( check ) {
        //Ensure the model is capable of mass balance using the protocol
        check_support(model).or_else( error_or_warning );
    }
    return {}; // important to return for the expected to be properly created!
}
```

and

```c++
    static auto error_or_warning(const ProtocolError& err) -> expected<void, ProtocolError> {
        // Log warnings, but throw errors
        switch(err.error_code()){
            case Error::PROTOCOL_ERROR:
                throw err;
                break;
            case Error::INTEGRATION_ERROR:
            case Error::UNITIALIZED_MODEL:
            case Error::UNSUPPORTED_PROTOCOL:
            case Error::PROTOCOL_WARNING:
                std::cerr << err.to_string() << std::endl;
                return make_unexpected<ProtocolError>( ProtocolError(std::move(err) ) );
            default:
                throw err;
        }
        assert (false && "Unreachable code reached in error_or_warning");
    }
```

Are we basically saying "run check support, if it's error type is an integrationerror, an uninitializederror, unsupportedprotocol, or protocolwarning, eat it, otherwise raise the error"? Would it make more sense to just return the result of `check_support(model).or_else(error_or_warning)` since `initialize` is also an `expected`?

PhilMiller:
Here's a rundown C++17-standardized language and library features that I think we'll benefit from adopting:
* `__has_include`, which is already used conditionally in the codebase
* `[[nodiscard]]` and `[[maybe_unused]]` attributes
* structured bindings
* `if constexpr`
* Basic parallel algorithms and execution policies [P0024](https://wg21.link/p0024)
* `std::btye`
* `std::string_view`
* `std::optional`
* `std::filesystem`

Sadly, `std::span` comes in C++20

robertbartel:
> Sadly, std::span comes in C++20

Based on my discussion with Chris earlier, it may also be possible to migrate to C++20 very soon, if there is compelling reason to do so.   

The key reason for just C++17 here was because we (or someone) would first need to (re)confirm that suitable compilers are available on certain key systems that can support C++20.  Given that extra requirement, and without being able to immediately think of something for which we required C++20, I figured let's just get to 17.  But it seems we may now have something.

src/forcing/CMakeLists.txt
robertbartel: I've removed this now.  Didn't think too hard about it in the first pass, but I agree.

robertbartel:
Not sure I'm totally following your last comment, @christophertubbs.  The previous check errors were related to discarded-value expressions that involved something marked as `nodiscard`.   With stricter compiler flags set, that caused builds to fail.  You may be trying to get at something else about those functions though; if so, please elaborate.

PhilMiller:
> > Sadly, std::span comes in C++20
> 
> Based on my discussion with Chris earlier, it may also be possible to migrate to C++20 very soon, if there is compelling reason to do so.
> 
> The key reason for just C++17 here was because we (or someone) would first need to (re)confirm that suitable compilers are available on certain key systems that can support C++20. Given that extra requirement, and without being able to immediately think of something for which we required C++20, I figured let's just get to 17. But it seems we may now have something.

Besides `std::span` (for which we currently use Boost's implementation), the other thing I see standardized in C++20 that would be nice for us is (Gregorian) Calendar support in `std::chrono`. That would let us cut way down on mucking with `time_t`, and shift to meaningful representations that we can consume in configs and emit in output files and logs.

PhilMiller:
Of course, we could have been using Boost.Date_Time if we were open to using a non-header Boost library.

PhilMiller:
OK, wow, looking into that, https://www.boost.org/doc/libs/1_70_0/doc/html/date_time/posix_time.html#date_time.posix_time.time_iterators looks incredibly convenient for better encoding what we are doing.

christophertubbs:
> Not sure I'm totally following your last comment, @christophertubbs. The previous check errors were related to discarded-value expressions that involved something marked as `nodiscard`. With stricter compiler flags set, that caused builds to fail. You may be trying to get at something else about those functions though; if so, please elaborate.

Is the issue partly that we're being overzealous with the use of `NODISCARD` and `expected`? Using an `expected` means you're juggling `NODISCARD` objects, so is it appropriate to return a `NODISCARD` value if pretty much the entire purpose is to throw or return an error if the resulting value was invalid? If we _do_ want to collect that error for handling, would it not make more sense to return that error with the return value of the function?

i.e. has the need for the `(void)` fix identified a design issue?

hellkite500:
> > Not sure I'm totally following your last comment, @christophertubbs. The previous check errors were related to discarded-value expressions that involved something marked as `nodiscard`. With stricter compiler flags set, that caused builds to fail. You may be trying to get at something else about those functions though; if so, please elaborate.
> 
> Is the issue partly that we're being overzealous with the use of `NODISCARD` and `expected`? Using an `expected` means you're juggling `NODISCARD` objects, so is it appropriate to return a `NODISCARD` value if pretty much the entire purpose is to throw or return an error if the resulting value was invalid? If we _do_ want to collect that error for handling, would it not make more sense to return that error with the return value of the function?
> 
> i.e. has the need for the `(void)` fix identified a design issue?

This was an _intentional_ design decision, and I'll try to explain it here...

The protocol interface and library is intended to be reasonably self contained, and I adopted `expected` throughout most of it's code paths for error handling.  Since `ngen` utilizes exceptions and exception handling for most of its error mechanics, I intentionally brought the two together via the `NgenBMIProtocols` container.  It effectively delegates calls to run a protocol, captures the `expected` return and handles it via `or_else`, where there is a single path to `throw` if configured as a fatal error.

The intent at the engine level is a minimal surface API that allows a protocol to register and run based on protocol specific configuration, and to that interface effectively no-op when the protocol is either unsupported or some other error occurs that shouldn't interrupt the main simulation (unless specifically configured to do so).

The protocol implementations themselves make heavy use of the expected returns, and keeping the `nodiscard` annotation helps implementations ensure they are handling errors and code paths intentionally.  The path's that Bobby has flagged here showed up due to different compiler support for `nodiscard` class types, as the `expected` type itself is annotated.  On my local setup with a slightly older clang, this warning wasn't emmitted when an expected type wasn't used/bound or intentionally `void`ed. 

The only place I could see changing the semantics slightly in Bobby's proposed fixes is the case in `check_support()`, where we could still handle the error in the `or_else` and return the unexpected back to `initialize`, but since `initialize` is used in construction, dropping the return explicitly with `(void)` makes sense.  The ngen call site fixes are also fine, unless we want to start introducing/evolving the error handling through the engine, which was intentionally out of the scope of the protocol implementations.  If and/or when that decision is made, minor changes to the error paths in the protocol library can remove all `throws` and just propagate errors through the `unexpected` path.

src/utilities/bmi/mass_balance.cpp
```diff
@@ -183,7 +183,7 @@ auto NgenMassBalance::initialize(const ModelPtr& model, const Properties& proper
     }
     if ( check ) {
         //Ensure the model is capable of mass balance using the protocol
-        check_support(model).or_else( error_or_warning );
+        (void) check_support(model).or_else( error_or_warning );
```
hellkite500: After thinking about this a bit, this could return the `or_else` result just the same, and that may be more appropriate, especially for alternative code paths that may use this in the future.
```suggestion
        return check_support(model).or_else( error_or_warning );
```

src/utilities/bmi/mass_balance.cpp
```diff
@@ -29,7 +29,7 @@ namespace models { namespace bmi { namespace protocols {
 
 NgenMassBalance::NgenMassBalance(const ModelPtr& model, const Properties& properties) :
   check(false), is_fatal(false), tolerance(1.0E-16), frequency(1){
-    initialize(model, properties);
+    (void) initialize(model, properties);
```
hellkite500, line 32: because this is called from the constructor, we propagate exceptions if required, but otherwise set internal states to proceed with no-ops when some initialization error occurs, there's nothing we can/should do with the return in this case, and `(void)` seems appropriate.

christophertubbs: Would it be worth doing something like:

```c++
expected<void, std::exception> initialization_status = initialize(model, properties);

if (!initialization_status) {
    std::cerr << initialization_status.error().what() << '\n';
}
```

No output dropping _and_ messages end up going _somewhere_. Just a rough wag - syntax is probably off. What I mean by possible design issue is that we're missing out on possible uses of the output. It _should_ be returning an `expected`, that part is good, the fact that we're not utilizing its functionality is what got me curious. The `return` rather than just the `(void)` catching is somewhat what I meant.

robertbartel:
@christophertubbs, @hellkite500, we seem to have conflated improving the BMI protocol code with moving to C++17, in a way that I don't think is absolutely necessary and seems to have gone beyond trivial.   We could get into a nuanced argument about scope, but I think it'd be reasonable and easier to just separate any further changes to the protocol implementation to there own task.

christophertubbs:
I like Nels' suggested change for returning the expected, but I think it's good all around. I'd feel better if someone with more experience with more nuanced understanding of C++ rubber stamped the PR, but I'll go ahead and approve if you think I'm just being too cautious.

PhilMiller:
> I like Nels' suggested change for returning the expected, but I think it's good all around. I'd feel better if someone with more experience with more nuanced understanding of C++ rubber stamped the PR, but I'll go ahead and approve if you think I'm just being too cautious.

<Puts on my "WG21 members listen to me" hat>
I guess I can actually dig into these details and assuage the concerns.</hat>

PhilMiller:
My first observation about the `expected` usage in protocols is that the standardized version in C++23 actually omits `[[nodiscard]]`, and this is an option in the nonstd-lite implementation Nels has imported. That obviously wouldn't make the code correct, but the particular diagnostic we're seeing is 'extra' in that sense.

src/utilities/bmi/mass_balance.cpp
PhilMiller: In the case(s) where you're saying `or_else(error_or_warning)`, `return` seems to be clearly appropriate to me.

src/utilities/bmi/mass_balance.cpp
PhilMiller: I think Chris's proposal to output the error message if one exists and wasn't thrown as an exception sounds good. Let's just go with that and get on with merging this PR.

hellkite500: I looked back through the mass balance particularly to see what was happening with error/warning messages, and they actually do get back to cerr via the `error_or_warning` function that called throughout `initialize` and `check_support`.  Adding this here will actually duplicate the message.

robertbartel:
I accepted @hellkite500's suggested change.  It looks like there is still doubt in the other review comment thread, and in the main issue comments, on how best to proceed in other parts.  I still think that's best tracked separately, and I've opened #962 for this.

Even with the scopes separated, we could conceivably block this PR on #962 if any of you feel strongly that the latter needs to be addressed first.   That unnecessary in my opinion, but I'm open to hearing otherwise.

hellkite500 (approved):
I'm good with merging at this point

PR #963, opened 2026-05-09
Add a nexus NetCDF output path that gathers to rank 0 in MPI builds

PhilMiller:
The existing NetCDF nexus output for MPI builds relies on hdf5-parallel support in the underlying NetCDF library. This poses testing/deployment issues, because that's hard to get configured and built correctly. It also poses performance issues due to the substantial synchronization entailed relative to the small volumes of data being transferred. 

## Additions

- In builds with MPI support, NetCDF nexus output can now be gathered to rank 0 and written by that single process using the non-parallel NetCDF API

## Changes

- `NGEN_WITH_PARALLEL_NETCDF` becomes a CMake configuration option, defaulting to off - i.e. in MPI builds, the new gather-to-rank-0 path will be the default

## Testing

`test_unit` and `test_mpi_unit` in builds with
1. non-parallel NetCDF (`test_mpi_unit` exercises the new code)
2. parallel NetCDF and `NGEN_WITH_PARALLEL_NETCDF=OFF` (`test_mpi_unit` exercises the new code)
3. parallel NetCDF and `NGEN_WITH_PARALLEL_NETCDF=ON` (`test_mpi_unit` exercises the existing code)

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,6 +57,7 @@ option(NGEN_QUIET            "Silence output"                     OFF)
 cmake_dependent_option(NGEN_WITH_ROUTING "Build with t-route integration" ON "NGEN_WITH_PYTHON" OFF)
 cmake_dependent_option(NGEN_UPDATE_GIT_SUBMODULES "Update submodules on configure" ON "GIT_FOUND;NGEN_HAS_GIT_DIR" OFF)
 cmake_dependent_option(NGEN_WITH_COVERAGE "Build with test coverage" OFF "NGEN_WITH_TESTS" OFF)
+cmake_dependent_option(NGEN_WITH_PARALLEL_NETCDF "Build with NetCDF using HDF5-parallel support" OFF "NGEN_WITH_NETCDF" OFF)
 
 option(BMI_FORTRAN_ISO_C_LIB_DIR "Directory hint for middleware Fortran shared lib handling iso_c_binding" "${NGEN_ROOT_DIR}/extern/iso_c_fortran_bmi/cmake_build")
 option(BMI_FORTRAN_ISO_C_LIB_NAME "Name for middleware Fortran shared lib handling iso_c_binding" "iso_c_bmi")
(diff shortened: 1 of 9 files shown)


PhilMiller:
Need to update the docs about this, since it relaxes the dependencies, but adds a build-time config option

robertbartel (commented):
Tentatively, I approve.  I do have a few comments, with the first being more immediate and the others more thoughts for the future.

* Before final approval, I think there should be some VPU- or CONUS-sized test experiments (perhaps there have been already?) on suitable systems to sanity check this behaves as expected at scale and to get a practical sense of the performance difference 
* The growing complexity of the design with preprocessor macros and conditional compilation stands out a little to me, but that can be addressed later.
* With Parallel NetCDF it was not possible make use of NetCDF's built in chunk caching, but I think this alternative implementation does not run into the same restriction (see [this link](https://support.hdfgroup.org/documentation/hdf5/latest/group___f_a_p_l.html#ga034a5fc54d9b05296555544d8dd9fe89) that I also reference in code comments)
  * Again, that can wait to looked at until some future task
  * If we do go that route, I think that'll will likely further benefit from all this being reorganized using more than just preprocessor macros

robertbartel:
I've done some relatively small and simple comparison tests.  These were using the v2.2 10U VPU (~84,000 catchments, ~40,000 nexuses), 24 CPUs, and 720 time steps.  I used a local workstation, not a HPC system.

| Method | Output `NGen::simulation` Time |
|---- | ----|
| Previous _Parallel NetCDF_ Method | 504.785 |
| _Gather to Rank 0_ Method | 490.22 |

So this is about 3% faster.  If that holds on other systems (I would expect it to at least be close), that's pretty significant on larger/longer simulations, on top of being easier to set up dependency-wise.  I'll wait a little longer before I approve to give @hellkite500 or @aaraney a chance to chime in, but at this point I'm satisfied.

As a side note, there is very likely more performance to be gained in the _Parallel NetCDF_ method.  Just deactivating the `NC_COLLECTIVE` variable access, which is currently hard-coded to be on, would reduce the _Parallel NetCDF_ method to 470.271.  So we should continue to look more at both of these in the future.

PhilMiller:
I'll readily acknowledge the preprocessor cruft bothers me too. I just don't see a much better way around it at the moment. Maybe it could be factored out at least partially.

In an overall assessment, I was motivated to actually get this implemented and posted by Nels describing the configuration/build difficulties implied by the existing NetCDF-on-HDF5-parallel implementation. If the performance is equal up to CONUS scale, and we don't have any of that complexity, I count it as a win.

PhilMiller:
And I do expect to see performance differences be bigger at larger scales of runs and larger degrees of parallelism, especially distributed parallelism across compute hosts.

hellkite500 (approved):

PR #967, opened 2026-05-13
Core and Formulation refactor

PhilMiller:
Various refactorings and rearrangements in advance of factoring out much of the dynamic simulation-driving code from `NGen.cpp` `main()`

## Changes

- Change `Layer` and `SurfaceLayer` to take flow output variable arguments and set values in them
- Allocate vectors to receive the data from those output arguments in `main()`, to be encapsulated in the follow-on refactoring.
- Moved code from headers to `.cpp` files for `Layer`, `Formulation`, and `Catchment_Formulation`
- Reformatting where it preceded substantive changes, to avoid merge conflicts

## Testing

1. Local runs of ctest on most commits
2. CI jobs

## Notes

- The commits are structured as they are, rather than a cleaner roll-up of the substantive changes they sum to, to maintain direct correspondence with their sources in the NGWPC `development` branch. This makes it easier for me to track what's already been incorporated as I pull in additional changes. It also maintains accurate attribution of relevant parts of the encompassed work to Ian Todd.

## Checklist


src/core/Layer.cpp
```diff
@@ -0,0 +1,82 @@
...
+        catch(std::exception& e){
+            std::string msg = e.what();
+            msg = msg+" at timestep "+std::to_string(output_time_index)
+                +" ("+current_timestamp+")"
+                +" at feature id "+id;
+            throw std::runtime_error(msg);
+        }
+#if NGEN_WITH_ROUTING
+        int results_index = catchment_indexes[id];
+        catchment_outflows[results_index] += response;
```
PhilMiller, line 44: There may be a unit bug here, depending on what we ultimately want. The `response` value comes as a depth in meters, not a mass or volume flux as computed below. Maybe this should get moved down below and use `response_m_s` or `response_m_h`, or maybe we keep the raw depth value.

The values in `catchment_outflows` are not currently used, so there will be no impact right now.

src/NGen.cpp
```diff
@@ -525,90 +540,130 @@ int main(int argc, char* argv[]) {
...
+    std::vector<double> nexus_downstream_flows;
+#if NGEN_WITH_ROUTING
+    size_t catchment_collection_size = catchment_collection->get_size();
+    catchment_outflows.resize(catchment_collection_size * num_times, 0.0);
+    for (int i = 0; i < catchment_collection_size; ++i) {
+        auto feature = catchment_collection->get_feature(i);
+        std::string feature_id = feature->get_id();
+        catchment_indexes[feature_id] = i;
+    }
+    nexus_downstream_flows.resize(nexus_indexes.size() * num_times, 0.0);
```
PhilMiller, line 603: As Nels @hellkite500 notes, these vectors will be unreasonably large for runs longer than typical forecasts, like the 2+ years used for calibration, retrospective, etc. We're going to work out what to do with them in this PR in the interim.

PhilMiller: I've just disabled the allocation and accumulation of these vectors, since they're not used in the code as it exists right now. We can work out their handling in the next round, once they potentially become useful.

src/realizations/catchment/Catchment_Formulation.cpp
```diff
@@ -0,0 +1,62 @@
...
+    {
+        // Assume the catchment ID is equal to or embedded in the formulation `id`
+        size_t idx = id.find(".");
+        set_catchment_id( idx == std::string::npos ? id : id.substr(0, idx) );
+    }
+
+    void Catchment_Formulation::config_pattern_substitution(geojson::PropertyMap &properties, const std::string &key,
+                                                            const std::string &pattern, const std::string &replacement)
+    {
+	std::stringstream ss;
```
PhilMiller: Stray here

include/core/Layer.hpp
```diff
@@ -101,82 +102,11 @@ namespace ngen
...
-            {
-                simulation_time.advance_timestep();
-            }       
-        }
-        
+        virtual void update_models(boost::span<double> catchment_outflows, 
+                                   std::unordered_map<std::string, int> &catchment_indexes,
+                                   boost::span<double> nexus_downstream_flows,
+                                   std::unordered_map<std::string, int> &nexus_indexes,
+                                   int current_step);
```
PhilMiller, line 109: I just noticed that the `current_step` parameter is unused in both implementing classes. I think this is a relic of when I was passing the entire vectors as output arguments, and these classes were expected to compute the correct offset for the step being run.

In the interest of avoiding merge conflicts for the time being, I'm going to leave that alone, but will plan to come back and clean it up.

include/core/Layer.hpp
```diff
@@ -101,82 +102,11 @@ namespace ngen
...
-            } //done catchments   
-
-            ++output_time_index;
-            if ( output_time_index < simulation_time.get_total_output_times() )
-            {
-                simulation_time.advance_timestep();
-            }       
-        }
-        
+        virtual void update_models(boost::span<double> catchment_outflows,
```
hellkite500, line 105: It seems evident that exfiltrating the catchment and nexus flows are intended to support some form of routing integration, but it seems odd that both are needed -- the nexus flows *are* the accumulated overland flow rates contributing to some nexus location.  I can maybe think of some use cases where having the raw un aggregated values could be used, but it doesn't fit.  I'm not completely sold on this interface expansion as exists, but since its not actually active given the disabling of populating these spans, it can likely exist for now.  Encapsulating the indexing and the data view into a single structure might be worth considering for future use of this generalizable concept.

src/core/Layer.cpp
```diff
@@ -0,0 +1,83 @@
...
+            std::string msg = e.what();
+            msg = msg+" at timestep "+std::to_string(output_time_index)
+                +" ("+current_timestamp+")"
+                +" at feature id "+id;
+            throw std::runtime_error(msg);
+        }
+#if NGEN_WITH_ROUTING && false
+        int results_index = catchment_indexes[id];
+	// XXX: This is currently accumulating in meters of depth, which may not be desirable
+        catchment_outflows[results_index] += response;
```
hellkite500, line 44: Its definitely not obvious why this *accumulates*

src/core/SurfaceLayer.cpp
```diff
@@ -26,45 +26,53 @@ void ngen::SurfaceLayer::update_models()
...
-        }
+        // Ensures only one side of the dual sided remote nexus actually does this
+        if (features.is_remote_sender_nexus(id)) continue;
         #endif
 
-        //Get the correct "requesting" id for downstream_flow
+        // Get the correct "requesting" id for downstream_flow
         const auto& nexus = features.nexus_at(id);
         const auto& cat_ids = nexus->get_receiving_catchments();
+
```
hellkite500, line 47: mildly redundant given the network checks before features are processed here...it may be that non dendritic networks are used at some point through different drivers, in which case the checks below would need to handle what to do in those cases, I suppose.  Doesn't hurt to have this, but it somewhat implies a constraint that may not be strict...probably worth some documentation/notes and not a code change at this point

src/core/SurfaceLayer.cpp
```diff
@@ -26,45 +26,53 @@ void ngen::SurfaceLayer::update_models()
...
+            // This is a terminal node, SHOULDN'T be remote, so ID shouldn't matter too much
             cat_id = "terminal";
         }
 
         //std::cerr << "Requesting water from nexus, id = " << id << " at time = " <<current_time_index << ",  percent = 100, destination = " << cat_id << std::endl;
         double contribution_at_t = features.nexus_at(id)->get_downstream_flow(cat_id, current_time_index, 100.0);
+
+#if NGEN_WITH_ROUTING && false
+        int nexus_index = nexus_indexes[id];
+        nexus_downstream_flows[nexus_index] += contribution_at_t;
```
hellkite500, line 68: again, not clear why this *accumulates*

src/NGen.cpp
```diff
@@ -444,9 +444,24 @@ int main(int argc, char* argv[]) {
...
-    //TODO don't really need catchment_collection once catchments are added to nexus collection
-    //Still using  catchments for geometry at the moment, fix this later
-    //catchment_collection.reset();
+    // TODO don't really need catchment_collection once catchments are added to nexus collection
+    // Still using  catchments for geometry at the moment, fix this later
+    // catchment_collection.reset();
+
+    // T-ROUTE data storage
+    std::unordered_map<std::string, int> nexus_indexes;
+#if NGEN_WITH_ROUTING
```
hellkite500, line 453: Should this similarly be disabled for now?

include/core/Layer.hpp
PhilMiller: It's not *just* routing integration, it's also file output similar to what exists now (i.e. CSV-per-feature), and other consumptive uses of those fundamental atoms of NextGen simulation result.

Shortly down the line, `NgenSimulation` does actually provide some of that encapsulated "indexing and data view [in] a single structure". The exposed interface for retrieval of values is by ID, not by arbitrary array index.

For now, for the sake of integrating the rest of the dependent changes, I'm going to ask you to give me some slack on this increment.

src/core/Layer.cpp
PhilMiller: IIRC, it's with a mind toward accommodating multiple `Layer`s feeding into a single feature. That specific thing isn't fully baked in.

It's more baked for nexuses than catchments, where we did end up relying on accumulation to change how data was gathered up to pass to t-route.

src/core/SurfaceLayer.cpp
PhilMiller: Right now, the redundant checks mean that the code is going to need to be changed by anyone who wants to work with the dendritic-network assumption lifted.

I think I made or asked for this change from pre-existing 0 vs >0 to structurally distinguish the actual categories we address of 0, 1, >1.

I think at this point, this code is a 'leaf' change, and not subject to further merge conflict, but I do think it's an improvement over the status quo, that is easier to revise later with the change than without.

PhilMiller: Regarding my last line, the existing code making the dendritic assumption will silently compute wrong results on a non-dendritic network. As far as I'm concerned, that is the worst possible failure mode for a scientific application. Erroring out should definitely be preferable.

src/NGen.cpp
PhilMiller: I think some bits of `NgenSimulation` may use that, regardless. It shouldn't be a big deal, relative to the output arrays, since it's not scaling with the duration of the simulation.

include/core/Layer.hpp
hellkite500: My comment was just to get thoughts down, I'm ok with this as a stopgap given no actual functional changes are currently reliant

hellkite500 (approved):
Approving for merge as *mostly* a refactor, but staging for other work to land.  There are likely parts of this we will revisit as the saga continue.

PR #558, opened 2023-07-03
Compare Python & NumPy versions from configure/build time to runtime

PhilMiller:
Fixes #545

## Additions

- Add `src/core/NGen_Python_Build_Versions.in]` to have CMake generate a corresponding `.cpp` file with the Python/NumPy versions

## Changes

- Get versions of Python and NumPy seen at runtime
- Check runtime versions against configure/build-time versions

## Testing

1. Tests fail if I `pip install numpy==1.24.0` in my virtual environment after building with 1.25.0 installed

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows project standards (link if applicable)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future todos are captured in comments
- [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

PR #968, opened 2026-05-15
Import NGWPC Python InterpreterUtil changes

PhilMiller:
We encountered a bug resulting from an exception safety deficiency in the `InterpreterUtil`, where it would create `NULL` module map entries when loading something failed, and later attempts would then dereference those invalid entries.

## Additions

- Introduce a regression test that is confirmed to actually fail under the original conditions in which the bug was observed

## Changes

- Fix the exception safety bug
- Shift code from the header to a separate source file in its own library target

## Notes

I suggest reviewing each commit separately, since the test and fix are small and substantive, while the last is larger and just a structural refactoring.

The bugfix and code reorganization are as in the NGWPC `development` branch, while the other commits are new to this PR.

The bug reproduction and subsequent fix are only strictly applicable when the code is built as C++14 rather than C++17. The latter specified expression evaluation order such that the existing code became safe as written. As noted in the commit messages, I think it's still preferable to make this case explicit.

## Testing

1. Introduced test fails without the subsequent fix, if the codebase is forcibly built as C++14 rather than C++17, and passes with it
2. Github CI

## Screenshots

Before:
```
[----------] 1 test from InterpreterUtilTest
[ RUN      ] InterpreterUtilTest.FailedImportDoesNotPoisonModuleMap
/Users/phil/Code/noaa/ngen/test/utils/python/InterpreterUtil_Test.cpp:63: Failure
Expected: module = interp->getModule(bogus_name) throws an exception of type py::error_already_set.
  Actual: it throws nothing.
/Users/phil/Code/noaa/ngen/test/utils/python/InterpreterUtil_Test.cpp:64: Failure
  module.ptr()
    Which is: NULL
  module_saved.ptr()
    Which is: 0x1053fb6f0
[  FAILED  ] InterpreterUtilTest.FailedImportDoesNotPoisonModuleMap (0 ms)
[----------] 1 test from InterpreterUtilTest (0 ms total)
```

After:
```
[----------] 1 test from InterpreterUtilTest
[ RUN      ] InterpreterUtilTest.FailedImportDoesNotPoisonModuleMap
WARNING: importTopLevelModule: ngen_definitely_not_a_real_module_xyzzy: ModuleNotFoundError: No module named 'ngen_definitely_not_a_real_module_xyzzy'
Already imported modules: numpy, pathlib, sys, test_bmi_py, 
WARNING: importTopLevelModule: ngen_definitely_not_a_real_module_xyzzy: ModuleNotFoundError: No module named 'ngen_definitely_not_a_real_module_xyzzy'
Already imported modules: numpy, pathlib, sys, test_bmi_py, 
[       OK ] InterpreterUtilTest.FailedImportDoesNotPoisonModuleMap (0 ms)
[----------] 1 test from InterpreterUtilTest (0 ms total)
```

## Checklist

### DIFF
--- a/include/utilities/python/InterpreterUtil.hpp
+++ b/include/utilities/python/InterpreterUtil.hpp
@@ -5,7 +5,6 @@
 
 #if NGEN_WITH_PYTHON
 
-#include <cstdlib>
 #include <map>
 #include <pybind11/embed.h>
 #include <pybind11/stl.h>
@@ -32,7 +31,6 @@ namespace utils {
         class InterpreterUtil {
 
         public:
-            static std::shared_ptr<InterpreterUtil> getInstance() {
                 /**
                  * @brief Singleton instance of embedded python interpreter.
                  * 
@@ -54,20 +52,7 @@ namespace utils {
                  * https://github.com/pybind/pybind11/issues/1598
                  * 
                  */
-                //Functionally, a global instance variable
-                //This can be made a class attribute, but would need to find a place to put a single definition
-                //e.g. make a .cpp file
-                static std::weak_ptr<InterpreterUtil> _instance;
-                std::shared_ptr<InterpreterUtil> instance = _instance.lock();
-                if(!instance){
-                    //instance is null
-                    InterpreterUtil* pt = new InterpreterUtil();
-                    instance = std::shared_ptr<InterpreterUtil>(pt, Deleter{});
-                    //update the weak ref
-                    _instance = instance;
-                }
-                return instance;
-            }
+            static std::shared_ptr<InterpreterUtil> getInstance();
 
             struct Deleter {
                 /**
@@ -82,41 +67,7 @@ namespace utils {
             friend Deleter;
 
         private:
-            InterpreterUtil() {
-                guardPtr = std::make_shared<py::scoped_interpreter>();
-                // Go ahead and do these
-                importTopLevelModule("pathlib");
-                importTopLevelModule("sys");
-                Path = importedTopLevelModules["pathlib"].attr("Path");
-
-                py::list venv_package_dirs = getVenvPackagesDirOptions();
-                // Add any found options
-                for (auto &opt : venv_package_dirs) {
-                    addToPath(std::string(py::str(opt.attr("parent"))));
-                    addToPath(std::string(py::str(opt)));
-                }
-
-                py::object python_version_info = importedTopLevelModules["sys"].attr("version_info");
-                py::str runtime_python_version = importedTopLevelModules["sys"].attr("version");
-                int major = py::int_(python_version_info.attr("major"));
-                int minor = py::int_(python_version_info.attr("minor"));
-                int patch = py::int_(python_version_info.attr("micro"));
-                if (major != python_major
-                    || minor != python_minor
-                    || patch != python_patch) {
-                    throw std::runtime_error("Python version mismatch between configure/build ("
-                                             + std::string(python_version)
-                                             + ") and runtime (" + std::string(runtime_python_version) + ")");
-                }
-
-                importTopLevelModule("numpy");
-                py::str runtime_numpy_version = importedTopLevelModules["numpy"].attr("version").attr("version");
-                if(std::string(runtime_numpy_version) != numpy_version) {
-                    throw std::runtime_error("NumPy version mismatch between configure/build ("
-                                             + std::string(numpy_version)
-                                             + ") and runtime (" + std::string(runtime_numpy_version) + ")");
-                }
-            }
+            InterpreterUtil();
 
             ~InterpreterUtil() = default;
 
@@ -131,35 +82,14 @@ namespace utils {
              *
              * @param directoryPath The desired directory to add to the Python system path.
              */
-            static void addToPyPath(const std::string &directoryPath) {
-                getInstance()->addToPath(directoryPath);
-            }
+            static void addToPyPath(const std::string &directoryPath);
 
             /**
              * Add to the Python system path of this instance.
              *
              * @param directoryPath The desired directory to add to the Python system path.
              */
-            void addToPath(const std::string &directoryPath) {
-                std::tuple<py::list, std::vector<std::string>> sysPathTuple = getSystemPath();
-                py::list sys_path = std::get<0>(sysPathTuple);
-                std::vector<std::string> sys_path_vector = std::get<1>(sysPathTuple);
-
-                py::object requestedDirPath = Path(directoryPath);
-                if (py::bool_(requestedDirPath.attr("is_dir")())) {
-                    if (std::find(sys_path_vector.begin(), sys_path_vector.end(), directoryPath) == sys_path_vector.end()) {
-#ifdef __APPLE__
-                        sys_path.attr("insert")(1, py::str(directoryPath));
-#else
-                        sys_path.attr("insert")(sys_path_vector.size(), py::str(directoryPath));
-#endif
-                    }
-                }
-                else {
-                    std::string dirPath = py::str(requestedDirPath);
-                    throw std::runtime_error("Cannot add non-existing directory '" + dirPath + "' to Python PATH");
-                }
-            }
+            void addToPath(const std::string &directoryPath);
 
             /**
              * Return bound Python top level module handle from the singleton instance, importing the module if needed.
@@ -171,9 +101,7 @@ namespace utils {
              * @param name Name of the desired top level module or package.
              * @return Handle to the desired top level Python module.
              */
-            static py::object getPyModule(const std::string &name) {
-                return getInstance()->getModule(name);
-            }
+            static py::object getPyModule(const std::string &name);
 
             /**
              * Return bound Python module handle from the singleton instance, importing a top-level module if necessary.
@@ -189,9 +117,7 @@ namespace utils {
              *                         and (when applicable) submodules.
              * @return Handle to the desired Python module or type.
              */
-            static py::object getPyModule(const std::vector<std::string> &moduleLevelNames) {
-                return getInstance()->getModule(moduleLevelNames);
-            }
+            static py::object getPyModule(const std::vector<std::string> &moduleLevelNames);
 
             /**
              * Return a bound handle to a top-level Python module, importing the module if necessary.
@@ -203,12 +129,7 @@ namespace utils {
              * @param name Name of the desired top level module or namespace package.
              * @return Handle to the desired top level Python module.
              */
-            py::object getModule(const std::string &name) {
-                if (!isImported(name)) {
-                    importTopLevelModule(name);
-                }
-                return importedTopLevelModules.find(name)->second;
-            }
+            py::object getModule(const std::string &name);
 
             /**
              * Return a bound handle to a Python module or type, importing the top-level module if necessary.
@@ -220,14 +141,7 @@ namespace utils {
              *                         and (when applicable) submodules.
              * @return Handle to the desired Python module or type.
              */
-            py::object getModule(const std::vector<std::string> &moduleLevelNames) {
-                // Start with top-level module name, then descend through attributes as needed
-                py::object module = getModule(moduleLevelNames[0]);
-                for (size_t i = 1; i < moduleLevelNames.size(); ++i) {
-                    module = module.attr(moduleLevelNames[i].c_str());
-                }
-                return module;
-            }
+            py::object getModule(const std::vector<std::string> &moduleLevelNames);
 
         protected:
 
@@ -236,21 +150,7 @@ namespace utils {
              *
              * @return The absolute path of the site packages directory, as a string.
              */
-            py::list getVenvPackagesDirOptions() {
-                // Look for a local virtual environment directory also, if there is one
-                const char* env_var_venv = std::getenv("VIRTUAL_ENV");
-                py::object venv_dir = env_var_venv != nullptr ? Path(env_var_venv): py::none();
-
-                if (!venv_dir.is_none() && py::bool_(venv_dir.attr("is_dir")())) {
-                    // Resolve the full path
-                    venv_dir = venv_dir.attr("resolve")();
-                    // Get options for the packages dir
-                    py::list site_packages_options = py::list(venv_dir.attr("glob")("**/site-packages/"));
-                    return site_packages_options;
-                }
-
-                return py::list();
-            }
+            py::list getVenvPackagesDirOptions();
 
         public:
             /**
@@ -259,29 +159,14 @@ namespace utils {
              * @return A tuple containing the Python list object and a C++ vector of strings, each representing the
              *         current Python system path.
              */
-            inline std::tuple<py::list, std::vector<std::string>> getSystemPath() {
-                py::list sys_path = importedTopLevelModules.find("sys")->second.attr("path");
-                std::vector<std::string> sys_path_vector(sys_path.size());
-                size_t i = 0;
-                for (auto item : sys_path) {
-                    sys_path_vector[i++] = py::str(item);
-                }
-                return std::make_tuple(sys_path, sys_path_vector);
-            }
+            std::tuple<py::list, std::vector<std::string>> getSystemPath();
+
             /**
              * Find any virtual environment site packages directory, starting from options under the current directory.
              *
              * @return The absolute path of the site packages directory, as a string.
              */
-            std::string getDiscoveredVenvPath() {
-                // Look for a local virtual environment directory also, if there is one
-                const char* env_var_venv = std::getenv("VIRTUAL_ENV");
-                if(env_var_venv != nullptr){
-                    //std::string r(env_var_venv);
-                    return std::string(env_var_venv);
-                }
-                return std::string("None");
-            }
+            std::string getDiscoveredVenvPath();
 
         protected:
             /**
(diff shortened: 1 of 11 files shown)


include/utilities/python/InterpreterUtil.hpp
```diff
@@ -335,7 +335,20 @@ namespace utils {
...
+                    importedTopLevelModules[topLevelName] = std::move(module);
+                }
+                catch (std::exception &e) {
+                    std::stringstream ss;
+                    ss << "importTopLevelModule: " << topLevelName << ": " << e.what() << std::endl;
+                    ss << "Already imported modules: " <<
+                    for (const auto& module : importedTopLevelModules) {
+                        ss << module.first << ", ";
+                    }
+                    LOG(ss.str(), LogLevel::WARNING);
```
PhilMiller: Eh, that's not gonna build as-is. I'll make a cleanup commit

PhilMiller:
The `test_unit` CI job doesn't enable Python, so that's not exhibited as red here. Will address somehow.

PhilMiller:
Lol, so I realized in trying to reproduce the error in a unit/regression test to red/green this PR that the move to C++17 somewhat obviates the change. https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0145r3.pdf guarantees that the RHS is fully evaluated, and hence the exception thrown, before the LHS starts to be evaluated and creates the `NULL` default-initialized map entry.

I think it's maybe still worth integrating this regardless, 
* in case someone wants to apply it to a slightly older version of the code, 
* to work down the differences from the NGWPC version of the codebase
* to make the exception safety more explicit, rather than relying on the subtle language behavior

include/utilities/python/InterpreterUtil.hpp
```diff
@@ -32,7 +31,6 @@ namespace utils {
         class InterpreterUtil {
```
christophertubbs, line 31: Maybe out of scope, but I wish there was a bit more of a clear "This is why we're juggling `public` and `private` scope definitions" comment. Made sense after a bit of research, though.

src/utilities/python/InterpreterUtil.cpp
```diff
@@ -0,0 +1,218 @@
...
+static inline const char* safe(const char* s) {
+    return s ? s : "<null>";
+}
+
+namespace utils {
+    namespace ngenPy {
+
+        using namespace pybind11::literals; // to bring in the `_a` literal
+
+        /* static */ std::shared_ptr<InterpreterUtil> InterpreterUtil::getInstance() {
```
christophertubbs, line 25: The commented out static is confusing for _me_. It'd be a little easier for someone like me to read if it were removed.

src/utilities/python/InterpreterUtil.cpp
```diff
@@ -0,0 +1,218 @@
...
+                    InterpreterUtil* pt = new InterpreterUtil();
+                    instance = std::shared_ptr<InterpreterUtil>(pt, Deleter{});
+                    //update the weak ref
+                    _instance = instance;
+                }
+                return instance;
+            }
+
+        InterpreterUtil::InterpreterUtil() {
+                guardPtr = std::make_shared<py::scoped_interpreter>();
```
christophertubbs, line 63: `this->` would make my life easier. Without it, I have to go hunting for more information on it. Not a huge deal, but it would make it easier for the lower skilled C++ dev.

src/utilities/python/InterpreterUtil.cpp
```diff
@@ -0,0 +1,218 @@
...
+                    addToPath(std::string(py::str(opt.attr("parent"))));
+                    addToPath(std::string(py::str(opt)));
+                }
+
+                py::object python_version_info = importedTopLevelModules["sys"].attr("version_info");
+                py::str runtime_python_version = importedTopLevelModules["sys"].attr("version");
+                int major = py::int_(python_version_info.attr("major"));
+                int minor = py::int_(python_version_info.attr("minor"));
+                int patch = py::int_(python_version_info.attr("micro"));
+                if (major != python_major
```
christophertubbs, line 81: patch version is a bit too extreme. Unless there's a super important reason for it, I'd prefer it'd be taken out as it might be a little brittle. An unrelated system update may require a full rebuild of the binary (if I understand this correctly).

src/utilities/python/InterpreterUtil.cpp
```diff
@@ -0,0 +1,218 @@
...
+                    || minor != python_minor
+                    || patch != python_patch) {
+                    throw std::runtime_error("Python version mismatch between configure/build ("
+                                             + std::string(python_version)
+                                             + ") and runtime (" + std::string(runtime_python_version) + ")");
+                }
+
+                importTopLevelModule("numpy");
+                py::str runtime_numpy_version = importedTopLevelModules["numpy"].attr("version").attr("version");
+                if(std::string(runtime_numpy_version) != numpy_version) {
```
christophertubbs, line 91: Same concern of strict version comparison. It's obviously a little trickier to be less strict here, but it'd help make the system more resilient if it just went "Major minor? Fine". That is, unless it's doing a further check to ensure that some internal processing/runtime linkage assumption is correct.

src/utilities/python/InterpreterUtil.cpp
```diff
@@ -0,0 +1,218 @@
...
+                    std::string version_str = runtime_numpy_version.cast<std::string>();
+                    const char* version_cstr = version_str.c_str();
+                    throw std::runtime_error("NumPy version mismatch between configure/build ("
```
christophertubbs, line 92: Any reason `version_cstr` isn't used within the message?

src/utilities/python/InterpreterUtil.cpp
```diff
@@ -0,0 +1,218 @@
...
+#endif
+                    }
+                }
+                else {
+                    std::string dirPath = py::str(requestedDirPath);
+                    throw std::runtime_error("Cannot add non-existing directory '" + dirPath + "' to Python PATH");
+                }
+            }
+
+            /* static */ py::object InterpreterUtil::getPyModule(const std::string &name) {
```
christophertubbs, line 124: What are the chances that this could lead to a `py::object` leakage (i.e. the `py::object` living longer than the instance)?

christophertubbs (commented):
I'd _prefer_ the strict version checking be relaxed a tad - requiring `3.11` or something? Great. Requiring `3.11.6`, specifically? A tad dicey. 

A potential risk factor I've found was a lack of gil locking (unless I missed it somewhere). That's a ticking multithreaded time bomb. 

I understand that a good bit of my concerns are a tad out of scope - if you're just trying to fix one thing and move implementation code out of the header, you don't necessarily want to rock the boat _too_ much. **But**... there are technically new lines of code so I reserve the right to complain.

I'm setting this review as just a comment - someone with a higher skill level should review. A lot of this is just a tad too clever for me and I think most of my complaints are preexisting issues.

src/utilities/python/InterpreterUtil.cpp
robertbartel: Phil can perhaps add more detail, but I'm fairly sure we do actually want this to break if ngen was not compiled using exactly the same version of Python that is then used at runtime.  I vaguely recall seeing issues related to numpy otherwise.

src/utilities/python/InterpreterUtil.cpp
robertbartel: See comment above:  I think there is a need for this, because the implementation is (independent of this code) not always able to handled version mismatches.

src/utilities/python/InterpreterUtil.cpp
```diff
@@ -0,0 +1,217 @@
...
+#include <pybind11/numpy.h>
+
+#include <cstdlib>
+#include <map>
+#include <vector>
+#include <tuple>
+
+namespace py = pybind11;
+
+static inline const char* safe(const char* s) {
```
robertbartel: Is this really needed?  I don't see anywhere where it is used, nor was it present when everything was in the header.

src/utilities/python/InterpreterUtil.cpp
robertbartel: Here, I agree:  are these two local variables actually needed?

robertbartel (changes requested):
I see a couple of very small things that, assuming I haven't missed something, strictly speaking should be fixed.  But otherwise, this looks good to me.  Tweak or explain those, and then I'll come back and approve.

src/utilities/python/InterpreterUtil.cpp
PhilMiller: Cf https://github.com/NOAA-OWP/ngen/pull/558

src/utilities/python/InterpreterUtil.cpp
PhilMiller: No, that's a relic of other NGWPC stuff that's irrelevant. I'll drop it.

src/utilities/python/InterpreterUtil.cpp
PhilMiller: Another relic of other NGWPC stuff. I'll match the code to what was removed from the header

PhilMiller: Removed.

PhilMiller:
Chris, the rest of your comments that I haven't made changes for concern pre-existing code, that's just being moved from a header file to a source file. So, I think they're out of scope for addressing in this PR. To the extent they ought to be addressed, that is work that can be planned down the line. It's definitely not on our budget or tasking right now.

I'll still reply to the comments.

include/utilities/python/InterpreterUtil.hpp
PhilMiller: Yeah, I dunno. There feels like some sort of intuitive order in these declarations. I'm just not messing with it here. Perhaps the juggling is more tangible because there are no longer function bodies interspersed to hide that it's going on.

src/utilities/python/InterpreterUtil.cpp
PhilMiller: That's there indicating how this member function was declared in the header, making it clear here that it is not an instance member function. C++ is kinda wonky on which bits of text in a function declaration are the "function signature", to name its definition, and which are not.

src/utilities/python/InterpreterUtil.cpp
PhilMiller: I've worked in a whole bunch of different C++ code bases, and `this->` has not been idiomatic in any of them. It's kinda taken as a signature of someone who learned to code in Java, and never really acclimated to the switch to C++.

More common for marking member variables as distinct from argument and locals were prefix `m_foo` or just `_foo`, or suffix `foo_` (my preference).

In the `m_` case, globals were `g_`, function arguments `a_`, static variables `s_`, and local variables were unmarked. Even that was a pretty eccentric style, that I haven't seen in any code not sharing that library's heritage (LBL BoxLib/Chombo/AMREX).

src/utilities/python/InterpreterUtil.cpp
PhilMiller: Probably pretty high if misused. Thankfully, it's not much used throughout the codebase.

If we were writing in Rust, there would be a lifetime parameter on the return value matching that of the `InterpreterUtil` itself. We don't have lifetime parameters in C++. Maybe in C++ 2035 or something.

christophertubbs: > Probably pretty high if misused.

So it'll bite us immediately 😄 

Is there anything we can do to highlight "Hey, check this thing out here if you're running into this type of problem"? An inline comment would only help if you knew exactly where to look for it. If we had logging 100% locked down, we could have a debug statement there, but that's not really the case.

This is all just a spitball - I can't think of a solid solution since I don't think we have a diagnostic guide or anything for it.

Feel free to mark as 'resolved' if it's sort of a 'not much we can reasonably do' situation.

src/utilities/python/InterpreterUtil.cpp
christophertubbs: Bobby/Phil - mark this as resolved if my concern isn't worth addressing due to high version incompatibility risk.

src/utilities/python/InterpreterUtil.cpp
christophertubbs: > It's kinda taken as a signature of someone who learned to code in Java, and never really acclimated to the switch to C++.

Rude

Anyways, I generally go with `foo_` as well. I should have put "my" and "lower skilled C++ dev" in bold. We're going to have a lot of primary Python and R devs passing through this, so I always err on the side of "as obvious as humanly possible, even if it's a tad excessive". As it stands, `guardPtr` is a tad ambiguous (to me at least), hence the preference for `this->`.

A transition for this styling is out of scope, but I wanted the preference known at least.

Resolve when you see this. I'd do it, but I want to make sure I'm not shouting into the void.

src/utilities/python/InterpreterUtil.cpp
christophertubbs: That's the part I understand - the confusion was based on "why was this in the imp code in the first place - that wasn't necessary" not "isn't this necessary?!". 

Maybe something like:

```c++
/* static function */
std::shared_ptr<InterpreterUtil> InterpreterUtil::getInstance() {
```

would have reduced the readability stumble. That way it's more clear that there wasn't a signature stumble and that it's more of an annotation.

include/utilities/python/InterpreterUtil.hpp
christophertubbs: I _believe_ it's in that order because things need to be deconstructed in a highly specific order.

My primary concern is that someone will go through and try to "fix" it.

christophertubbs:
> Chris, the rest of your comments that I haven't made changes for concern pre-existing code, that's just being moved from a header file to a source file. So, I think they're out of scope for addressing in this PR. To the extent they ought to be addressed, that is work that can be planned down the line. It's definitely not on our budget or tasking right now.

Which is why I noted that you wouldn't want to make major changes when it comes to just moving implementation to the correct location.

The `gil` locking probably needs to be handled sooner rather than later, though, and I think it might result in the need for different access patterns since you probably don't want to play with items like `py::object` outside the scope of the lock.

Do you guys want me to go ahead and add a ticket for it or do you guys want to try and figure out the best way to address it?

src/utilities/python/InterpreterUtil.cpp
robertbartel: Heh, so, I spent a lot of time in Java, but I learned C and C++ first.  And C++ did use `this` first.  So perhaps it’s also a sign of someone being just old.

My initially reaction to the general use of trailing `_` for this is slightly uncomfortable hesitation.   FWIW, I’ve never especially liked the leading `_` in Python to designate protected/private (ish) variables.   I think for me that comes from it being a convention and not a rule.  Part of me also feels like the trailing `_` is something I might not notice (or notice I left out) in certain circumstances, while `this` is not.   Though maybe that’s just me; perhaps since I haven’t typically used that convention, I simply haven’t trained myself to instinctively look at that.

Regardless, probably another sign that the project needs a formalized, documented, enforced code style.

PhilMiller: On this part of the codebase, in particular, those primary R and Python devs are mostly going to have to steer clear. There's enough tricky C++ and Python interfacing stuff going on here that we can't really insulate it beyond saying "don't mess with the guts of `InterpreterUtil` unless you know what you're doing".

Thanks for the go-ahead clearance on this.

PhilMiller:
@christophertubbs If you anticipate efforts toward adding threading, yeah, open that ticket now. At the moment, we're probably not putting any effort into thinking deeper about it than you would.

robertbartel (approved):

src/utilities/python/InterpreterUtil.cpp
PhilMiller: Ok, so I just did a quick skim of the code for where `getPyModule` and `getModule` are actually used, and it's quite limited. Essentially just `Bmi_Py_Adapter`. I don't see any reason that usage would really expand within the ngen codebase. I think the risk of new mis-use is at least reasonably contained.

src/utilities/python/InterpreterUtil.cpp
PhilMiller: From the perspective of "let the compiler help you write correct code", some sort of marking of member variable names (`m_`, leading or trailing `_`) is preferable to `this->`, because the marked name has to be used for the code to compile, while qualification with `this->` is a syntactically optional convention.

christophertubbs:
Assume people were trying to thread with it _yesterday_. Thread safety absolutely needs to be locked in, though some of the stuff regarding `py::object` may need to have altered access patterns in order to be truly safe. We've already encountered code that are segfault machines and the more we put off fixes the deeper in the hole we get.

include/utilities/python/InterpreterUtil.hpp
PhilMiller: Huh, it can't be a construction/destruction order thing, because all of the member variables are grouped together in a single `private` section at the bottom of the class definition.

I think the various member functions and the nested `struct Deleter` are ordered like they are basically for exposition, and not any technical reason.

PhilMiller: Oh, yeah, and the member variables in that final `private` section do have a comment explicitly noting that their order matters for destruction.

PhilMiller:
I really wish the various user and developer groups of ngen were a bit more consolidated, so that stuff like the threading issues were universally visible. Maybe the CIROH org repositories are the real center, and we should stop pretending that the OWP repos are the single point of truth.

PR #969, opened 2026-05-20
fix(routing): syntax error at isNexus() call in nexus indexing

hellkite500:
Commit 03c86a52e from PR #967 has a syntax error that only shows up when building with routing.  This is the fix for that syntax error.


PhilMiller:
Could you also change `test_troute_integration` to build the `ngen` target, so that CI would have caught this? It currently only builds the `test_routing_pybind` target

PhilMiller (dismissed):

PhilMiller (approved):

PR #974, opened 2026-05-21
Extract a NgenSimulation class encapsulating the time-dependent logic from main()

PhilMiller:
This is one step along a path toward both modularizing the NGen codebase. This serves multiple development goals, including state saving and an eventual ability to load and drive NGen as a library, from a server or a variety of different UIs.

## Additions

- `class NgenSimulation`

## Changes

- Move a bunch of code from `main()` into member functions of `NgenSimulation`
- Move ownership of master `Simulation_Time` instance out of `Formulation_Manager`

## Testing

1. Local ctest
2. Github CI
3. A VPU for a month, with confirmation that results are identical

## Notes

- There's routing logic for driving t-route via the NGWPC-developed BMI implementation. It's conditionalized under `NGEN_WITH_ROUTING_TROUTE_BMI`. It sits alongside the existing implementation that just executes t-route as if it were a standalone script. The data retention in the BMI form is potentially problematic for large-scale, long-duration simulations. This could be broken into some sort of chunked or incremental execution, but that's not been developed yet on the t-route side, and implementing it here is out of scope.

## Checklist

### DIFF
--- a/src/NGen.cpp
+++ b/src/NGen.cpp
@@ -20,6 +20,8 @@
 #include <boost/algorithm/string.hpp>
 #include <boost/range/algorithm/sort.hpp>
 
+#include <NgenSimulation.hpp>
+
 #ifdef WRITE_PID_FILE_FOR_GDB_SERVER
 #include <unistd.h>
 #endif // WRITE_PID_FILE_FOR_GDB_SERVER
@@ -602,9 +604,6 @@ int main(int argc, char* argv[]) {
     auto time_done_init                             = std::chrono::steady_clock::now();
     std::chrono::duration<double> time_elapsed_init = time_done_init - time_start;
 
-    // Now loop some time, iterate catchments, do stuff for total number of output times
-    auto num_times = sim_time->get_total_output_times();
-
     // T-ROUTE data storage
     std::vector<double> catchment_outflows;
     std::unordered_map<std::string, int> catchment_indexes;
@@ -620,67 +619,15 @@ int main(int argc, char* argv[]) {
     nexus_downstream_flows.resize(nexus_indexes.size() * num_times, 0.0);
 #endif // NGEN_WITH_ROUTING
 
-    for (int count = 0; count < num_times; count++) {
-        // The Inner loop will advance all layers unless doing so will break one of two constraints
-        // 1) A layer may not proceed ahead of the master simulation object's current time
-        // 2) A layer may not proceed ahead of any layer that is computed before it
-        // The do while loop ensures that all layers are tested at least once while allowing
-        // layers with small time steps to be updated more than once
-        // If a layer with a large time step is after a layer with a small time step the
-        // layer with the large time step will wait for multiple timesteps from the preceeding
-        // layer.
+    auto simulation = std::make_unique<NgenSimulation>(manager,
+                                                       layers,
+                                                       catchment_indexes,
+                                                       catchment_outflows,
+                                                       nexus_indexes,
+                                                       nexus_downstream_flows);
 
-        // this is the time that layers are trying to reach (or get as close as possible)
-        auto next_time = sim_time->next_timestep_epoch_time();
-
-        // this is the time that the layer above the current layer is at
-        auto prev_layer_time = next_time;
-
-        // this is the time that the least advanced layer is at
-        auto layer_min_next_time = next_time;
-        do {
-            for (auto& layer : layers) {
-                auto layer_next_time = layer->next_timestep_epoch_time();
-
-                // only advance if you would not pass the master next time and the previous layer
-                // next time
-                if (layer_next_time <= next_time && layer_next_time <= prev_layer_time) {
-                    if (count % 100 == 0) {
-		      std::cout << "Updating layer: " << layer->get_name() << "\n";
-                    }
-#if NGEN_WITH_ROUTING && false
-                    boost::span<double> catchment_span(catchment_outflows.data() + (count * catchment_indexes.size()),
-                                                       catchment_indexes.size());
-                    boost::span<double> nexus_span(nexus_downstream_flows.data() + (count * nexus_indexes.size()),
-                                                   nexus_indexes.size());
-#else
-                    boost::span<double> catchment_span;
-                    boost::span<double> nexus_span;
-#endif
-                    layer->update_models(
-                        catchment_span,
-                        catchment_indexes,
-                        nexus_span,
-                        nexus_indexes,
-                        count
-                    ); // assume update_models() calls time->advance_timestep()
-                    prev_layer_time = layer_next_time;
-                } else {
-                    layer_min_next_time = prev_layer_time = layer->current_timestep_epoch_time();
-                }
-
-                if (layer_min_next_time > layer_next_time) {
-                    layer_min_next_time = layer_next_time;
-                }
-            } // done layers
-        } while (layer_min_next_time < next_time
-        ); // rerun the loop until the last layer would pass the master next time
-
-        if (count + 1 < num_times) {
-            sim_time->advance_timestep();
-        }
-
-    } // done time
+    // Currently breaks routing data transfer; will move routing down into this
+    simulation->run_catchments();
 
     // Close nexus output file(s)
     nexus_outputs_mgr->close();
--- a/src/NGen.cpp
+++ b/src/NGen.cpp
@@ -623,7 +623,8 @@ int main(int argc, char* argv[]) {
     auto simulation = std::make_unique<NgenSimulation>(manager,
                                                        layers,
                                                        catchment_indexes,
-                                                       nexus_indexes);
+                                                       nexus_indexes,
+                                                       mpi_rank);
 
     simulation->run_catchments();
 
--- a/src/NGen.cpp
+++ b/src/NGen.cpp
@@ -646,22 +646,9 @@ int main(int argc, char* argv[]) {
     MPI_Barrier(MPI_COMM_WORLD);
 #endif
 
-#if NGEN_WITH_ROUTING
-    if (mpi_rank == 0)
-    { // Run t-route from single process
-        if(manager->get_using_routing()) {
-          //Note: Currently, delta_time is set in the t-route yaml configuration file, and the
-          //number_of_timesteps is determined from the total number of nexus outputs in t-route.
-          //It is recommended to still pass these values to the routing_py_adapter object in
-          //case a future implmentation needs these two values from the ngen framework.
-          int number_of_timesteps = sim_time->get_total_output_times();
-
-          int delta_time = sim_time->get_output_interval_seconds();
-          
-          router->route(number_of_timesteps, delta_time); 
-        }
+    if (manager->get_using_routing()) {
+        simulation->run_routing();
     }
-#endif
 
     auto time_done_routing = std::chrono::steady_clock::now();
     std::chrono::duration<double> time_elapsed_routing = time_done_routing - time_done_simulation;
--- a/src/NGen.cpp
+++ b/src/NGen.cpp
@@ -601,14 +601,6 @@ int main(int argc, char* argv[]) {
         }
     }
 
-    auto time_done_init                             = std::chrono::steady_clock::now();
-    std::chrono::duration<double> time_elapsed_init = time_done_init - time_start;
-
-    LOG("[TIMING]: Init: " + std::to_string(time_elapsed_init.count()), LogLevel::INFO);
-
-    // Now loop some time, iterate catchments, do stuff for total number of output times
-    auto num_times = manager->Simulation_Time_Object->get_total_output_times();
-
     // T-ROUTE data storage
     std::unordered_map<std::string, int> catchment_indexes;
 #if NGEN_WITH_ROUTING && false
@@ -625,6 +617,9 @@ int main(int argc, char* argv[]) {
                                                        std::move(catchment_indexes),
                                                        std::move(nexus_indexes),
                                                        mpi_rank);
+
+    auto time_done_init                             = std::chrono::steady_clock::now();
+    std::chrono::duration<double> time_elapsed_init = time_done_init - time_start;
 
     simulation->run_catchments();
 
--- a/src/NGen.cpp
+++ b/src/NGen.cpp
@@ -625,7 +625,6 @@ int main(int argc, char* argv[]) {
                                                        catchment_indexes,
                                                        nexus_indexes);
 
-    // Currently breaks routing data transfer; will move routing down into this
     simulation->run_catchments();
 
     // Close nexus output file(s)
--- a/src/NGen.cpp
+++ b/src/NGen.cpp
@@ -647,7 +647,7 @@ int main(int argc, char* argv[]) {
 #endif
 
     if (manager->get_using_routing()) {
-        simulation->run_routing();
+        simulation->run_routing(features);
     }
 
     auto time_done_routing = std::chrono::steady_clock::now();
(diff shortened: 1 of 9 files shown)


src/core/NgenSimulation.cpp
```diff
@@ -0,0 +1,274 @@
...
+        if (mpi_rank == 0) {
+            // update root's local data for running t-route below
+            routing_nexus_indexes = &all_nexus_indexes;
+            routing_nexus_downflows = &all_nexus_downflows;
+        }
+    }
+#endif // NGEN_WITH_MPI
+
+    if (mpi_rank_ == 0) { // Run t-route from single process
+        LOG(LogLevel::INFO, "Running T-Route on nexus outflows.");
```
PhilMiller: Need to clear out the LOG calls that are NGWPC-style

PhilMiller:
Gonna back off the routing changes in here, to narrow the focus

src/NGen.cpp
```diff
@@ -436,8 +438,6 @@ int main(int argc, char* argv[]) {
     { // Run t-route from single process
     if(manager->get_using_routing()) {
       std::cout<<"Using Routing"<<std::endl;
-      std::string t_route_config_file_with_path = manager->get_t_route_config_file_with_path();
-      router = std::make_unique<routing_py_adapter::Routing_Py_Adapter>(t_route_config_file_with_path);
```
PhilMiller, line 440: Maybe we prefer to keep this here, and pass `router` into `NgenSimulation`, to maintain the division of construction vs execution logic.

PhilMiller: The way this t-route integration works, I'm inclined to treat the construction as just part of executing it, since there's no intermediate interaction.

src/core/NgenSimulation.cpp
```diff
@@ -0,0 +1,284 @@
...
+            if (layer_next_time <= next_time && layer_next_time <= prev_layer_time) {
+                if (simulation_step_ % 100 == 0) {
+                    std::cout << "Updating layer: '" + layer->get_name() + "' at output step " + std::to_string(simulation_step_) << std::endl;
+                }
+
+#if NGEN_WITH_ROUTING && false
+                boost::span<double> catchment_span(catchment_outflows_.data() + (simulation_step_ * catchment_indexes_.size()),
+                                                   catchment_indexes_.size());
+                boost::span<double> nexus_span(nexus_downstream_flows_.data() + (simulation_step_ * nexus_indexes_.size()),
+                                               nexus_indexes_.size());
```
donaldwj, line 95: Can this code ever execute with the #if NGEN_WITH_ROUTING && false. If this is incomplete it should be removed from the PR. If it waiting on other work it needs to be commented.

src/core/NgenSimulation.cpp
```diff
@@ -0,0 +1,284 @@
...
+            }
+        } // done layers
+    } while (layer_min_next_time < next_time); // rerun the loop until the last layer would pass the master next time
+
+}
+
+int NgenSimulation::get_nexus_index(std::string const& nexus_id) const
+{
+    auto iter = nexus_indexes_.find(nexus_id);
+    return (iter != nexus_indexes_.end()) ? iter->second : -1;
```
donaldwj, line 123: We should probably just make a hashmap with nexus_id as the key.

src/core/NgenSimulation.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+        if (mpi_rank == 0) {
+            // update root's local data for running t-route below
+            routing_nexus_indexes = &all_nexus_indexes;
+            routing_nexus_downflows = &all_nexus_downflows;
+        }
+    }
+#endif // NGEN_WITH_MPI
+
+    if (mpi_rank_ == 0) { // Run t-route from single process
```
donaldwj, line 223: If the bmi t-route is being run by a single rank I don.t think we gain anything. At least run a sub process t-route can use the local node, We may need t-route designed to run with mpi and to run one time step at a time without needing to retain all data for this to be usefull.

src/NGen.cpp
donaldwj: I dont see a problem with keeping it here if that makes more sense to you.

donaldwj (commented):
There are a few things I commented on. That we should discuss and the macos bmi test failed. After those issues are worked out I see no problem approving this.

src/core/NgenSimulation.cpp
PhilMiller: This is directly lifting and shifting the same logic from `main()` in that regard.

donaldwj: Yes that makes sense when we are starting a script and waiting for the job to finish. But I would expect bmi-troute to be started on all ranks and sequentially feed data just like catchments and nexi are.

donaldwj (dismissed):
Looks good

src/core/NgenSimulation.cpp
PhilMiller: You probably already concluded similarly, but for subsequent readers: t-route currently isn't developed to the point of being able to do that parallelization, and that's out of the scope of this PR. This PR does partially encapsulate the breadth of the ngen codebase that would be modified to accomplish that, though.

src/core/Layer.cpp
```diff
@@ -45,7 +45,7 @@ void ngen::Layer::update_models(boost::span<double> catchment_outflows,
                 +" at feature id "+id;
             throw std::runtime_error(msg);
         }
-#if NGEN_WITH_ROUTING && false
+#if NGEN_WITH_ROUTING && NGEN_WITH_ROUTING_TROUTE_BMI
         int results_index = catchment_indexes[id];
 	// XXX: This is currently accumulating in meters of depth, which may not be desirable
```
PhilMiller: Whitespace error here

src/NGen.cpp
```diff
@@ -700,22 +640,16 @@ int main(int argc, char* argv[]) {
...
-          //number_of_timesteps is determined from the total number of nexus outputs in t-route.
-          //It is recommended to still pass these values to the routing_py_adapter object in
-          //case a future implmentation needs these two values from the ngen framework.
-          int number_of_timesteps = sim_time->get_total_output_times();
-
-          int delta_time = sim_time->get_output_interval_seconds();
-          
-          router->route(number_of_timesteps, delta_time); 
-        }
+    manager->finalize();
```
PhilMiller: This `finalize()` call was moved up to release memory from the catchment models in advance of calling t-route, to reduce peak footprint and possibly avoid memory exhaustion conditions. It looks like a merge conflict resolution meant I didn't remove the call below.

Should I keep that optimization in this PR, or just remove it here and address resource releasing pragmatics separately?

src/core/NgenSimulation.cpp
```diff
@@ -0,0 +1,286 @@
...
+}
+
+NgenSimulation::~NgenSimulation() = default;
+
+void NgenSimulation::run_catchments()
+{
+    // Now loop some time, iterate catchments, do stuff for total number of output times
+    auto num_times = get_num_output_times();
+
+    for (; simulation_step_ < num_times; simulation_step_++) {
```
robertbartel, line 46: At first look I find this somewhat confusing:  why is there `simulation_step_`?  Isn't this redundant state for something modeled (even if only implicitly) by `sim_time_`?

src/NGen.cpp
robertbartel, line 666: I know you already know this `finalize()` call is still here, but flagging it to help it be remembered.

src/core/NgenSimulation.cpp
```diff
@@ -0,0 +1,286 @@
...
+        // Make room for this output step's results
+        catchment_outflows_.resize(catchment_outflows_.size() + catchment_indexes_.size(), 0.0);
```
robertbartel, line 49: Also, won't these collections get really, untenably large if the spatial and time domains are big?

src/core/Layer.cpp
```diff
@@ -45,9 +45,9 @@ void ngen::Layer::update_models(boost::span<double> catchment_outflows,
                 +" at feature id "+id;
             throw std::runtime_error(msg);
         }
-#if NGEN_WITH_ROUTING && false
+#if NGEN_WITH_ROUTING && NGEN_WITH_ROUTING_TROUTE_BMI
         int results_index = catchment_indexes[id];
-	// XXX: This is currently accumulating in meters of depth, which may not be desirable
+        // XXX: This is currently accumulating in meters of depth, which may not be desirable
```
robertbartel, line 50: If we are going to enable (albeit conditionally) this code, and given other concerns we've had elsewhere with units aligning, we should make sure this the right units accumulate.  Alternatively, if the proper conversion is being done elsewhere, then this comment should be updated to reflect that.

src/NGen.cpp
```diff
@@ -599,88 +599,28 @@ int main(int argc, char* argv[]) {
...
-    std::chrono::duration<double> time_elapsed_init = time_done_init - time_start;
-
-    // Now loop some time, iterate catchments, do stuff for total number of output times
-    auto num_times = sim_time->get_total_output_times();
-
     // T-ROUTE data storage
-    std::vector<double> catchment_outflows;
     std::unordered_map<std::string, int> catchment_indexes;
-    std::vector<double> nexus_downstream_flows;
 #if NGEN_WITH_ROUTING && false
```
robertbartel: Given several other lines similar are being changed to `#if NGEN_WITH_ROUTING && NGEN_WITH_ROUTING_TROUTE_BMI`, should this have been changed as well?  And if not, does this code just need to be removed?  

And if neither, could we add some kind of explanation in comments on why it's left here but not reachable?

src/core/NgenSimulation.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+        advance_models_one_output_step();
+
+        if (simulation_step_ + 1 < num_times) {
+            sim_time_->advance_timestep();
+        }
+    }
+}
+
+void NgenSimulation::advance_models_one_output_step()
```
robertbartel, line 61: Related to the comment about keeping current time state in two places, I'm uneasy about how this works.  Perhaps that the problem is actually rooted on the layer side, and issues here are really a side effect, but my concern is that this could allow the time to when individual models/layers have advanced to become misaligned, in a non-recoverable way.  

As an example - and in tandem with how the end of the loop in `run_catchments()` above works, what about a scenario where: 
* the "global" simulation time step is 3600
* the layer 0 time step is 900 (or 500)
* the layer 1 time step is 3600

I might just need more coffee this morning, but at a glance those look like they'd be problematic.  I'll go through deliberately to make sure I'm completely following all the nuance, stepping through the logic properly, and not missing restrictions imposed elsewhere that prevent any problematic scenarios here.  But wanted to go ahead and add the comment.

robertbartel (commented):
I've still got a little more reviewing I need to do, and I haven't yet done test runs, but I wanted to go ahead and make sure a few comments were added.

PhilMiller:
Without getting into the details right here, I'll ask for a comparison against the existing logic moved from NGen.cpp main(). The intention was that there not be a change in behavior.

robertbartel:
> Without getting into the details right here, I'll ask for a comparison against the existing logic moved from NGen.cpp main(). The intention was that there not be a change in behavior.

My response to this varies a bit from item to item, depending on which comments you're applying it to (it would seem not all, and also not clear on some).   But fundamentally, I think the act of teasing apart and encapsulating the functionality - as well as side effects that emerge from doing so - weaken the applicability of a strategy to absolutely not change anything about the way the current functionality operates.  I get the idea, and tend to largely agree in general to minimize the scope, but I don't think things can remain that simple here.   

Take my `simulation_step_` redundant state comment.   Duplicating state may have been an existing problem, but the changes for the surrounding/related code expanded the complexity from a local variable to a member variable of another class.  I think there's at least a reasonable argument that this does not qualify as maintaining the same behavior, and that controlling the loop using a more robust state mechanism (as long as it mostly already existed) would not be changing the behavior to a significantly larger degree.

I strongly suspect your note above does apply to my `advance_models_one_output_step()` comment, and I'm much closer to agreeing here.  There is (or would be, hypothetically) something problematic, though, about creating a function if it doesn't work in a manner consistent with what its name suggests (again, I haven't confirmed this, but it's one of my concerns).   To avoid inflating scope here, the right thing may be requiring that be fixed separately, before it's moved/extracted to an encapsulating class.  But I'm not sure whether that'd actually be easier.

christophertubbs:
A 1:1 translation of functionality and behavior from one location utilizing a functional and global scope to a separate location using class state isn't quite possible, but it may be _close_. I agree with Bobby that this movement will serve to exacerbate previously existing flaws. He's especially correct when it comes to issues with vector element accumulation and non-uniform time steps. Both of those seem like they may require a greater effort to address, however.

In order to keep changes small, I suggest creating the new, unwired artifacts in a single delivery/PR and another delivery/PR for wiring up main to this new structure, all while spamming warnings, messages, and tickets meant to make it incredibly obvious that the rough edges on that translation, along with the glaring pre-exiting issues, get smoothed out asap. If necessary, a separate executable target with a separate main file and function may be used until the new approach outclasses the old.

src/core/NgenSimulation.cpp
PhilMiller: We talked about this, and tentatively agreed that it would be appropriate to move the step progress counter into `Simulation_Time` along with the actual modeled time value, but that it would unduly broaden the scope of this PR to do it here.

src/NGen.cpp
PhilMiller: Decided to back this out, to narrow this PR scope, and it being low-impact for current production-targeted efforts that don't run routing from NGen, and the risk of screwing up other bits of finalization.

src/NGen.cpp
PhilMiller: Will keep this one

src/core/NgenSimulation.cpp
```diff
@@ -0,0 +1,286 @@
...
+    : simulation_step_(0)
+    , sim_time_(std::make_shared<Simulation_Time>(sim_time))
+    , layers_(std::move(layers))
+    , catchment_indexes_(std::move(catchment_indexes))
+    , nexus_indexes_(std::move(nexus_indexes))
+    , mpi_rank_(mpi_rank)
+    , mpi_num_procs_(mpi_num_procs)
+{
+    catchment_outflows_.reserve(catchment_indexes_.size() * get_num_output_times());
+    nexus_downstream_flows_.reserve(nexus_indexes_.size() * get_num_output_times());
```
PhilMiller, line 35: Conditionalize these and the `resize` below

src/core/NgenSimulation.cpp
robertbartel: After some discussion with Phil, while I am going to investigate this separately, it doesn't need to hold up this PR.  Multi-layered configurations get basically no use right now, so even if there is a problem here, it has no practical impact at the moment.  And the overall direction this PR is taking is good.

src/core/Layer.cpp
robertbartel: Another thing after talking with Phil some yesterday ... while we are technically making this executable, in practice, it won't be for a while (BMI t-route integration won't be fully ready for a while).  I have opened #976 specifically for tracking this, so I'm comfortable allowing the PR to continue with this as it currently is.

src/core/NgenSimulation.cpp
robertbartel: I've opened #977 now so we can track that separately.

robertbartel (approved):
I think this is a good direction to take things in.  There are still some issues carried over from old code, but their impacts will be minor enough in the foreseeable future that I'm comfortable deferring dealing with them.  I've opened a few issues for tracking some of that.

I've also done a simple test for 1 VPU, for 1 month, using this branch.   Execution is successful and results are identical to previous test runs of the same config.

PhilMiller:
Confirmed with Chris that I should go ahead and merge, and he has no blocking objection.

PR #978, opened 2026-06-04
MPI interfacing and usage cleanup

PhilMiller:
In pursuit of the ngen codebase being more (re-)usable in context other than the current `main()` driver in `NGen.cpp`, begin to parameterize MPI functionality by the MPI communicator to be used.

## Changes

- All of `parallel_utilities` now take `MPI_Comm` arguments
- `main()` and `NgenSimulation` pass `MPI_COMM_WORLD` as argument to those functions for now
- CI now builds the entire codebase (CMake target `all`) with `NGEN_WITH_MPI=ON`
- CMake specifies usage of the MPI C API from C++ code, and an MPI 3 dependence

## Removals

- A debug print of formulation configuration loading that hard-coded `MPI_COMM_WORLD` and `std::cout`

## Testing

1. CI, newly modified to actually build all of the code with MPI enabled
2. VPU-scale for a month produces equal results

## Checklist


include/realizations/config/formulation.hpp
```diff
@@ -58,16 +58,7 @@ namespace realization{
...
-        // If running MPI job, output with only one processor
-        #if NGEN_WITH_MPI
-        int mpi_rank;
-        MPI_Comm_rank(MPI_COMM_WORLD, &mpi_rank);
-        if (mpi_rank == 0)
-        #endif
-        {
-            geojson::JSONProperty::print_property(parameters.at("modules"));
         }
-      }
```
PhilMiller, line 70: The diff of this brace vs the other one looks funky like this because the existing indentation was wrong, and gets corrected here.

include/utilities/parallel_utils.h
```diff
@@ -20,6 +20,7 @@
 #define NGEN_MPI_PROTOCOL_TAG 101
 #endif
 
+#include <mpi.h>
```
PhilMiller, line 23: Note that there's an existing `#if NGEN_WITH_MPI` above.

This does introduce a header dependency, but `mpi.h` is typically quite light.

src/utilities/parallel_utils.cpp
```diff
@@ -11,20 +11,24 @@
...
             codeBuffer = status ? MPI_HF_SUB_CODE_GOOD : MPI_HF_SUB_CODE_BAD;
-            MPI_Send(&codeBuffer, 1, MPI_UNSIGNED_SHORT, 0, 0, MPI_COMM_WORLD);
+            MPI_Send(&codeBuffer, 1, MPI_UNSIGNED_SHORT, 0, 0, comm);
         }
         // In rank 0, the first step is to receive and process codes from the other ranks into unified global status
         else {
             for (int i = 1; i < mpi_num_procs; ++i) {
-                MPI_Recv(&codeBuffer, 1, MPI_UNSIGNED_SHORT, i, 0, MPI_COMM_WORLD, MPI_STATUS_IGNORE);
+                MPI_Recv(&codeBuffer, 1, MPI_UNSIGNED_SHORT, i, 0, comm, MPI_STATUS_IGNORE);
                 // If any is ever "not good", overwrite status to be "false"
```
PhilMiller, line 32: This is a badly non-scalable pattern, but NGen is not being run at scales of process counts for that to really bite us.

Fixing it would be out of scope for this PR's intentions.

PhilMiller:
I intend to post follow-on PRs parameterizing `NgenSimulation` and the NetCDF nexus output code over the MPI communicator. I'm treating that as out of scope for this PR, whose primary goal is to just get the utilities off of hard-coding `MPI_COMM_WORLD`

include/realizations/config/formulation.hpp
```diff
@@ -58,16 +58,7 @@ namespace realization{
                 //Create the nested formulations in order of definition
                 nested.push_back(Formulation(module.second));
             }
-        // If running MPI job, output with only one processor
-        #if NGEN_WITH_MPI
-        int mpi_rank;
-        MPI_Comm_rank(MPI_COMM_WORLD, &mpi_rank);
```
robertbartel, line 64: With this removal, it looks like the `#include <mpi.h>` earlier in the file is now unnecessary.

include/utilities/parallel_utils.h
```diff
@@ -161,17 +157,16 @@ namespace parallel {
      * @param baseCatchmentFile The base catchment data file from which the rank-specific files are derived.
      * @param baseNexusFile The base nexus data file from which the rank-specific files are derived.
      * @param sendingRank The rank that will be the source and send data to ranks on different hosts.
-     * @param mpi_rank The rank of the current process.
-     * @param mpi_num_procs The number of MPI ranks.
-     * @param hostIdForRank Pointer to array of size ``mpi_num_procs`` that maps ranks (index) to hosts, with each
+     * @param comm The MPI communicator over whose ranks files are distributed.
+     * @param hostIdForRank Pointer to array of size equal to the number of ranks in ``comm`` that maps ranks (index) to hosts, with each
```
robertbartel, line 161: Nit-picking a bit here, but the line length for this is a little long.

src/utilities/parallel_utils.cpp
```diff
@@ -66,80 +74,27 @@ namespace parallel {
...
-        int totalLength = 0;
-        for (int i = 0; i < mpi_num_procs; ++i) {
-            // Displace each rank's string by the sum of the length of the previous rank strings
-            recvDisplacements[i] = totalLength;
-            // Adding the extra to make space for the \0
-            actualHostnameCStrLength[i] += 1;
-            // Then update the total length
-            totalLength += actualHostnameCStrLength[i];
+        MPI_Comm split_by_host;
+        MPI_Comm_split_type(comm, MPI_COMM_TYPE_SHARED, mpi_rank, MPI_INFO_NULL, &split_by_host);
```
robertbartel, line 83: Overall this new approach for the function seems correct (i.e. in application as well as implementation), even if not exactly equivalent.  From what I can tell, that won't matter, and moreover, I don't know how much longer it will be necessary at all.  The only thing that seems to use this is `subdivide_hydrofabric`, and we should consider whether the latter needs to be removed.  That level of change is probably out of scope for this, but let me know if you disagree.

All that said, I think there might be a subtle problem related to MPI version requirements.  `MPI_Comm_split_type` requires MPI-3, but the _CMakeLists.txt_ doesn't specify a version.  The _doc/DEPENDENCIES.md_ file also expressly states there are no version requirements.

I'll admit here that I have not exhaustively confirmed this potential issue wasn't already in place in some other part of the code.  Still, probably worth cleaning up.

robertbartel (changes requested):
I (with some assistance) found one item that does need attention, related to the version requirements for the MPI dependency and related documentation.  But overall I think this looks good. 

I did run one of my standard test simulations (a particular VPU for a particular month and set of modules, etc.).  I confirmed both successful operation and that the same results were produced.

include/realizations/config/formulation.hpp
PhilMiller: Nice, will change, thanks

include/utilities/parallel_utils.h
PhilMiller: That's uneven enough across the codebase that I wasn't really thinking about it. Happy to change if you'd like

src/utilities/parallel_utils.cpp
PhilMiller: I was conscious of the MPI 3 dependency, and that it was probably new to this change.

MPI 3 was standardized in 2012. It should be ubiquitous, long-ago implemented in every library I know of, but I guess is worth at least checking in CMake. I'm not sure it's worth calling out in the documentation, since I think that would mostly create ultimately-unnecessary work for users.

I checked with Chris and Donald that the MPI library on WCOSS2 supports MPI 3. That's probably the oldest possible installation that we'd plan to support anyway.

PhilMiller: Regarding functional equivalence, as far as I can tell from the previous code, its comments, and its call site, any result that gives distinct numbers for processes on distinct hosts is sufficient, without the numbers being contiguous, ordered, or anything else.

I'd agree that we should consider getting rid of `subdivide_hydrofabric`, at least in current form.

src/utilities/parallel_utils.cpp
PhilMiller: My fix for it would probably be to `MPI_Allreduce` the status, and only do the ugly stuff if the status is a failure.

I also don't think "print out an error" really fits well with a function to "identify distributed success/failure"/ At most, this sort of function should just have an output variable of an array of failed processes, if that's even worthwhile

robertbartel (approved):

PR #944, opened 2026-03-18
feat: reduce netcdf reads

aaraney:
NOTE: I need to prune some of the feature creep that has bled into this branch.

Only read data required for simulation.
Increase time dimension read buffer size.

### Previous Behavior

All features in netcdf file were read regardless if they were in the simulation domain.

## Changes

- Only read features from netcdf required for a simulation.
- Increase time dimension read buffer size.

PR #982, opened 2026-06-25
Netcdf Time unit parsing for cf "since" convention support

hellkite500:
PR #944 (commit 
https://github.com/NOAA-OWP/ngen/pull/944/changes/fbfce108ba62c668e0ca7a81e77a731abfea0a6d) introduced this feature on that branch.  Based on review in that PR, this functionality was pulled into its own PR. 

This refactors the meta data parsing and handling from the constructor into some helper functions, then applies the logic from the original commit to enable support for "unit since epoch" time units.

## Additions

- New helper functions which simplify construction logic

## Removals

- Dropped a duplicate warning

## Changes

- Can parse time units as a unit string *or* a "unit since epoch" string.

## Testing

1. Helper functions have tests for the parsing and the conventions being applied.

## Checklist

1. All netcdf tests passing locally

- [x] Linux
- [x] MacOS

### DIFF
--- a/include/forcing/NetCDFPerFeatureDataProvider.hpp
+++ b/include/forcing/NetCDFPerFeatureDataProvider.hpp
@@ -56,15 +56,14 @@ namespace data_access
         {
             TimeUnit unit;                  //the unit raw time values are stored in
             double scale_factor;            //multiplier converting a raw value to seconds
-            std::time_t epoch_start_time;   //reference epoch, in seconds since the Unix epoch
+            std::optional<std::time_t> epoch_start_time;   //reference epoch, in seconds since the Unix epoch, if specified
         };
 
         /**
          * @brief Interpret a NetCDF time `units` string as a time unit and scale factor.
          *
-         * The returned @ref TimeInfo carries the matched unit and scale factor; its
-         * @ref TimeInfo::epoch_start_time is left at the Unix epoch (0) since a bare units
-         * string does not, on its own, specify a reference epoch.
+         * Handles both bare units and the CF form "<unit> since <date>", in which case
+         * the parsed reference epoch is returned in the result.
          *
          * @param units_str the raw `units` attribute value (e.g. "ns", "hours")
          * @return the parsed @ref TimeInfo, or std::nullopt if @p units_str is unrecognized
--- a/include/forcing/NetCDFPerFeatureDataProvider.hpp
+++ b/include/forcing/NetCDFPerFeatureDataProvider.hpp
@@ -18,6 +18,8 @@
 #include <mutex>
 #include "assert.h"
 #include <iomanip>
+#include <optional>
+#include <ctime>
 #include <boost/compute/detail/lru_cache.hpp>
 
 #include <UnitsHelper.hpp>
@@ -46,6 +48,37 @@ namespace data_access
             TIME_MICROSECONDS,
             TIME_NANOSECONDS
         };
+
+        /**
+         * @brief Time metadata for netcdf time units.
+         */
+        struct TimeInfo
+        {
+            TimeUnit unit;                  //the unit raw time values are stored in
+            double scale_factor;            //multiplier converting a raw value to seconds
+            std::time_t epoch_start_time;   //reference epoch, in seconds since the Unix epoch
+        };
+
+        /**
+         * @brief Interpret a NetCDF time `units` string as a time unit and scale factor.
+         *
+         * The returned @ref TimeInfo carries the matched unit and scale factor; its
+         * @ref TimeInfo::epoch_start_time is left at the Unix epoch (0) since a bare units
+         * string does not, on its own, specify a reference epoch.
+         *
+         * @param units_str the raw `units` attribute value (e.g. "ns", "hours")
+         * @return the parsed @ref TimeInfo, or std::nullopt if @p units_str is unrecognized
+         */
+        static std::optional<TimeInfo> interpret_time_units(const std::string& units_str);
+
+        /**
+         * @brief Parse a reference-epoch timestamp into seconds since the Unix epoch.
+         *
+         * @param epoch_str the timestamp text (e.g. "01/01/1970 00:00:00")
+         * @param format a std::get_time / strptime style format string
+         * @return seconds since the Unix epoch
+         */
+        static std::time_t parse_epoch(const std::string& epoch_str, const std::string& format);
 
         /**
          * @brief Factory method that creates or returns an existing provider for the provided path.
@@ -143,6 +176,19 @@ namespace data_access
         const std::string& get_ncvar_units(const std::string& name);
 
         void test_data_is_readable();
+
+        /**
+         * @brief Read the time unit, scale factor and reference epoch from a `Time` variable.
+         *
+         * Reads the `units` and `epoch_start` attributes and resolves them via
+         * @ref interpret_time_units and @ref parse_epoch, warning and falling back to
+         * sensible defaults (seconds, Unix epoch) when an attribute is absent or
+         * unrecognized.
+         *
+         * @param time_var the file's `Time` variable
+         * @return the resolved time metadata
+         */
+        TimeInfo get_time_metadata(const netCDF::NcVar& time_var);
 
     };
 }
--- a/src/forcing/NetCDFPerFeatureDataProvider.cpp
+++ b/src/forcing/NetCDFPerFeatureDataProvider.cpp
@@ -77,9 +77,10 @@ std::time_t NetCDFPerFeatureDataProvider::parse_epoch(const std::string& epoch_s
     std::tm tm{};
     std::stringstream s(epoch_str);
     s >> std::get_time(&tm, format.c_str());
-    //std::time_t epoch_start_time = mktime(&tm);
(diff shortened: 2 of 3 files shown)


PhilMiller:
The NGWPC PR I think touches on this also. Maybe worth picking out and integrating their changes, and then building on that accordingly, to minimize conflicts.

src/forcing/NetCDFPerFeatureDataProvider.cpp
```diff
@@ -30,6 +30,138 @@ void NetCDFPerFeatureDataProvider::cleanup_shared_providers()
...
+    std::string time_unit_str = units_str;
+    std::string unit_epoch_str = "";
+
+    // CF conventions may have units of the form "<unit> since <date>"
+    size_t pos = time_unit_str.find(" since ");
+    if(pos != std::string::npos)
+    {
+        unit_epoch_str = time_unit_str.substr(pos + 7); // 7 is length of " since "
+        time_unit_str.erase(pos);
+    }
```
PhilMiller:
```suggestion
    std::string since = " since ";
    size_t since_pos = units_str.find(since);
    if(since_pos != std::string::npos)
    {
        time_unit_str = units_str.substr(0, since_pos);
        unit_epoch_str = units_str.substr(since_pos + since.length());
    }
```
I think it's preferable to express this as splitting the input `units_str` on `" since "`

src/forcing/NetCDFPerFeatureDataProvider.cpp
```diff
@@ -30,6 +30,138 @@ void NetCDFPerFeatureDataProvider::cleanup_shared_providers()
...
+    else
+    {
+        return std::nullopt;
+    }
+
+    if(!unit_epoch_str.empty())
+    {
+        //CF convention in time units formats as "YYYY-MM-DD HH:MM:SS"
+        info.epoch_start_time = parse_epoch(unit_epoch_str, "%Y-%m-%d %H:%M:%S");
+    }
```
PhilMiller: Rather than setting `unit_epoch_str` up above, and then parsing it down here, why not just (attempt to) parse it at the point it's extracted?

src/forcing/NetCDFPerFeatureDataProvider.cpp
```diff
@@ -30,6 +30,138 @@ void NetCDFPerFeatureDataProvider::cleanup_shared_providers()
...
+            time_unit_att.getValues(time_unit_str);
+        }
+
+        // set time unit, scale factor, and (for CF units) the reference epoch
+        if ( auto parsed = interpret_time_units(time_unit_str) )
+        {
+            info = *parsed;
+        }
+        else {
+            log_stream << "Warning using defualt time units\n";
```
PhilMiller:
```suggestion
            log_stream << "Warning using default time units\n";
```
There's no need to reproduce this old typo

src/forcing/NetCDFPerFeatureDataProvider.cpp
```diff
@@ -30,6 +30,138 @@ void NetCDFPerFeatureDataProvider::cleanup_shared_providers()
...
+        }
+    }
+    catch(const netCDF::exceptions::NcException& e){
+        std::cerr<<e.what()<<std::endl;
+        log_stream << "Warning: Couldn't read time unit attribute, using defualt time unit of Seconds\n";
```
PhilMiller:
```suggestion
        log_stream << "Warning: Couldn't read time unit attribute, using default time unit of Seconds\n";
```
Ditto typo

src/forcing/NetCDFPerFeatureDataProvider.cpp
```diff
@@ -30,6 +30,138 @@ void NetCDFPerFeatureDataProvider::cleanup_shared_providers()
...
+
+    // fall back to the epoch_start attribute only when the units string didn't supply an epoch
+    if ( !info.epoch_start_time.has_value() )
+    {
+        try {
+            auto epoch_att = time_var.getAtt("epoch_start");
+
+            if ( epoch_att.isNull() )
+            {
+                log_stream << "Warning using defualt epoc string\n";
```
PhilMiller:
```suggestion
                log_stream << "Warning using default epoch string\n";
```

src/forcing/NetCDFPerFeatureDataProvider.cpp
```diff
@@ -30,6 +30,138 @@ void NetCDFPerFeatureDataProvider::cleanup_shared_providers()
...
+            else
+            {
+                std::string epoch_start_str;
+                epoch_att.getValues(epoch_start_str);
+                info.epoch_start_time = parse_epoch(epoch_start_str, "%D %T");
+            }
+        }
+        catch(const netCDF::exceptions::NcException& e) {
+            std::cerr<<e.what()<<std::endl;
+            log_stream << "Warning using defualt epoc string\n";
```
PhilMiller:
```suggestion
            log_stream << "Warning using default epoch string\n";
```

src/forcing/NetCDFPerFeatureDataProvider.cpp
```diff
@@ -30,6 +30,138 @@ void NetCDFPerFeatureDataProvider::cleanup_shared_providers()
...
+    }
+    else if ( time_unit_str ==  "s" || time_unit_str == "seconds" )
+    {
+        info.unit = TIME_SECONDS;
+        info.scale_factor = 1;
+    }
+    else if ( time_unit_str ==  "ms" || time_unit_str == "milliseconds" )
+    {
+        info.unit = TIME_MILLISECONDS;
+        info.scale_factor = .001;
```
PhilMiller: Since this is all being moved around and revised, I'd prefer to have these factors written in scientific notation:

```suggestion
        info.scale_factor = 1.0e-3;
```
Etc, so that reading to review for correctness is not a matter of counting `0` characters

hellkite500: Done in 9f51f948e

src/forcing/NetCDFPerFeatureDataProvider.cpp
hellkite500: Done in 9f51f948e

src/forcing/NetCDFPerFeatureDataProvider.cpp
hellkite500: Done in 9f51f948e

src/forcing/NetCDFPerFeatureDataProvider.cpp
hellkite500: Done in 9f51f948e

src/forcing/NetCDFPerFeatureDataProvider.cpp
hellkite500: Done in 9f51f948e

src/forcing/NetCDFPerFeatureDataProvider.cpp
hellkite500: Done in 9f51f948e

src/forcing/NetCDFPerFeatureDataProvider.cpp
hellkite500: Done in 9f51f948e

hellkite500:
> The NGWPC PR I think touches on this also. Maybe worth picking out and integrating their changes, and then building on that accordingly, to minimize conflicts.

Given the refactoring step of the meta data functions, the conflicts are going to be real regardless, but I'm of the opinion that having these helpers with tests help identify future drift are good to have in place.

Conceptually, the NGWPC PR that touches on the units/epoch mechanics that are worked here are similar, and the only real mechanical difference is a "throw" vs "warning" on "unknown unit".  But both branches take a similar "use default seconds unit when no attribute is present", which seems somewhat counter to me...I left the warning for now, this can be trivially changed if needed.

The rest of the provider changes from that PR touch other parts of the constructor/get_value functions that shouldn't be impacted here.  The takeaway is a quick manual conflict resolution that simply replaces the inline constructor code (in NGWPC) for handling the unit meta data to the the functions used here.  As I get to the heart of #944, I'll take a close look at the dimension handling more closely, and may pull some of that into there more concretely. If that happens, then the resolution on the provider would simply be `theirs` on this file, as the functional pieces would be already be hand picked and rolled in.

include/forcing/NetCDFPerFeatureDataProvider.hpp
```diff
@@ -47,6 +49,36 @@ namespace data_access
...
+         *
+         * @param units_str the raw `units` attribute value (e.g. "ns", "hours")
+         * @return the parsed @ref TimeInfo, or std::nullopt if @p units_str is unrecognized
+         */
+        static std::optional<TimeInfo> interpret_time_units(const std::string& units_str);
+
+        /**
+         * @brief Parse a reference-epoch timestamp into seconds since the Unix epoch.
+         *
+         * @param epoch_str the timestamp text (e.g. "01/01/1970 00:00:00")
```
PhilMiller, line 76: Obviously not something to change in this PR, but I wish CF conventions used ISO 8601 format

PhilMiller: Oh, in the source where this is called, it pretty nearly does. It's just this example in the comment doesn't

PhilMiller (approved):

PR #983, opened 2026-06-25
Netcdf Forcing Provider QoL

hellkite500:
The following commits from [PR #944](https://github.com/NOAA-OWP/ngen/pull/944) were pulled into this branch.

- [optional linking with hdf5 for enhanced netcdf filters](https://github.com/NOAA-OWP/ngen/pull/944/changes/9cc6f4d36c5bf7176aa25bba5947996f1f11e67f) — CMake reworked to link the exported HDF5 target rather than the raw library files
- [error reporting on failed netcdf file open](https://github.com/NOAA-OWP/ngen/pull/944/changes/71f0eaab6071d5473ab87ab3890fcfc577d257d8)
- [ensure netcdf variables are readable](https://github.com/NOAA-OWP/ngen/pull/944/changes/aa81a64578170840c00be42c38cfb3e5868c0d76)
- [improve method name readability (rename to `test_data_is_readable`)](https://github.com/NOAA-OWP/ngen/pull/944/changes/af31de5a9389f8d0e52844b4541dc75361cfa623)
- [throw if no netcdf features](https://github.com/NOAA-OWP/ngen/pull/944/changes/17a21c11fea783f8ac998c6b0eae99e79e41f5df)
- [use simpler syntax for getting vector backing pointer](https://github.com/NOAA-OWP/ngen/pull/944/changes/f498acff3f005fbde6c459112b1bf71f15025eb7)

## Additions

- Optional build flag for linking HDF5
- Runtime check that netcdf vars are readable at construction time

## Changes

- Errors thrown when files are not readable or no features are read

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows project standards (link if applicable)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

PR #984, opened 2026-06-25
Feature: netcdf chunked reads

hellkite500:
This is the core of #944 after factoring out #982 and #983.  The 18 remaining commits were logically grouped into the 6 on this branch.  Note that this is based on the HEAD of #982 currently, will need rebased once that merges.

## Additions

- Catchment-id **hinting** (`hint_shared_provider_id`) so the provider reads only the catchments a simulation requests, not every feature in the file.
- A **time-dimension value cache** that reads a span of time slices per access instead of repeated single-step reads.
- `align_cache_with_chunks()` to size the cache from the variable's on-disk chunking when present.
- Unit tests for the chunked/hinted read paths (`...UsingIds`, `...Span`, `...MultiSpan`, etc.).

## Removals

- Reading **all** features in the file regardless of the simulation domain — only required catchments are now read.
- The default constructor (`= delete`) — there is no valid un-parameterized provider.

## Changes

- `get_value` reworked to read through, and serve from, the chunked time-dimension cache.
- Cache size expressed as a named const (`N_EXPECTED_FORCING_VARS`) rather than a magic number.

## Testing

1. All netcdf provider tests pass locally

## Known issue: cache sizing

Per @PhilMiller's profiling note on #944: the cache is currently sized to match the working set. A functional cache needs to be **larger** than the working set — a request spanning more than one time step will thrash an exactly-matched cache. Carried over from #944.

## Coordination

- `get_value` overlaps the NGWPC reader rework in #979 (its dimension-ordering / `imap` changes touch the same function). Will look at cherry picking that work here.

## Checklist

- [x] Code follows project standards
- [ ] New functions are documented
- [ ] Project documentation has been updated (CHANGELOG "Unreleased")
- [x] Reviewers requested

- [x] Linux
- [x] MacOS

### DIFF
--- a/src/forcing/NetCDFPerFeatureDataProvider.cpp
+++ b/src/forcing/NetCDFPerFeatureDataProvider.cpp
@@ -23,6 +23,7 @@ std::shared_ptr<NetCDFPerFeatureDataProvider> NetCDFPerFeatureDataProvider::get_
     return p;
 }
 
+
 void NetCDFPerFeatureDataProvider::cleanup_shared_providers()
 {
     const std::lock_guard<std::mutex> lock(shared_providers_mutex);
@@ -251,6 +252,10 @@ NetCDFPerFeatureDataProvider::NetCDFPerFeatureDataProvider(std::string input_pat
         throw std::runtime_error("Provided NetCDF file has no features");
     }
 
+    // include all catchments in the "default" chunk
+    auto pair = std::pair<size_t, size_t>(0, num_ids);
+    chunks.push_back(pair);
+
     //TODO: split into smaller slices if num_ids is large.
     cache_slice_c_size = num_ids;
 
@@ -267,7 +272,9 @@ NetCDFPerFeatureDataProvider::NetCDFPerFeatureDataProvider(std::string input_pat
         loc_ids.push_back(str);
         id_pos[str] = loc++;
     });
-
+    // Make sure we were able to read an actual string type
+    // each id should start with "cat-"
+    assert(loc_ids[0].size() > 4);;
     // correct string release
     nc_free_string(num_ids,&string_buffers[0]);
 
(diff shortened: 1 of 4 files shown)


aaraney (dismissed):
Looks good to me!

src/forcing/NetCDFPerFeatureDataProvider.cpp
```diff
@@ -575,6 +727,34 @@ void NetCDFPerFeatureDataProvider::test_data_is_readable() {
...
+        var.getChunkingParameters(mode, chunk_sizes);
+        if (mode == netCDF::NcVar::ChunkMode::nc_CHUNKED && chunk_sizes.size() >= 2) {
+            cache_slice_t_size = chunk_sizes[1];  // Assume second dimension is time
+            break;  // Use the first suitable chunk size found
+        }
+    }
+    if(cache_slice_t_size > num_times){
+        cache_slice_t_size = num_times;
+    }else if(cache_slice_t_size <= 0){
+        cache_slice_t_size = 24; // default to 24 if no chunking info found
```
PhilMiller: I don't think this case can ever be hit, because there's the in-class initializer `cache_slice_t_size = 18;`. Did you mean to change that to `= 0` when you added this?

src/forcing/NetCDFPerFeatureDataProvider.cpp
```diff
@@ -575,6 +727,34 @@ void NetCDFPerFeatureDataProvider::test_data_is_readable() {
...
+void NetCDFPerFeatureDataProvider::align_cache_with_chunks()
+{
+    auto num_times = nc_file->getDim("time").getSize();
+
+    auto var_set = nc_file->getVars();
+    for (const auto& var_pair : var_set) {
+        const std::string& var_name = var_pair.first;
+        auto var = var_pair.second;
+        // Skip the Time variable itself and ids, look for data variables with at least 2 dimensions
+        if (var_name == "Time" || var_name == "ids" || var_name == "catchment-id" || var.getDimCount() < 2) continue;
```
PhilMiller, line 739: I think the `DimCount` check covers all of this?

PhilMiller (changes requested):
I'm concerned about `cache_slice_t_size` as noted above

PhilMiller (approved):

src/forcing/NetCDFPerFeatureDataProvider.cpp
hellkite500: Fixed and squashed.

PR #988, opened 2026-06-30
Fix Nexus ID Handling in NetCDF Output

robertbartel:
Account for nexus ids with identical numeric components existing in geospatial domain, and ensure proper behavior of nexus output writing.  In particular, handle the scenario, e.g., with `nex-1` and `tnx-1` (the latter being a terminal nexus), without having an id collision in NetCDF output files, by using full string nexus ids in NetCDF output.

Implemented by switching the nexus id variable to a fixed-length `NC_CHAR` type to maintain compatibility with parallel write support.  Adds another dimension for this length so it can be applied to the nexus id NetCDF variable.

Also includes a helper function for packing the ids into null-padded char buffers, to facilitate MPI-gather approach to multi-rank writes.


robertbartel:
Pinging @shorvath-noaa on this in case he wants to follow the issue.

Also, I did confirm that the CSV files do not need any changes related to this.   That manager doesn't actually write the nexus id to the CSV file contents; rather, the full nexus id was already a part of the filename.

PhilMiller:
I haven't looked at the code yet, but for the ID gathering, would the existing `gather_strings` be suitable?

robertbartel:
> I haven't looked at the code yet, but for the ID gathering, would the existing gather_strings be suitable?

I think so @PhilMiller.  Good call; I'd forgotten it was there.  I'll give this another once-over to make sure there isn't something I'm not considering, and then tweak things to re-use that instead.

include/utilities/output/PerFormulationNexusOutputMgr.hpp
```diff
@@ -184,10 +184,44 @@ namespace utils
 
         static std::string parse_netcdf_return_code(const int nc_status);
 
+        /**
+         * Fixed character width used to store each nexus feature id string in the managed NetCDF file.
+         *
+         * The nexus identity is written as a fixed-width, null-padded character representation rather than a numeric
+         * value, so that the full id string (prefix included, e.g. ``nex-1`` vs ``tnx-1``) is preserved and distinct
+         * ids never collapse together.  This same width sizes both the packing helper (@ref pack_nexus_id) and the
+         * NetCDF string-length dimension.
```
PhilMiller:
```suggestion
         * Fixed character width used to store each nexus feature ID string in the managed NetCDF file.
         *
         * The nexus identity is written as a fixed-width, null-padded character representation rather than a numeric
         * value, so that the full ID string (prefix included, e.g. ``nex-1`` vs ``tnx-1``) is preserved and distinct
         * IDs never collapse together.  This same width sizes both the packing helper (@ref pack_nexus_id) and the
         * NetCDF string-length dimension.
```

Unless we're talking about the primal urges of the various nexus features, of course.

include/utilities/output/PerFormulationNexusOutputMgr.hpp
```diff
@@ -184,10 +184,44 @@ namespace utils
...
+         * remaining bytes up to @ref nexus_id_string_width are set to the null character.  An id whose length exactly
+         * equals @ref nexus_id_string_width fills the entire buffer with no null terminator.  To avoid silently
+         * truncating an unexpectedly long id, an id longer than @ref nexus_id_string_width is a hard error (consistent
+         * with the manager's existing throw-on-malformed-id behavior).
+         *
+         * @param nexus_id The full nexus id string (prefix included) to pack.
+         * @param buffer Pointer to a caller-provided buffer of at least @ref nexus_id_string_width chars to pack into.
+         * @throw std::runtime_error if the id length exceeds @ref nexus_id_string_width.
+         */
+        static void pack_nexus_id(const std::string& nexus_id, char* buffer);
```
PhilMiller, line 210: Could that be `boost::span<char>` instead of `char*`?

include/utilities/output/PerFormulationNexusOutputMgr.hpp
```diff
@@ -184,10 +184,44 @@ namespace utils
...
+
         const std::string nc_dim_name_nexus_id = "feature_id";
         const std::string nc_dim_name_time = "time";
         const std::string nc_var_name_flow = "runoff_rate";
```
PhilMiller: I just noticed - these should probably be `static`.

With C++20, we'll even be able to make them `constexpr` @christophertubbs

include/utilities/output/PerFormulationNexusOutputMgr.hpp
```diff
@@ -184,10 +184,44 @@ namespace utils
...
 
+        /**
+         * Name of the NetCDF dimension giving the fixed character length of each stored nexus feature id string.
+         *
+         * The ``feature_id`` variable is a 2-D character array dimensioned by the nexus dimension and this
+         * string-length dimension (whose size is @ref nexus_id_string_width), so the full id string is preserved
+         * verbatim rather than reduced to a numeric value.
+         */
+        const std::string nc_dim_name_id_str_len = "feature_id_str_len";
```
PhilMiller:
```suggestion
        static const std::string nc_dim_name_id_str_len = "feature_id_str_len";
```

src/utilities/CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ if(NGEN_WITH_MPI)
     target_link_libraries(ngen_parallel
       PUBLIC
         pybind11::embed
+        NGen::python
```
PhilMiller, line 25: Is this an intended part of this PR? Is it because of `subdivide_hydrofabric`?

PhilMiller:
This is maybe a bit of over-engineering, but would it make sense to do an `MPI_Reduce(MPI_MAX)` on the nexus feature ID lengths, and use that to dimension the array, rather than hard-coding a limit? It seems like a small cost now, to spare someone later who comes in with long, descriptive nexus IDs and sees a job failure. E.g. `"nex-upper-colorado-mile-137.4"`

include/utilities/output/PerFormulationNexusOutputMgr.hpp
robertbartel: Eh, I suppose it could, but does that really get us much in this situation, other than encapsulating the size (which is already constant)?

PhilMiller:
And the flip side of that is that with the current relatively compact identifiers, the array will be smaller.

include/utilities/output/PerFormulationNexusOutputMgr.hpp
PhilMiller: If nothing else, it clarifies that it's a particular block of memory with that size, and not a null-terminated C-style string starting at that address.

include/utilities/output/PerFormulationNexusOutputMgr.hpp
robertbartel: I'd say out of scope, but the new item should be added this way, and it makes sense to ensure they are all consistent.  I'll change shortly.

PhilMiller (dismissed):
With `gather_strings` no longer duplicated, I'll leave it to your discretion whether/how to address the other comments.

include/utilities/output/PerFormulationNexusOutputMgr.hpp
robertbartel: As noted elsewhere, yes, this make sense, but I'll change separately and make the others consistent with this at the same time.

src/utilities/CMakeLists.txt
robertbartel: It is intended.  I did not dig heavily into why as part of this, but adding the `ngen_parallel` library as a dependency of the `ngen_output` library (needed for the `gather_strings` change you suggested) surfaced an under-declared issue that was causing a failure in the Actions checks.

As for whether `subdivide_hydrofabric` is responsible, or something else, that seems like something that really shouldn't be in scope for this.  I did say "shouldn't" rather than "isn't," as I see a reasonable argument for it being in scope.   But I'd also argue that if it is, it'd be better to roll-back everything for the `gather_strings` adjustment just to avoid having to deal with that here.

PhilMiller: It's fine. We can consider splitting that out of `parallel_utils` down the line.

PhilMiller (approved):

robertbartel:
> This is maybe a bit of over-engineering, but would it make sense to do an MPI_Reduce(MPI_MAX) on the nexus feature ID lengths, and use that to dimension the array, rather than hard-coding a limit? It seems like a small cost now, to spare someone later who comes in with long, descriptive nexus IDs and sees a job failure. E.g. "nex-upper-colorado-mile-137.4"

I thought about something dynamic like this but decided against it.   I think the cost is minimal, and I had initial concerns about the potential for output files made by the same version of ngen to potentially be differently sized in this regard.  For expediency, I didn't think that all the way through to conclude if there was actually a problem, so we could reconsider.  It just didn't seem worth it.  There might have been another reason also that I'm forgetting at the moment.

PhilMiller:
I don't think there'd be any problem with the nexus ID array being variably sized, since readers should be looking at the size in the file regardless. Ngen might increase the size later, so they can't use a constant. I don't think there's any sort of resource exhaustion/DoS threat, either, since these IDs are already in memory in ngen.

Regardless, your call on whether it's worth implementing now.

include/utilities/output/PerFormulationNexusOutputMgr.hpp
```diff
@@ -184,10 +184,44 @@ namespace utils
...
 
+        /**
+         * Fixed character width used to store each nexus feature id string in the managed NetCDF file.
+         *
+         * The nexus identity is written as a fixed-width, null-padded character representation rather than a numeric
+         * value, so that the full id string (prefix included, e.g. ``nex-1`` vs ``tnx-1``) is preserved and distinct
+         * ids never collapse together.  This same width sizes both the packing helper (@ref pack_nexus_id) and the
+         * NetCDF string-length dimension.
+         */
+        static constexpr size_t nexus_id_string_width = 32;
```
christophertubbs, line 195: Not a dis-qualifier, but I'm not a huge fan of that hardcoded 32 for nexus id widths. That's probably valid for 99.99999% of all expected use cases going forward, but wouldn't the system be a bit more flexible if it were able to determine the width from all input ids at runtime? If one of the core ideas behind next gen was to allow just about anyone to experiment with really weird ideas (aside from being able to build a production series of models), wouldn't a firm size be a bit of a damper? What does everyone else think? I understand that saying "Hey! Let's jump through all the ranks to find the widest size!" isn't trivial.

robertbartel: @christophertubbs my thought on this was that it was safer (preferable when under time constraints) and, as noted, should be good enough for the foreseeable future.  There is something that doesn't sit well with me about the ngen having nexus IDs in NetCDF output be of "fixed" but not predetermined length, and then have that fix-length be circumstantial (in an obfuscated manner) and based primarily on external data for a particular simulation. 

I can't immediately point to a problem with this, but I would expect any problems to be subtle and lagged if they ever manifest.  Contrast that with this approach:  there is a good chance this will be sufficient for the lifetime of the framework, and it will be immediately clear for a scenario when it isn't.

I'm certainly open to discussing improvements to this in the future, though.

PR #989, opened 2026-07-01
perf: return catchment id by const reference

hellkite500:
get_catchment_id() returned std::string by value, allocating a copy on every call. It is invoked per-catchment, per-timestep on the formulation output path (and in CatchmentAggrDataSelector construction), so the copies add up on large/CONUS runs.


PhilMiller (approved):
I don't mind the copy avoidance, though I'm curious whether there's data supporting this being a noticeable cost, or just intuition. If there is data, it would be good to make some note of it here for future reference.

hellkite500:
This one was more of a side effect of the initial "Seeing a lot of memory pressure running conus runs".  #990 was the biggest winner from actual testing, but this change went along for the ride.

PR #973, opened 2026-05-21
Formulation_Manager refactor

PhilMiller:
Preparatory for the `NgenSimulation` refactoring.

## Changes

- Pull `Simulation_Time_Object` out of `Formulation_Manager` to a local variable in `main()`

## Testing

1. Local ctest
2. Github CI

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows project standards (link if applicable)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

PR #991, opened 2026-07-01
Fix bug where catchment output goes to stdout if CSV output disabled

robertbartel:
Through a combination of #943 and #973, a bug was introduced such that, if catchment CSV output was disabled, catchment output data would instead be written directly to standard out.  

This resolves that bug by avoiding using/keeping the default stdout stream as the output stream for catchment formulations.  If regular CSV output is disabled, handle this case and explicitly set to the null stream.


PhilMiller (approved):
I'd really like to make the notion that `Formulation` has any sort of responsibility for output go away, but I guess it needs to keep working correctly in the meantime.

PR #992, opened 2026-07-01
Fix bug in nexus output for serial

robertbartel:
A bug was preventing the contributing catchments from being set properly for nexuses in serial executions.  This had two side effects:  noise in the logging, and all `0` values for nexus CSV output.  This addresses that by properly associating the upstream.

### DIFF
--- a/src/core/HY_Features.cpp
+++ b/src/core/HY_Features.cpp
@@ -63,8 +63,9 @@ HY_Features::HY_Features(network::Network network, std::shared_ptr<Formulation_M
         }
         else if(hy_features::identifiers::isNexus(feat_type))
         {
+            origins = network.get_origination_ids(feat_id);
             _nexuses.emplace(feat_id, std::make_unique<HY_PointHydroNexus>(
-                                          HY_PointHydroNexus(feat_id, destinations) ));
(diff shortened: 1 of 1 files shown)


PhilMiller:
I'm trying to dig a bit into the details here.

The noisiness is about the warning message from `SurfaceLayer::update_models`?

Which exact things were producing incorrect 0 outputs? Did this only become broken recently, and if so, by what?

aaraney:
Using this seems to have fixed the issue. Thanks, @robertbartel!

aaraney (approved):

robertbartel:
@PhilMiller, if I'm following things correctly, the contributing catchments were simply never being set for nexus objects in serial builds.    Actual flow values were still being set and accessed as needed, somehow.  Just not the catchments.   This seems to go really far back in the history of the repo.

But, since the flow values were there, the `HY_PointHydroNexus::get_downstream_flow` function _mostly_ worked correctly, and typically nexus output files got the right values.   I say "mostly" though, because there were problems with certain hydrofabric edge cases causing errors.  I "fixed" this in #931 earlier this year, evidently without anyone noticing that all the outputs were `0` because of the other obscured issue.

PR #998, opened 2026-07-07
Refactor existing catchment output handling and add features required to optimize large runs

hellkite500:
This PR features a core optimization needed for CONUS scale runs on a distributed file system, namely output aggregation. In its current form, ngen will create a single catchment output file for each divide being modeled, and put them in `output_root`.  On a distributed file system, this consumes a lot of open file handles, and introduces contention at initialization in creating all the output files.

A *simple* aggregation method was implemented to work around this, but required touching a lot of the output code to bring into a form that was mergable (e.g. configurability, testing, ect...).

So this PR was motivated primarily by two distributed-filesystem optimizations: per-rank catchment output aggregation, and a per-rank subdirectory for nexus CSV output:

d7680649a feat(output): per-rank subdirectory for nexus CSV output

(the nexus optimization being somewhat subsumed by the netcdf nexus output feature).

Since getting the catchment aggregation mechanism properly implemented would have touched a significant portion of the catchment I/O code, it was decided to tackle a long overdue refactor of the model output mechanics.  The refactor establishes an abstract output-manager interface (similar to the NexusOutputManager) and cuts all csv/string output handling out of the core formulations, which now deal in typed (double) outputs and a positionally aligned header vector.  The catchment aggregation mechanism rides directly on this manager, so it lands together with the abstraction:

9965a7635 feat(output): add catchment output manager abstraction (CSV backend)

The ability to actually *use* aggregation requires some configuration management, introduced in

5e72e762f feat(output): add the output configuration block

which encapsulates output configuration for parsing from the realization configuration file (and it maintains backwards compatibility, for now).

With the configuration in place, a factory selects and constructs the appropriate manager from that configuration:

257d982dd feat(output): select and route catchment output through the configuration

The driver and layers are then wired to build output managers via that factory and route formulation output through them (formulations supply typed values; the manager owns the file layout, including aggregation):

42909c46f refactor(output): route catchment output through the manager

The last functional part is the distributed-filesystem optimization for nexus csv outputs, letting each rank store its outputs in a sub-directory to reduce locking contention:

d7680649a feat(output): per-rank subdirectory for nexus CSV output

this feature was implemented on the existing nexus csv manager, with configuration mechanics built on top of 5e72e762f.

Finally, documentation for the new output configuration block:

6a01598dc docs(config): document the output configuration block

Sorry for the roundabout history/description there.  All that to say, reviewing in commit order may be most helpful, using this description as the guide post.

## Additions

- `utils::CatchmentOutputsMgr` — an abstract, typed catchment-output interface modeled on `NexusOutputsMgr`, keyed by `(formulation_id, catchment_id)`, with a CSV backend `CatchmentCsvOutputMgr`.
- Per-rank aggregated catchment CSV: all of a rank's catchments in one shared file (leading `catchment_id` column, header written once) as an alternative to one file per catchment.
- `realizations::config::Output` — a parsed `output` config block: per-domain (catchment/nexus) `enable` / `format` / `grouping` / `rank_subdir` plus a global value `precision`, with fallback to the deprecated top-level keys; also normalizes the output root (trailing-slash) at construction, without touching the filesystem — the managers create their own directories.
- `realization::config::make_catchment_output_mgr` — factory that builds the backend and layout from the parsed config.
- Per-rank subdirectory (`rank_<N>/`) for nexus CSV output under MPI.
- Unit tests: `Output_Test` (config parsing/resolution), `CatchmentOutput_Test` (factory selection), `CatchmentCsvOutputMgr_Test` (CSV layouts), `PerNexusCsvOutputMgr_Test` (nexus subdir).

## Removals

- Catchment-output responsibility removed from `HY_Features` / `HY_Features_MPI` — features are now topology-only; the driver owns the manager.
- Legacy string-projection formulation output API (`get_output_header_line` / `get_output_line_for_timestep`) and the dead `HY_CatchmentArea::set_output_stream` / `write_output`.
- Per-field output accessors on `Formulation_Manager` (`get_output_root`, `is_disable_catchment_output`, `is_using_per_formulation_nexus_files`) — consolidated into a single `get_output_config()`.

## Changes

- Formulations now expose typed output — `get_output_fields()` (hoisted to `Catchment_Formulation`) and a new `get_output_values_for_timestep()` returning `vector<double>` — replacing the string/stream path.
- `NGen` constructs, registers per catchment, and injects the catchment output manager into the layers; output is committed once at end-of-run.
- `DomainLayer` output is keyed as `(formulation_id="layer_<id>", catchment_id=<name>)`, writing to `<root>layer_<id>/<name>.csv`.
- The parsed `Output` struct is the single source of truth; deprecated top-level keys are honored only through its construction, so existing configs are unaffected (with a one-time deprecation warning).

## Testing

1. Build from the build root, then run the affected suites (verified locally, macOS/Clang with MPI + NetCDF + Python):
   - `test_realization_config` — **24 passed** (Output parsing/resolution, factory selection, `Formulation_Manager` construction + one real-formulation integration)
   - `test_catchment_output_mgr` — **7 passed** (per-feature, aggregated header-once + split-by-formulation, per-formulation subdir, throw-after-close, throw-on-unregistered-catchment, precision rendering)
   - `test_nexus_output` — **2 passed** (nexus per-rank subdir)
2. The BMI formulation suites (`Bmi_*_Formulation_Test`) were updated for the typed output API; run the full set via `ctest` / CI for the remaining suites.

## Notes

- Behavior-preserving by default: `grouping=per_feature` keeps the long-standing flat `<root><catchment_id>.csv`; aggregation and rank subdirs are opt-in via config.
- CSV is the only backend today. The interface is data-push (no formulation-type dependency), so a NetCDF/parquet backend can be added as a sibling without touching formulations; non-CSV formats currently throw.

## Todos

- Once merged, the wiki will need updated as well based on the configuration keys/changes.
- This refactor enables several possible next steps which should come as separate PR's
- One critical TODO: There is now significant dead code that should be removed, but I think that's best left as a follow up PR to keep this reviewable. 

## Checklist

- [x] Code can be automatically merged (no conflicts) — branch is based on current `origin/master` (0 commits behind)
- [x] Code follows project standards
- [x] Passes all existing automated tests — *output-related suites verified locally; full suite via CI (see Testing)*

All newly added and refactored tests have been run/passed locally

- [x] Linux
- [x] MacOS

### DIFF
--- a/doc/REALIZATION_CONFIGURATION.md
+++ b/doc/REALIZATION_CONFIGURATION.md
@@ -21,15 +21,38 @@ The Configuration is a key-value object and must contain these three first level
 
 ## Optional Top-Level Keys
 
+### `output`
+The configuration may optionally contain an `output` object that controls where and how catchment and nexus output are written. All fields are optional:
+
+* `root` — root output directory for both catchment and nexus output; created if it does not already exist (default: the working directory, `./`).
+* `precision` — number of significant digits used when a text (CSV) backend renders values (default: `9`).
+* `catchment` and `nexus` — per-domain settings, each an object with:
+  * `enable` — whether this domain's output is written (default: `true`; set `catchment.enable` to `false` to skip catchment files).
+  * `format` — serialization format, `"csv"` (default) or `"netcdf"`. NetCDF is currently supported for nexus output only.
+  * `grouping` — how a domain's records are distributed across files (the file-*granularity* axis):
+    * `"per_feature"` (default) — one file per catchment / per nexus.
+    * `"per_formulation"` — the catchments of a formulation aggregate into one file with a leading `catchment_id` column (one file per formulation). Fewer, larger files, with columns uniform within each file by construction. (Nexus CSV aggregation is not yet implemented.)
+  * `rank_subdir` — whether an MPI rank's files are placed under a `rank_<N>/` subdirectory (the filesystem-*layout* axis, orthogonal to `grouping`; default `false`). Only affects distributed (multi-rank) runs — a no-op in serial — and reduces filesystem contention when many ranks share a directory. For `per_formulation` under MPI it is applied automatically regardless of this setting, since each rank must write its own file to avoid collisions.
+
+```
+"output": {
+    "root": "/path/to/output/",
+    "catchment": { "enable": true, "format": "csv", "grouping": "per_formulation", "rank_subdir": true },
+    "nexus":     { "enable": true, "format": "csv", "grouping": "per_feature",     "rank_subdir": true }
+}
+```
+
+> [!NOTE]
+> The `output` object supersedes the top-level `output_root`, `disable_catchment_output`, and `per_formulation_nexus_files` keys documented below. Those remain supported for backward compatibility and are honored only when no `output` object is present (with a deprecation warning); new configurations should use `output`.
+
 ### `output_root`
-The configuration may optionally contain an `output_root` key with a user-defined root output directory as the key, for nexus and catchment outputs.
+_Deprecated: prefer `output.root`._ The configuration may optionally contain an `output_root` key with a user-defined root output directory as the key, for nexus and catchment outputs.
 
 ### `disable_catchment_output`
-
-The configuration may optionally contain a `disable_catchment_output` key, with a boolean value.  When set to `true`, catchment output data files will not be written (default: `false`).
+_Deprecated: prefer `output.catchment.enable` set to `false`._ The configuration may optionally contain a `disable_catchment_output` key, with a boolean value.  When set to `true`, catchment output data files will not be written (default: `false`).
 
 ### `per_formulation_nexus_files`
-The configuration may optionally contain a `per_formulation_nexus_files` key with a boolean value to indicate per-formulation, NetCDF files should be used for writing nexus data, rather than the default of per-nexus CSV files.  Note that if `per_formulation_nexus_files` is set to `true`, the `catchments` cannot be used to define formulations for individual catchments, and the global formulation config must be used for all catchments.
+_Deprecated: prefer `output.nexus.format` set to `"netcdf"`._ The configuration may optionally contain a `per_formulation_nexus_files` key with a boolean value to indicate per-formulation, NetCDF files should be used for writing nexus data, rather than the default of per-nexus CSV files.  Note that if `per_formulation_nexus_files` is set to `true`, the `catchments` cannot be used to define formulations for individual catchments, and the global formulation config must be used for all catchments.
 
 > [!IMPORTANT]
 > NetCDF support must be turned on for the ngen build to use this option for per-formulation NetCDF file.  This is done by including the `-DNGEN_WITH_NETCDF=ON` arg to CMake on the command line when generating a build directory.
@@ -51,25 +74,16 @@ Note that these are not exhaustive examples.
    "global": {},
    "time": {},
    "catchments": {},
-   "output_root": "/path/to/output/"
+   "output": { "root": "/path/to/output/" }
 } 
 ```
-or
+or, using the deprecated top-level keys (still supported when no `output` object is present):
 ```
 {
    "global": {},
    "time": {},
    "output_root": "/path/to/output/",
    "per_formulation_nexus_files": true|false
-} 
-```
-or
-```
-{
-   "global": {},
-   "time": {},
-   "output_root": "/path/to/output/",
-   "write_catchment_output": true|false
 } 
 ```
 
--- a/docker/ngen.dockerfile
+++ b/docker/ngen.dockerfile
@@ -33,7 +33,10 @@ RUN cmake -S . \
           -DNGEN_WITH_PYTHON:BOOL=OFF \
           -DNGEN_WITH_TESTS:BOOL=ON \
           -DNGEN_QUIET:BOOL=ON \
-          -DNGEN_WITH_EXTERN_SLOTH:BOOL=ON
+          -DNGEN_WITH_EXTERN_SLOTH:BOOL=ON \
+          `# Rocky 8 ships GCC 8, whose std::filesystem lives in a separate library; link it` \
+          `# explicitly. --start-group avoids ordering issues with the static libstdc++fs.` \
+          -DCMAKE_EXE_LINKER_FLAGS="-Wl,--start-group -lstdc++fs"
 
 RUN cmake --build /ngen_build \
           --target testbmicppmodel ngen \
--- a/docker/ngen.dockerfile
+++ b/docker/ngen.dockerfile
@@ -4,8 +4,16 @@ FROM rockylinux:${ROCKYLINUX_TAG}
 RUN dnf update -y \
     && dnf install -y dnf-plugins-core epel-release \
     && dnf repolist \
-    && dnf install -y --allowerasing tar git gcc-c++ gcc make cmake udunits2-devel coreutils \
+    && dnf install -y --allowerasing tar git gcc-toolset-12 make cmake udunits2-devel coreutils \
     && dnf clean all
+
+# Rocky 8's system compiler is GCC 8, whose libstdc++ lacks complete C++17 <filesystem>
+# support (notably std::hash<std::filesystem::path>). Build with the gcc-toolset-12 SCL
+# toolchain instead, and point the runtime loader at its libstdc++.
+ENV PATH="/opt/rh/gcc-toolset-12/root/usr/bin:${PATH}" \
+    LD_LIBRARY_PATH="/opt/rh/gcc-toolset-12/root/usr/lib64:/opt/rh/gcc-toolset-12/root/usr/lib" \
+    CC="/opt/rh/gcc-toolset-12/root/usr/bin/gcc" \
+    CXX="/opt/rh/gcc-toolset-12/root/usr/bin/g++"
 
 ARG BOOST_VERSION="1.86.0"
 RUN export BOOST_ARCHIVE="boost_$(echo ${BOOST_VERSION} | tr '\.' '_').tar.gz" \
@@ -33,10 +41,7 @@ RUN cmake -S . \
           -DNGEN_WITH_PYTHON:BOOL=OFF \
           -DNGEN_WITH_TESTS:BOOL=ON \
           -DNGEN_QUIET:BOOL=ON \
-          -DNGEN_WITH_EXTERN_SLOTH:BOOL=ON \
-          `# Rocky 8 ships GCC 8, whose std::filesystem lives in a separate library; link it` \
-          `# explicitly. --start-group avoids ordering issues with the static libstdc++fs.` \
-          -DCMAKE_EXE_LINKER_FLAGS="-Wl,--start-group -lstdc++fs"
+          -DNGEN_WITH_EXTERN_SLOTH:BOOL=ON
 
 RUN cmake --build /ngen_build \
           --target testbmicppmodel ngen \
--- a/include/core/DomainLayer.hpp
+++ b/include/core/DomainLayer.hpp
@@ -4,6 +4,10 @@
 #include "Catchment_Formulation.hpp"
 #include "Layer.hpp"
 #include "State_Exception.hpp"
+#include "CatchmentOutputsMgr.hpp"
+
+#include <memory>
+#include <string>
 
 namespace ngen
 {
@@ -34,16 +38,22 @@ namespace ngen
          * @param features collection of HY_Features associated with the domain
          * @param idx index of the layer
          * @param formulation Formulation associated with the domain
+         * @param output_mgr Shared, driver-owned catchment output manager (may be null when output is
+         *        disabled). The layer registers as its own feature (formulation id "layer_<id>",
+         *        feature id = the layer name) in that manager, so its output honors the configured
+         *        format/grouping and lands in its own file alongside the surface catchments.
          */
         DomainLayer(
                 const LayerDescription& desc,
                 const Simulation_Time& s_t,
                 feature_type& features,
                 long idx,
-                std::shared_ptr<realization::Catchment_Formulation> formulation):
-                    Layer(desc, s_t, features, idx), formulation(formulation)
+                std::shared_ptr<realization::Catchment_Formulation> formulation,
+                std::shared_ptr<utils::CatchmentOutputsMgr> output_mgr):
+                    Layer(desc, s_t, features, idx, std::move(output_mgr)), formulation(formulation),
+                    output_formulation_id("layer_" + std::to_string(desc.id)),
+                    output_catchment_id(desc.name)
         {
-            formulation->write_output("Time Step,""Time,"+formulation->get_output_header_line(",")+"\n");
         }
 
         /***
@@ -73,9 +83,12 @@ namespace ngen
                             +" (layer id: "+std::to_string(description.id)+")";
                 throw models::external::State_Exception(msg);
             } 
-            std::string output = std::to_string(output_time_index)+","+current_timestamp+","+
-            formulation->get_output_line_for_timestep(output_time_index)+"\n";
-            formulation->write_output(output);
+            if (catchment_output_mgr) {
+                catchment_output_mgr->receive_data_entry(
+                    output_formulation_id, output_catchment_id,
+                    utils::time_marker(output_time_index, simulation_time.get_current_epoch_time(), current_timestamp),
+                    formulation->get_output_values_for_timestep(output_time_index));
+            }
             ++output_time_index;
             if ( output_time_index < simulation_time.get_total_output_times() )
             {
@@ -86,6 +99,9 @@ namespace ngen
 
         private:
         std::shared_ptr<realization::Catchment_Formulation> formulation;
+        //! This layer's identity in the shared manager, split into its (formulation, catchment) key.
+        std::string output_formulation_id;
+        std::string output_catchment_id;
     };
 }
 
--- a/include/core/HY_Features.hpp
+++ b/include/core/HY_Features.hpp
@@ -3,6 +3,7 @@
 
 #include <unordered_map>
 #include <set>
+#include <memory>
 
 #include <HY_Catchment.hpp>
 #include <HY_HydroNexus.hpp>
@@ -35,9 +36,10 @@ namespace hy_features {
      *     auto r = features.catchment_at(id);
      *     auto r_c = dynamic_pointer_cast<realization::Catchment_Formulation>(r);
      *     double response = r_c->get_response(time_index, 3600.0);
-     *     std::string output = std::to_string(time_index)+", examlpe_time_stamp,"+
-     *                          r_c->get_output_line_for_timestep(time_index)+"\n";
-     *     r_c->write_output(output);
+     *     // outputs for the timestep as a vector<double> of values,
+     *     // in the order the output fields are defined.
+     *     auto fields = r_c->get_output_fields();
+     *     auto values = r_c->get_output_values_for_timestep(time_index);
      * }
      * @endcode
      */
--- a/include/core/HY_Features_MPI.hpp
+++ b/include/core/HY_Features_MPI.hpp
@@ -7,6 +7,7 @@
 
 #include <unordered_map>
 #include <set>
+#include <memory>
 
 #include <HY_Catchment.hpp>
 #include <HY_PointHydroNexusRemote.hpp>
--- a/include/core/Layer.hpp
+++ b/include/core/Layer.hpp
@@ -8,12 +8,15 @@
 #include "State_Exception.hpp"
 #include "geojson/FeatureBuilder.hpp"
 #include <boost/core/span.hpp>
+#include <memory>
 
 namespace hy_features
 {
     class HY_Features;
     class HY_Features_MPI;
 }
+
+namespace utils { class CatchmentOutputsMgr; }
 
 namespace ngen
 {
@@ -32,41 +35,46 @@ namespace ngen
                 const LayerDescription& desc, 
                 const std::vector<std::string>& p_u, 
                 const Simulation_Time& s_t, 
-                feature_type& f, 
-                geojson::GeoJSON cd, 
-                long idx) :
+                feature_type& f,
+                geojson::GeoJSON cd,
+                long idx,
+                std::shared_ptr<utils::CatchmentOutputsMgr> cat_output_mgr) :
             description(desc),
             processing_units(p_u),
             simulation_time(s_t),
             features(f),
             catchment_data(cd),
-            output_time_index(idx)
+            output_time_index(idx),
+            catchment_output_mgr(cat_output_mgr)
         {
 
         }
 
         /**
          * @brief Construct a minimum layer object
-         * 
-         * @param desc 
-         * @param s_t 
-         * @param f 
-         * @param idx 
+         *
+         * @param desc
+         * @param s_t
+         * @param f
+         * @param idx
+         * @param cat_output_mgr Catchment output sink for this layer (may be null when output is disabled)
          */
         Layer(
-                const LayerDescription& desc, 
-                const Simulation_Time& s_t, 
+                const LayerDescription& desc,
+                const Simulation_Time& s_t,
                 feature_type& f,
-                long idx) :
+                long idx,
+                std::shared_ptr<utils::CatchmentOutputsMgr> cat_output_mgr) :
             description(desc),
             simulation_time(s_t),
             features(f),
-            output_time_index(idx)
+            output_time_index(idx),
+            catchment_output_mgr(cat_output_mgr)
         {
 
         }
 
-        virtual ~Layer() {}
+        virtual ~Layer();   // out-of-line; see Layer.cpp
 
         /***
          * @brief Return the next timestep that will be processed by this layer in epoch time units
@@ -119,7 +127,9 @@ namespace ngen
         feature_type& features;
         //TODO is this really required at the top level? or can this be moved to SurfaceLayer?
         const geojson::GeoJSON catchment_data;
-        long output_time_index;       
+        long output_time_index;
+        //! Catchment output sink for this layer's per-timestep push; null when output is disabled.
+        std::shared_ptr<utils::CatchmentOutputsMgr> catchment_output_mgr = nullptr;
 
     };
 }
--- a/include/core/SurfaceLayer.hpp
+++ b/include/core/SurfaceLayer.hpp
@@ -17,8 +17,9 @@ namespace ngen
                 feature_type& f, 
                 geojson::GeoJSON cd, 
                 long idx,
-                const std::shared_ptr<utils::NexusOutputsMgr> &nexus_outputs_mgr) :
-                    Layer(desc,p_u,s_t,f,cd,idx), 
+                const std::shared_ptr<utils::NexusOutputsMgr> &nexus_outputs_mgr,
+                std::shared_ptr<utils::CatchmentOutputsMgr> catchment_output_mgr) :
+                    Layer(desc,p_u,s_t,f,cd,idx,catchment_output_mgr),
                     nexus_outputs_mgr(nexus_outputs_mgr)
         {
 
--- a/include/realizations/catchment/Bmi_Formulation.hpp
+++ b/include/realizations/catchment/Bmi_Formulation.hpp
@@ -6,6 +6,7 @@
 #include <utility>
 #include <vector>
 #include <memory>
+#include <boost/core/span.hpp>
 #include "Catchment_Formulation.hpp"
 #include "GenericDataProvider.hpp"
 
@@ -139,27 +140,6 @@ namespace realization {
         }
 
         /**
-         * Get the values making up the header line from get_output_header_line(), but organized as a vector of strings.
-         *
-         * @return The values making up the header line from get_output_header_line() organized as a vector.
-         */
-        const std::vector<std::string> &get_output_header_fields() const {
-            return output_header_fields;
-        }
-
-        /**
-         * Get a header line appropriate for a file made up of entries from this type's implementation of
-         * ``get_output_line_for_timestep``.
-         *
-         * Note that like the output generating function, this line does not include anything for time step.
-         *
-         * @return An appropriate header line for this type.
-         */
-        std::string get_output_header_line(std::string delimiter) const override {
-            return boost::algorithm::join(get_output_header_fields(), delimiter);
-        }
-
-        /**
          * Get the names of variables in formulation output.
          *
          * Get the names of the variables to include in the output from this formulation, which should be some ordered
@@ -232,6 +212,15 @@ namespace realization {
 
         void set_output_header_fields(const std::vector<std::string> &output_headers) {
             output_header_fields = output_headers;
+        }
+
+        /**
+         * The configured/derived output header (display) field names, positionally parallel to
+         * @ref get_output_variable_names -- header i is the output name for variable i. A concrete
+         * subclass uses these as each column's output_name when building @ref get_output_fields.
+         */
+        boost::span<const std::string> get_output_header_field_names() const {
+            return output_header_fields;
         }
 
         /**
--- a/include/realizations/catchment/Bmi_Module_Formulation.hpp
+++ b/include/realizations/catchment/Bmi_Module_Formulation.hpp
@@ -61,28 +61,46 @@ namespace realization {
         boost::span<const std::string> get_available_variable_names() const override;
 
         /**
-         * Get a delimited string with all the output variable values for the given time step.
+         * Get the output variable values for the given time step, one per output column and positionally aligned
+         * with @ref get_output_fields.
          *
-         * This method is useful for preparing calculated data in a representation useful for output files, such as
-         * CSV files.
+         * The values are the model's calculated outputs for the time step; the time step index itself is not
+         * included.
          *
-         * The resulting string contains only the calculated output values for the time step, and not the time step
-         * index itself.
-         *
-         * An empty string is returned if the time step value is not in the range of valid time steps for which there
-         * are calculated values for all variables.
-         *
-         * The default delimiter is a comma.
-         *
-         * Implementations will throw `invalid_argument` exceptions if data for the provided time step parameter is not
-         * accessible.  Note that, for this type, only the last processed time step is accessible, because formulations
-         * do not save results from previous time steps.  This also has the consequence of there being no valid set of
-         * arguments before a least one call to @ref get_response has been made.
+         * Only the last processed time step is accessible, because formulations do not save results from previous
+         * time steps; there is thus no valid time step before at least one call to @ref update has been made.
+         * Throws `std::invalid_argument` if @p timestep is not the current (last processed) time step.
          *
          * @param timestep The time step for which data is desired.
-         * @return A delimited string with all the output variable values for the given time step.
+         * @return The output values for the given time step, one per output column.
          */
-        std::string get_output_line_for_timestep(int timestep, std::string delimiter) override;
+        std::vector<double> get_output_values_for_timestep(int timestep) override;
+
+        /**
+         * Get this formulation's output fields (header name + units) in output order. Units come
+         * straight from the backing BMI model (@c GetVarUnits) per output variable; std::nullopt when
+         * the model reports none (GetVarUnits fails / BMI_FAILURE) rather than failing the run.
+         */
+        std::vector<utils::OutputField> get_output_fields() const override {
+            // The output variables and their header (output) names are 1:1 and positionally aligned
+            // (enforced at construction) -- variable i is written under header i. Build one field per
+            // variable: units come from the model for that variable; the output name is its header
+            // (which defaults to the variable name, or is a configured alias).
+            const std::vector<std::string> &variables = get_output_variable_names();
+            boost::span<const std::string> headers = get_output_header_field_names();
+            std::vector<utils::OutputField> fields;
+            fields.reserve(variables.size());
+            for (std::size_t i = 0; i < variables.size(); ++i) {
+                std::optional<std::string> units;
+                try {
+                    units = get_bmi_model()->GetVarUnits(variables[i]);
+                } catch (...) {
+                    units = std::nullopt;
+                }
+                fields.emplace_back(variables[i], headers[i], std::move(units));
+            }
+            return fields;
+        }
 
         /**
          * Get the model response for a time step.
@@ -441,7 +459,7 @@ namespace realization {
          * The member serves as an implicit marker of how many time steps have been processed so far.  Knowing this is
          * required to maintain valid behavior in certain things, such as @ref get_response (we may want to process
          * multiple time steps forward to a particular index other than the next, but it would not be valid to receive
-         * a ``t_index`` earlier than the last processed time step) and @ref get_output_line_for_timestep (because
+         * a ``t_index`` earlier than the last processed time step) and @ref get_output_values_for_timestep (because
          * formulations do not save results from previous time steps, only the results from the last processed time step
          * can be used to generate output).
          */
--- a/include/realizations/catchment/Catchment_Formulation.hpp
+++ b/include/realizations/catchment/Catchment_Formulation.hpp
@@ -6,8 +6,7 @@
 #include "Formulation.hpp"
 #include <HY_CatchmentArea.hpp>
 #include "GenericDataProvider.hpp"
-
-#define DEFAULT_FORMULATION_OUTPUT_DELIMITER ","
+#include "utilities/output/CatchmentOutputsMgr.hpp"   // utils::OutputField
 
 namespace realization {
 
@@ -31,36 +30,26 @@ namespace realization {
                                                     const std::string &pattern, const std::string &replacement);
 
             /**
-             * Get a header line appropriate for a file made up of entries from this type's implementation of
-             * ``get_output_line_for_timestep``.
+             * Get this formulation's output fields -- name plus metadata (units, ...) -- in the same
+             * order as the values returned by @ref get_output_values_for_timestep.
              *
-             * Note that like the output generating function, this line does not include anything for time step.
+             * The name is the output header field (configured, or defaulting to the output variable
+             * name); the units come from the same source that produces each value. So each field is
+             * positionally aligned with, and describes, the corresponding value.
              *
-             * A default implementation is provided for inheritors of this type, which includes only "Total Discharge."
-             *
-             * @return An appropriate header line for this type.
+             * @return The ordered output fields.
              */
-            virtual std::string get_output_header_line(std::string delimiter=DEFAULT_FORMULATION_OUTPUT_DELIMITER) const;
+            virtual std::vector<utils::OutputField> get_output_fields() const = 0;
 
             /**
-             * Get a formatted line of output values for the given time step as a delimited string.
+             * Get the output values for the given time step, positionally aligned with @ref get_output_fields.
              *
-             * This method is useful for preparing calculated data in a representation useful for output files, such as
-             * CSV files.
+             * Values are returned as raw doubles, with no formatting or precision applied by the formulation.
              *
-             * The resulting string will contain calculated values for applicable output variables for the particular
-             * formulation, as determined for the given time step.  However, the string will not contain any
-             * representation of the time step itself.
-             *
-             * An empty string is returned if the time step value is not in the range of valid time steps for which there
-             * are calculated values for all variables.
-             *
-             * @param timestep The time step for which data is desired.
-             * @param delimiter The value delimiter for the string.
-             * @return A delimited string with all the output variable values for the given time step.
+             * @param timestep The time step for which values are desired.
+             * @return The output values for @p timestep, one per output column.
              */
-            virtual std::string get_output_line_for_timestep(int timestep,
-                                                             std::string delimiter = DEFAULT_FORMULATION_OUTPUT_DELIMITER) = 0;
+            virtual std::vector<double> get_output_values_for_timestep(int timestep) = 0;
 
             /**
              * Execute the backing model formulation for the given time step, where it is of the specified size, and
--- a/include/realizations/catchment/Formulation_Manager.hpp
+++ b/include/realizations/catchment/Formulation_Manager.hpp
@@ -14,6 +14,7 @@
 #include <sys/types.h>
 #include <unistd.h>
 #include <string>
+#include <iostream>
 
 #include <boost/property_tree/ptree.hpp>
 #include <boost/property_tree/json_parser.hpp>
@@ -25,6 +26,7 @@
 #include "realizations/config/routing.hpp"
 #include "realizations/config/config.hpp"
 #include "realizations/config/layer.hpp"
+#include "realizations/config/output.hpp"
 
 namespace realization {
 
@@ -40,10 +42,12 @@ namespace realization {
                 boost::property_tree::ptree loaded_tree;
                 boost::property_tree::json_parser::read_json(file_path, loaded_tree);
                 this->tree = loaded_tree;
+                initialize_output_config();
             }
 
             Formulation_Manager(boost::property_tree::ptree &loaded_tree) {
                 this->tree = loaded_tree;
+                initialize_output_config();
             }
 
             ~Formulation_Manager() = default;
@@ -54,18 +58,8 @@ namespace realization {
                 //and routing and other non feature specific tasks from this main function
                 //which has to iterate the entire hydrofabric.
 
-                // TODO: (later) consider whether this really belongs inside the global formulation
-                auto per_formulation_setting = tree.get_child_optional("per_formulation_nexus_files");
-                if (per_formulation_setting) {
-                    #if !NGEN_WITH_NETCDF
-                    throw std::runtime_error("ERROR: per_formulation_nexus_files is set to true, but NGEN was built without NetCDF support.");
-                    #else
-                    use_per_formulation_nexus_files = per_formulation_setting->get_value<bool>();
-                    #endif
-                }
-                else {
-                    use_per_formulation_nexus_files = false;
-                }
+                // Output configuration (output_config) is parsed at construction
+                // so get_output_config() is valid before read().
 
                 auto possible_global_config = tree.get_child_optional("global");
 
@@ -104,7 +98,6 @@ namespace realization {
                                 output_stream
                                 )
                             );
-                            domain_formulations.at(layer_desc.id)->set_output_stream(get_output_root() + layer_desc.name + "_layer_"+std::to_string(layer_desc.id) + ".csv");
                         }
                         //TODO for each layer, create deferred providers for use by other layers
                         //VERY SIMILAR TO NESTED MODULE INIT
@@ -137,7 +130,8 @@ namespace realization {
                 auto possible_catchment_configs = tree.get_child_optional("catchments");
 
                 // For now at least, this isn't allowed
-                if (possible_catchment_configs && use_per_formulation_nexus_files) {
+                if (possible_catchment_configs &&
+                    output_config.nexus.format == realization::config::OutputFormat::netcdf) {
                     std::string msg = "ERROR: Individual catchment formulation configs are not allowed when using "
                                       "per-formulation nexus files.";
                     std::cerr << msg;
@@ -223,8 +217,66 @@ namespace realization {
                 return this->formulations.empty();
             }
 
+            /**
+             * @brief Get the formatted output root: check the existence of the output_root directory defined
+             * in realization. If true, return the directory name. Otherwise, try to create the directory
+             * or throw an error on failure.
+             *
+             * @return std::string of the output root directory
+             */
+            std::string get_output_root() const {
+                const auto output_root = this->tree.get_optional<std::string>("output_root");
+                if (output_root != boost::none && *output_root != "") {
+                    // Check if the path ends with a trailing slash, otherwise add it.
+                    std::string str = output_root->back() == '/'
+                           ? *output_root
+                           : *output_root + "/";
+
+                    const char* dir = str.c_str();
+
+                    //use C++ system function to check if there is a dir match that defined in realization
+                    struct stat sb;
+                    if (stat(dir, &sb) == 0 && S_ISDIR(sb.st_mode)) {
+                        return str;
+                    } else {
+                        errno = 0;
+                        int result = mkdir(dir, 0755);
+                        if (result == 0)
+                            return str;
+                        // Another process/MPI rank may have created the directory between this rank's stat and
+                        // mkdir calls, so consider EEXIST as success as long as the path is a directory.
+                        if (errno == EEXIST && stat(dir, &sb) == 0 && S_ISDIR(sb.st_mode))
+                            return str;
+                        throw std::runtime_error("failed to create directory '" + str + "': " + std::strerror(errno));
+                    }
+                }
+
+                //for case where there is no output_root in the realization file
+                return "./";
+            }
+
+            /**
+             * Check if the formulation has catchment output writing disabled.
+             *
+             * @return bool
+             */
+            bool is_disable_catchment_output() const {
+                return tree.get_optional<bool>("disable_catchment_output").get_value_or(false);
+            }
+
+            /**
+             * Whether nexus output is written as per-formulation NetCDF files (the deprecated
+             * per_formulation_nexus_files mode, now expressed as nexus format == netcdf).
+             *
+             * @return bool
+             */
             bool is_using_per_formulation_nexus_files() const {
-                return use_per_formulation_nexus_files;
+                return output_config.nexus.format == realization::config::OutputFormat::netcdf;
+            }
+
+            //! The parsed output configuration block.
+            const realization::config::Output& get_output_config() const {
+                return output_config;
             }
 
             typename std::map<std::string, std::shared_ptr<Catchment_Formulation>>::const_iterator begin() const {
@@ -283,72 +335,25 @@ namespace realization {
             }
 
             /**
-             * @brief Get the formatted output root: check the existence of the output_root directory defined
-             * in realization. If true, return the directory name. Otherwise, try to create the directory
-             * or throw an error on failure.
-             *
-             * @code{.cpp}
-             * // Example config:
-             * // ...
-             * // "output_root": "/path/to/dir/"
-             * // ...
-             * const auto manager = Formulation_Manger(CONFIG);
-             * manager.get_output_root();
-             * //> "/path/to/dir/"
-             * @endcode
-             * 
-             * @return std::string of the output root directory
+             * Parse the output configuration from the realization tree (preferring
+             * the "output" block, falling back to the deprecated top-level keys),
+             * warn on deprecated usage, and validate build support. Called once at
+             * construction so output accessors are valid before read().
              */
-            std::string get_output_root() const {
-                const auto output_root = this->tree.get_optional<std::string>("output_root");
-                if (output_root != boost::none && *output_root != "") {
-                    // Check if the path ends with a trailing slash,
-                    // otherwise add it.
-                    std::string str = output_root->back() == '/'
-                           ? *output_root
-                           : *output_root + "/";
-
-                    const char* dir = str.c_str();
-
-                    //use C++ system function to check if there is a dir match that defined in realization
-                    struct stat sb;
-                    if (stat(dir, &sb) == 0 && S_ISDIR(sb.st_mode)) {
-                        return str;
-                    } else {
-                        errno = 0;
-                        int result = mkdir(dir, 0755);      
-                        if (result == 0)
-                            return str;
-                        // Another process/MPI rank may have created the directory between this rank's stat and
-                        // mkdir calls, so consider EEXIST as success as long as the path is a directory.
-                        if (errno == EEXIST && stat(dir, &sb) == 0 && S_ISDIR(sb.st_mode))
-                            return str;
-                        throw std::runtime_error("failed to create directory '" + str + "': " + std::strerror(errno));
-                    }
+            void initialize_output_config() {
+                output_config = realization::config::Output::from_realization(tree);
+                #if !NGEN_QUIET
+                if (output_config.from_legacy_keys) {
+                    std::cerr << "WARNING: top-level 'output_root', 'disable_catchment_output', and "
+                                 "'per_formulation_nexus_files' keys are deprecated; prefer an 'output' "
+                                 "config block (output.root / output.catchment / output.nexus)." << std::endl;
                 }
- 
-                //for case where there is no output_root in the realization file
-                return "./";
-
-            }
-
-             /**
-             * Check if the formulation has catchment output writing disabled.
-             *
-             * @code{.cpp}
-             * // Example config:
-             * // ...
-             * // "disable_catchment_output": true
-             * // ...
-             * const auto manager = Formulation_Manger(CONFIG);
-             * manager.is_disable_catchment_output();
-             * //> true
-             * @endcode
-             * 
-             * @return bool
-             */
-            bool is_disable_catchment_output() const {
-                return tree.get_optional<bool>("disable_catchment_output").get_value_or(false);
+                #endif
+                #if !NGEN_WITH_NETCDF
+                if (output_config.nexus.format == realization::config::OutputFormat::netcdf) {
+                    throw std::runtime_error("ERROR: nexus output format 'netcdf' requested, but NGEN was built without NetCDF support.");
+                }
+                #endif
             }
 
             /**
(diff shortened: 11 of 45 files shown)


include/core/DomainLayer.hpp
```diff
@@ -4,6 +4,11 @@
 #include "Catchment_Formulation.hpp"
 #include "Layer.hpp"
 #include "State_Exception.hpp"
+#include "CatchmentOutputsMgr.hpp"
+#include "CatchmentCsvOutputMgr.hpp"
```
christophertubbs: Can't tell if this would count as out of scope or not since it may revolve around a collective planning and design discussion, but embedding "We're going to use CSV output" here seems like a flaw in abstraction. To _me_ (I don't want to speak for _everyone_ here), I'd like to see a generic, _base_ output manager used with the actual manager produced via factory or something so we could follow some of the same interfaces if we want to pump outputs to csv, netcdf, parquet, etc. The `DomainLayer` wouldn't really need to know and it'd allow us and our users to expand in different, sometimes wacky, ways, right? I know its use below is just in a default value, but I'd really like to see the decision on what exactly the default _is_ defined elsewhere. Even if the externalized factory or whatever just produces the single type of value (the csv output manager), it would leave us an opening to produce other options later down the line.

include/utilities/output/CatchmentOutputsMgr.hpp
```diff
@@ -0,0 +1,136 @@
...
+// For utils::time_marker, shared with the nexus managers. (A later refactor can
+// move time_marker to a neutral header when the two hierarchies are unified.)
+#include "NexusOutputsMgr.hpp"
+
+namespace utils
+{
+    /**
+     * Abstract class for managing and writing to catchment data files.
+     */
+    class CatchmentOutputsMgr
```
christophertubbs, line 47: This interface seems _mildly_ suited specifically for CSV as it leans into the possibly jagged column definition/column value interface. That may not work well for netcdf or something wacky down the line.

How would this accommodate other formats? Is it even worth really worrying about at this time? I can see us getting into a situation of having a really CSV focused interface, needing to add features and ideas for other formats and having to change all integration points.

src/utilities/output/CatchmentCsvOutputMgr.cpp
```diff
@@ -0,0 +1,200 @@
...
+CatchmentCsvOutputMgr::~CatchmentCsvOutputMgr()
+{
+    close();
+}
+
+std::string CatchmentCsvOutputMgr::output_path(const std::string &formulation_id, const std::string &catchment_id)
+{
+    // The default formulation keeps the long-standing flat layout; any other formulation gets its
+    // own subdirectory so independent setups over the same catchment don't overwrite each other.
+    if (formulation_id == get_default_formulation_id()) {
```
christophertubbs: Does this need to account for `output_root_` ending without a `"/"` or is that handled elsewhere?

src/utilities/output/CatchmentCsvOutputMgr.cpp
```diff
@@ -0,0 +1,200 @@
...
+        throw std::runtime_error("Can't receive data on a closed CatchmentCsvOutputMgr");
+    }
+
+    std::ofstream* out = stream_for(formulation_id, catchment_id);
+    if (out == nullptr) {
+        throw std::runtime_error("CatchmentCsvOutputMgr received data for unregistered catchment '" + catchment_id + "'");
+    }
+
+    // In aggregated output each row carries the (full) catchment id so rows stay attributable.
+    (*out) << catchment_output_data_row(aggregate_, catchment_id, data_time_marker.sim_time_index,
```
christophertubbs: Should this instead be buffered in order to reduce writes? Maybe add the generated row to a collection and once a threshold is reached, write, possibly flush, then make sure everything is written at the time of `commit_writes`?

include/utilities/output/CatchmentOutputsMgr.hpp
```diff
@@ -0,0 +1,136 @@
...
+        /**
+         * A test of whether this instance is closed.
+         *
+         * @return Whether this instance is closed.
+         * @see close
+         */
+        virtual bool is_closed() = 0;
+
+    protected:
+        virtual ~CatchmentOutputsMgr() = default;
```
christophertubbs, line 122: Should this be set up to call `close` on destruction?

docker/ngen.dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG ROCKYLINUX_TAG=8
+ARG ROCKYLINUX_TAG=9
```
PhilMiller: Regarding the CI change, is there a devset or whatever it's called (or just a package) for GCC-newer-than-8 that we can install on Rocky Linux 8, to get the newer compiler version without also making this bigger shift as a side-effect?

Or, down below on CMake configuration, we could pass `-DCMAKE_EXE_LINKER_FLAGS=-lstdc++fs`

Or, maybe preserving the support for baseline standard Rocky 8 in the codebase is reason enough to add the compiler-version-conditional linkage in `CMakeLists.txt` that we talked about

include/utilities/output/CatchmentOutputsMgr.hpp
```diff
@@ -0,0 +1,136 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+------------------------------------------------------------------------
+*/
+
+// Catchment output manager, modeled on NexusOutputsMgr: data is pushed in via
+// receive_data_entry and the concrete backend (CSV now; NetCDF/parquet later)
```
PhilMiller: The comments here in the interface header shouldn't really be naming the implementations

include/utilities/output/CatchmentOutputsMgr.hpp
```diff
@@ -0,0 +1,136 @@
...
+         *
+         * @param formulation_id The id of the formulation producing this catchment's output.
+         * @param catchment_id The id for the catchment (e.g. "cat-42").
+         * @param columns The output column names for this catchment, in the order values will be
+         *        supplied to @ref receive_data_entry (a formulation's output header fields). The
+         *        backend decides how to render them (CSV header, NetCDF variable definitions, ...).
+         */
+        virtual void register_catchment(const std::string &formulation_id,
+                                        const std::string &catchment_id,
+                                        const std::vector<std::string> &columns) = 0;
```
PhilMiller: If we're doing this, I would *really* like to have the units of each column specified from the beginning

include/utilities/output/CatchmentOutputsMgr.hpp
hellkite500: This interface closely follows the NexusOutputsMgr interface at this time (which has both netcdf and csv backends), and *should* be reasonably flexible for most foreseeable needs.  The "fields" vs "values" separation allows something like NetCDF to use the fields to create vars, and the values to push into the arrays.

src/utilities/output/CatchmentCsvOutputMgr.cpp
hellkite500: That is handled by the output configuration parsing (output.hpp), by resolving and providing a canonical path through the output struct's root member
```cpp
//! Normalize a configured directory to a trailing-slash path and create it,
//! falling back to "./" when unset. Applied by from_realization so the resolved
//! root is materialized once there rather than at each point of use.
static std::string resolve_root(const std::string& configured_root)
{
    if (configured_root.empty()) {
        return "./";
    }
    std::string dir = configured_root.back() == '/' ? configured_root : configured_root + "/";
    std::error_code ec;
    std::filesystem::create_directories(dir, ec);
    if (ec) {
        throw std::runtime_error("failed to create output directory '" + dir + "': " + ec.message());
    }
    return dir;
}
```
    
    called from this function (same file)
    
```cpp
static Output from_realization(const boost::property_tree::ptree& realization_tree)
{
    auto block = realization_tree.get_child_optional(OUTPUT_CONFIG_KEY);
    Output out = block ? parse_block(*block) : parse_legacy(realization_tree);
    out.root = resolve_root(out.root);
    return out;
}
```

src/utilities/output/CatchmentCsvOutputMgr.cpp
hellkite500: The `ofstream` is buffered by default, the buffer *can* be tuned if needed at a later point.

docker/ngen.dockerfile
hellkite500: I'm not convinced on any of the options haha...I went with a major rocky bump mostly for simplicity and getting the CI image to run for now.  Rocky 8 is EoL in 2029, so ensuring compatibility with it for the next couple years may be worth the addition of inspection and linkage in the project's CMakeLists.  The other thing we could consider is bump the minimum compiler versions we claim to support from gcc 8 to 9.

A simple project CMake option is just always linking `-lstdc++fs` for now, which fixes the gcc 8 support and doesn't impact newer compilers.

include/utilities/output/CatchmentOutputsMgr.hpp
christophertubbs: How would this interface handle variable and field output metadata? The prod version of streamflow right now is:
```
int streamflow(feature_id=2776734);
  :long_name = "River Flow";
  :units = "m3 s-1";
  :coordinates = "latitude longitude";
  :grid_mapping = "crs";
```
  
 How would this interface need to grow to support this?

include/utilities/output/CatchmentOutputsMgr.hpp
```diff
@@ -0,0 +1,136 @@
...
+     */
+    class CatchmentOutputsMgr
+    {
+
+    public:
+        /**
+         * Declare a catchment and its output schema before any data is received for it.
+         *
+         * A backend uses this to prepare its sink (e.g. select/open the CSV file and write the
+         * header, or define NetCDF dimensions/variables). Call once per (formulation, catchment).
```
PhilMiller: The NetCDF case makes me think that this design isn't really fit for purpose. A NetCDF writer notionally needs to know what catchments and formulations it's aggregating for up front. There's nothing in the interface to say that calls to `register_catchment` are done, and calls to `receive_data_entry` will start.

I think the thing to do may be to drop `register_catchment` from the common interface, and instead have that functionality be more instance-specific. Either in a constructor, or stuff that the code calling the implementation-specific constructor knows how to do after construction.

include/utilities/output/CatchmentOutputsMgr.hpp
PhilMiller: Not in the base class, no, I don't think.

src/utilities/output/CatchmentCsvOutputMgr.cpp
PhilMiller: Depends if `output_root_` is specified to be a directory or is potentially has a filename prefix at the end. The constructor above isn't documented one way or the other.

In the directory case, we can safely add an extra `'/'` whether it's there or not, and should maybe also try creating it.

In the filename prefix case, tacking on `'/'` would be a mistake

@hellkite500 which is intended?

PhilMiller: (oops, sorry, Nels's comment hadn't loaded for me)

include/utilities/output/CatchmentOutputsMgr.hpp
```diff
@@ -0,0 +1,136 @@
...
+         */
+        virtual void close() = 0;
+
+        /**
+         * A test of whether this instance is closed.
+         *
+         * @return Whether this instance is closed.
+         * @see close
+         */
+        virtual bool is_closed() = 0;
```
PhilMiller, line 120: Do users of this interface actually need this? It only seems useful to implementations

test/utilities/output/CatchmentCsvOutputMgr_Test.cpp
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// Per-feature: one file per catchment, each with its own header; typed values
+// are joined in column order at the manager's configured precision.
+TEST(CatchmentCsvOutputMgr_Test, PerFeatureOneFilePerCatchment)
+{
+    const auto dir = std::filesystem::temp_directory_path();
+    const std::string root = (dir / "").string();
+    std::filesystem::remove(dir / "cat-1.csv");
+    std::filesystem::remove(dir / "cat-2.csv");
```
PhilMiller: The path pattern here could use some work, to allow for the tests running in parallel, and multiple instances running concurrently as well.

I'd recommend making a directory under `temp_directory_path()` named with `getpid()`, and then a directory inside that named for the individual test case with [`name()`](https://google.github.io/googletest/reference/testing.html#TestInfo::name)

Each test can clean up by calling `remove_all` on the test-specific directory, and then `remove` on the PID-specific directory, such that the last actually removes it.

include/core/DomainLayer.hpp
```diff
@@ -40,10 +45,22 @@ namespace ngen
...
+                    output_formulation_id("layer_" + std::to_string(desc.id)),
+                    output_catchment_id(desc.name)
         {
-            formulation->write_output("Time Step,""Time,"+formulation->get_output_header_line(",")+"\n");
+            // Domain output is one CSV per domain layer. The layer is its own formulation
+            // (id "layer_<id>") over the domain (catchment id = the layer name), so the manager
+            // writes it to "<output_root>layer_<id>/<name>.csv"; register writes the header.
+            // TODO: the backend is a plain per-feature CSV manager constructed directly here.
+            // Better configuration of the domain layer and output manager is possible, but for
+            // now this is the simplest approach to get domain-layer output written.
```
PhilMiller: This whole comment really applies up above, where `output_mgr` is initialized, and not.

The first half of the comment describes implementation details of the CsvOutputMgr, and should probably just be deleted.

include/core/HY_Features.hpp
```diff
@@ -115,6 +116,7 @@ namespace hy_features {
           return nullptr;
         }
 
+
```
PhilMiller: Stray added new line

include/core/HY_Features_MPI.hpp
```diff
@@ -70,6 +70,7 @@ namespace hy_features {
                 boost::adaptors::filtered([this](std::string const& id){ return !is_remote_sender_nexus(id);});
         }
 
+
```
PhilMiller: Stray added line, without which this file would be unmodified

include/core/Layer.hpp
```diff
@@ -32,15 +35,17 @@ namespace ngen
                 const LayerDescription& desc, 
                 const std::vector<std::string>& p_u, 
                 const Simulation_Time& s_t, 
-                feature_type& f, 
-                geojson::GeoJSON cd, 
-                long idx) :
+                feature_type& f,
+                geojson::GeoJSON cd,
+                long idx,
+                std::shared_ptr<utils::CatchmentOutputsMgr> cat_output_mgr = nullptr) :
```
PhilMiller: Do we really need the default value for `cat_output_mgr`? There's only one or two places constructing `Layer`. They should be explicit about what to do with output.

include/core/SurfaceLayer.hpp
```diff
@@ -17,8 +17,9 @@ namespace ngen
                 feature_type& f, 
                 geojson::GeoJSON cd, 
                 long idx,
-                const std::shared_ptr<utils::NexusOutputsMgr> &nexus_outputs_mgr) :
-                    Layer(desc,p_u,s_t,f,cd,idx), 
+                const std::shared_ptr<utils::NexusOutputsMgr> &nexus_outputs_mgr,
+                std::shared_ptr<utils::CatchmentOutputsMgr> catchment_output_mgr = nullptr) :
```
PhilMiller: Ditto on default value

include/realizations/catchment/Catchment_Formulation.hpp
```diff
@@ -31,36 +29,25 @@ namespace realization {
...
-             * CSV files.
-             *
-             * The resulting string will contain calculated values for applicable output variables for the particular
-             * formulation, as determined for the given time step.  However, the string will not contain any
-             * representation of the time step itself.
+             * Get the output values for the given time step, positionally aligned with @ref get_output_header_fields.
              *
-             * An empty string is returned if the time step value is not in the range of valid time steps for which there
-             * are calculated values for all variables.
+             * Values are returned as raw doubles, with no formatting or precision applied by the formulation.
```
PhilMiller, line 47: This sentence is kinda vacuously repeating the return type

src/core/HY_Features.cpp
```diff
@@ -2,6 +2,8 @@
 #include <HY_PointHydroNexus.hpp>
 #include <Formulation_Manager.hpp>
 
+#include <memory>
```
PhilMiller: Probably not actually necessary to add? If this is failing to compile without it, is it because it was previously missing, or because something else in a header added code depending on (e.g.) `shared_ptr` and didn't add this `#include`?

include/utilities/output/CatchmentOutputsMgr.hpp
hellkite500: I think this actually connects with Phil's request to carry units along with the field names.  Making the fields a struct and carrying additional attributes with them would allow a formulation to expose meta data (either queried from a backing model or connected via configuration) and the output manager can decide on what to do with those attributes.  I'll propose a change for this in a standalone commit we can look at and discuss.

src/core/HY_Features.cpp
PhilMiller: OK, I just checked - `HY_Features.hpp` uses `shared_ptr`, and doesn't `#include <memory>`. Please fix it there instead.

include/core/HY_Features.hpp
PhilMiller, line 5: Add `#include <memory>` for use of `shared_ptr`

include/core/HY_Features_MPI.hpp
PhilMiller, line 9: Also missing `#include <memory>`

src/core/HY_Features_MPI.cpp
```diff
@@ -2,6 +2,8 @@
 #include <HY_PointHydroNexusRemote.hpp>
 #include <Formulation_Manager.hpp>
 
+#include <memory>
```
PhilMiller: Ditto on doing this correctly in the header

src/realizations/catchment/Bmi_Module_Formulation.cpp
```diff
@@ -16,18 +16,21 @@ namespace realization {
...
-
-            for (const std::string& name : get_output_variable_names()) {
-                output_str += (output_str.empty() ? "" : ",") + std::to_string(get_var_value_as_double(0, name));
+            const std::vector<std::string>& names = get_output_variable_names();
+            std::vector<double> values;
+            values.reserve(names.size());
+            // ensure output values are positionally aligned
+            // with the field names
+            for (const std::string& name : names) {
+                values.push_back(get_var_value_as_double(0, name));
```
PhilMiller: I'm going to keep harping on the need for calling `get_value` with units specified, and doing away with `get_var_value_as_double`. I have that change extracted from the NGWPC work in an active PR.

include/utilities/output/CatchmentOutputsMgr.hpp
hellkite500: I think there's a nice way of carrying units through to the output and allowing configurable unit representation with auto conversion on the outputs.  I'll definitely work on carrying units through alongside the fields now, and defer the configuration/conversion to a future PR (should be pretty straight forward once this gets ironed out...)

src/NGen.cpp
```diff
@@ -624,6 +644,13 @@ int main(int argc, char* argv[]) {
     // Close nexus output file(s)
     nexus_outputs_mgr->close();
 
+    // Commit and close catchment output file(s). Output is committed once here at end-of-run
+    // rather than per-timestep, so writes buffer normally.
+    if (catchment_outputs_mgr) {
+        catchment_outputs_mgr->commit_writes();
+        catchment_outputs_mgr->close();
```
PhilMiller, line 687: The manager should be specified such that `close()` implies flushing and committing all of the data to be written. There should be no need for callers to call `commit_writes()` at the end of a run, as opposed to mid-run if they want to ensure durability.

test/realizations/catchments/Bmi_Fortran_Formulation_Test.cpp
```diff
@@ -340,9 +337,9 @@ TEST_F(Bmi_Fortran_Formulation_Test, GetOutputLineForTimestep_1_b) {
     while (i < 542)
         formulation.get_response(i++, 3600);
     double response = formulation.get_response(i, 3600);
-    std::string output = formulation.get_output_line_for_timestep(i, ",");
+    std::vector<double> output = formulation.get_output_values_for_timestep(i);
     //NOTE these answers are dependent on the INPUT vars selected and the data in the forcing file
-    ASSERT_EQ(output, "0.025400,0.000001,0.000000");
+    EXPECT_THAT(output, ::testing::Pointwise(::testing::DoubleNear(1e-15), std::vector<double>{0.025399999693036079, 1.1124674593096233e-06, 0.0}));
```
PhilMiller, line 343: Is this testing to too-fine precision? Should we actually match the precision of the previous formatted text, and test against that?

include/utilities/output/CatchmentOutputsMgr.hpp
hellkite500: I think this is the direction to go.  A `registration` method was originally required when trying to keep the existing pattern of creating output during the HY_Features initialization when we were initializing streams on demand.  This was the way of achieving that.  I later moved the entire construction of output out of those classes and into main, at which point the entire feature set is known and incremental registration isn't really needed.

I'll refactor this and test it, then push a review commit for this specific change so we can see what it looks like.

test/realizations/Formulation_Manager_Test.cpp
```diff
@@ -1206,21 +1204,14 @@ TEST_F(Formulation_Manager_Test, read_external_attributes) {
      * - Asserts that the expected values are contained within the output line at
```
PhilMiller: Output vector, not a line anymore

include/utilities/output/CatchmentCsvOutputMgr.hpp
```diff
@@ -29,14 +29,18 @@ limitations under the License.
...
+     *   - per-feature: one file per (formulation, catchment). For the default formulation id the file
+     *     is "<output_root><catchment_id>.csv" (the long-standing layout); any other formulation id is
+     *     placed in a per-formulation subdirectory, "<output_root><formulation_id>/<catchment_id>.csv",
+     *     so independent formulation setups over the same catchment never collide.
+     *   - aggregated: all catchments share one file (the aggregated filename) with a leading
+     *     catchment_id column; the header is written once. Aggregated output assumes a single
+     *     formulation (mirroring PerFormulationNexusOutputMgr) and does not split by formulation id.
+     *
+     * Rows are written immediately on receipt (@ref commit_writes only flushes), mirroring the prior
+     * behavior where each formulation wrote straight to its own stream. Values are rendered to text at
```
PhilMiller: We don't need to document historical code that's being deleted, a la "mirroring the prior behavior".

src/utilities/output/CatchmentCsvOutputMgr.cpp
```diff
@@ -49,26 +49,42 @@ namespace {
...
+        if (include_id) line += "catchment_id" + delimiter;
+        line += "Time Step" + delimiter + "Time" + delimiter + variable_header;
+        return line;
     }
 
     //! Build a CSV data row matching the column order of catchment_output_header_line.
-    std::string catchment_output_data_row(long time_step,
+    //! When include_id is true the row is prefixed with the (full) catchment id.
+    std::string catchment_output_data_row(bool include_id,
+                                           const std::string &catchment_id,
```
PhilMiller: This could be `std::optional<std::string const&> catchment_id` without the separate `bool`

src/utilities/output/CatchmentCsvOutputMgr.cpp
```diff
@@ -87,22 +103,48 @@ std::string CatchmentCsvOutputMgr::output_path(const std::string &formulation_id
...
+{
+    if (aggregate_) {
+        return aggregated_stream_.get();
+    }
+    auto it = per_feature_streams_.find(output_path(formulation_id, catchment_id));
+    return it == per_feature_streams_.end() ? nullptr : it->second.get();
+}
+
 void CatchmentCsvOutputMgr::register_catchment(const std::string &formulation_id, const std::string &catchment_id,
                                                const std::vector<std::string> &columns)
```
PhilMiller: The `columns` parameter no longer really makes sense in the aggregated case, since different formulations could then pass in different arguments, and only the first caller's would be meaningful.

src/utilities/output/CatchmentCsvOutputMgr.cpp
```diff
@@ -87,22 +103,48 @@ std::string CatchmentCsvOutputMgr::output_path(const std::string &formulation_id
...
+    if (aggregate_) {
+        // All catchments share one stream, so the file is opened and the header (with the leading
+        // catchment_id column) is written once, on the first registration.
+        if (!aggregated_stream_) {
+            aggregated_stream_ = std::make_shared<std::ofstream>();
+            aggregated_stream_->open(output_root_ + aggregated_filename_, std::ios::trunc);
+        }
+        if (!aggregated_header_written_) {
+            (*aggregated_stream_) << catchment_output_header_line(true, variable_header) << "\n";
+            aggregated_header_written_ = true;
```
PhilMiller: Is there a circumstance in which we'd construct and open the `aggregated_stream_`, but not immediately write the header row? Can the separate flag variable and condition just be folded into the former block?

include/realizations/config/output.hpp
```diff
@@ -0,0 +1,157 @@
...
+        //! falling back to "./" when unset. Applied by from_realization so the resolved
+        //! root is materialized once there rather than at each point of use.
+        static std::string resolve_root(const std::string& configured_root)
+        {
+            if (configured_root.empty()) {
+                return "./";
+            }
+            std::string dir = configured_root.back() == '/' ? configured_root : configured_root + "/";
+            std::error_code ec;
+            std::filesystem::create_directories(dir, ec);
```
PhilMiller: I'm a little uncomfortable with this functionality living inside the `config` space, and especially in the `config::Output` struct. Could this instead be a free function living near where `struct Output` is used?

PhilMiller: The empty -> `"./"` and trailing `'/'` functionality should stay in this class. It's just the `create_directories` call that bugs me.

docker/ngen.dockerfile
PhilMiller: I've been reading some more about it, and it looks like that would maybe get messy for clang users.

As noted in the CMake issue tracker https://gitlab.kitware.com/cmake/cmake/-/work_items/17834 there's no standard solution to this issue, and many project-specific partial solutions.

That's why I'm inclined toward just making a fix to our CI for now.

include/realizations/config/output.hpp
hellkite500: I considered this approach as well.  I ended up making this a private helper and only called through the `from_realization` method in an attempt to ensure that once a realization is parsed, the output root is canonical and guaranteed to exist.  The "near where struct Output" is used falls out of calling the `from_realization` method at the moment.  It could be changed, but my only concern is ensuring that the root is consistent and verified in all paths that depend on it.

docker/ngen.dockerfile
PhilMiller: So, the `CMAKE_EXE_LINKER_FLAGS` suggestion needs a bit of a tweak, but it does work:

> `-DCMAKE_EXE_LINKER_FLAGS="-Wl,--start-group -lstdc++fs"`

Added to the configure command around line 25 below works on Rocky 8.

It's even something we could document as a workaround.

For old Clang (or something else using old libc++, really), the library would be `-lc++fs`

include/realizations/catchment/Formulation_Manager.hpp
PhilMiller, line 38: Missing `initialize_output_config()` here?

PhilMiller: (this overlooked parallel change illustrates why I would like to eliminate the `stringstring` and `file_path` constructors, and make parsing to `ptree` the caller's responsibility)

test/realizations/Formulation_Manager_Test.cpp
```diff
@@ -1204,8 +1327,9 @@ TEST_F(Formulation_Manager_Test, read_external_attributes) {
...
      *   timestep `ts`.
      *
-     * @note The expected values are checked against the formulation's typed
-     *       output values for the timestep via `std::find`.
+     * @note The output line is checked by splitting it along its delimiter,
+     *       and parsing the resulting strings to doubles. Then, `std::find`
+     *       is used to check for the existence of the expected doubles.
```
PhilMiller: It looks like this may have been inadvertently reverted in a rebase conflict resolution

include/utilities/output/PerNexusCsvOutputMgr.hpp
```diff
@@ -37,8 +37,12 @@ namespace utils
          *
          * @param nexus_ids Nexus ids for which this instance manages data (in particular, local nexuses when using MPI).
          * @param output_root The output root for written files (as a string).
+         * @param subdirectory Optional subdirectory of @p output_root to place files in (e.g. "rank_3" to
+         *        group an MPI rank's files together). When non-empty it is created if it does not exist.
+         *        @p output_root itself is assumed to already exist.
          */
-        PerNexusCsvOutputMgr(const std::vector<std::string>& nexus_ids, const std::string &output_root);
+        PerNexusCsvOutputMgr(const std::vector<std::string>& nexus_ids, const std::string &output_root,
+                             const std::string &subdirectory = "");
```
PhilMiller: I'd prefer a slightly different interface here. Either of:
* the caller adds `subdirectory` to the passed `output_root` argument, or
* `std::optional<std::string const&> subdirectory = nullopt`

src/NGen.cpp
```diff
@@ -565,7 +565,18 @@ int main(int argc, char* argv[]) {
...
     }
     else {
-        nexus_outputs_mgr = std::make_shared<utils::PerNexusCsvOutputMgr>(nexus_ids, manager->get_output_config().root);
+        // CSV nexus output: one file per nexus. With per_rank grouping under MPI,
+        // each rank's files go in a rank_<N>/ subdirectory (created by the
+        // manager) to cut filesystem contention from many ranks sharing a dir.
+        std::string nexus_subdirectory;
+        #if NGEN_WITH_MPI
+        if (mpi_num_procs > 1 &&
+            manager->get_output_config().nexus.grouping == realization::config::OutputGrouping::per_rank) {
```
PhilMiller: Should this also be available for catchment CSV output, to have per-catchment files, grouped into subdirectories by rank?

src/realizations/catchment/Bmi_Module_Formulation.cpp
hellkite500: I revisited #959 for context here, and I think this work should base off of that.  I think there are some refinements to that PR we want to consider before merging, so I'll get some review feedback over there soon.

include/realizations/config/output.hpp
```diff
@@ -0,0 +1,157 @@
...
+    //! Serialization format for an output domain.
+    enum class OutputFormat { csv, netcdf };
+
+    //! How a domain's per-feature records are distributed across files:
+    //!   per_feature - one file per feature (one CSV per catchment / per nexus)
+    //!   per_rank    - grouped by MPI rank. For catchment CSV this is a single
+    //!                 aggregated file per rank with a feature-id column; for
+    //!                 nexus CSV this is per-feature files under a rank_N/ subdir.
+    //!                 Reduces filesystem contention on large MPI runs.
+    enum class OutputGrouping { per_feature, per_rank };
```
robertbartel: I’m not sure the design of the `OutputFormat` and `OutputGrouping` options fundamentally makes sense.  A `per_rank` grouping would require uniform outputs, which is not guaranteed for catchments (is it enforced somewhere).  Catchments really would need something akin to a `per_formulation` option.  

Nexus probably don't need `per_formulation` as strictly but probably would benefit.  Further, if `PerFormulationNexusOutputMgr` is still a thing (pretty sure it is), it doesn't make a lot of sense to not have such a grouping option.

robertbartel (changes requested):
I've only done a partial review, but there seems to me to be a very fundamental problem related to grouping options and the lack of (guaranteed) uniform catchment data.

include/utilities/output/PerNexusCsvOutputMgr.hpp
robertbartel: FWIW, I also don't really like this signature change.  The manager object only needs to be concerned about where output goes.  I don't see the benefit of decomposing that into two elements like this.  Why not just rely on the caller to prepare the arg appropriately?

include/realizations/config/output.hpp
hellkite500: I think there's a straight forward way to unify this, and I'll push a reviewable commit that hopefully makes it clear.

docker/ngen.dockerfile
PhilMiller, line 1: The commit message should be amended to match the content change in here.

include/utilities/output/CatchmentOutputsMgr.hpp
```diff
@@ -0,0 +1,183 @@
...
+        OutputField(std::string source_name, std::string output_name,
+                    std::optional<std::string> units = std::nullopt,
+                    std::map<std::string, std::string> attributes = {})
+            : source_name(std::move(source_name)), output_name(std::move(output_name)),
+              units(std::move(units)), attributes(std::move(attributes)) {}
+
+        //! Single name used for both source and output (an un-aliased column). The const char*
+        //! overload lets a bare string literal be a column (one user-defined conversion).
+        OutputField(std::string name) : OutputField(name, name) {}
+        OutputField(const char *name) : OutputField(std::string(name), std::string(name)) {}
```
PhilMiller: Please make these `explicit`, if you feel the simplified overloads really need to be offered.

include/utilities/output/CatchmentOutputsMgr.hpp
```diff
@@ -0,0 +1,183 @@
...
+     * One feature's output schema, supplied to a manager at construction.
+     *
+     * A backend uses the full set of these to lay out its sink before any data arrives (e.g. open
+     * files and write their headers). One per (formulation, feature).
+     */
+    struct FeatureDescriptor
+    {
+        std::string formulation_id;
+        std::string catchment_id;
+        std::vector<OutputField> fields;   //!< output fields, in the order values are supplied
```
PhilMiller, line 163: For a fixed formulation ID, do we actually want/expect to support varying sets of `OutputField` per catchment? I'm concerned by the per-catchment duplication.

include/utilities/output/CatchmentCsvOutputMgr.hpp
```diff
@@ -0,0 +1,114 @@
...
+         *        filename prefix. Created here (along with any per-formulation subdirectories) if it
+         *        does not exist, so callers need not pre-create it.
+         * @param aggregated_filename File name used for each formulation's shared file when @p
+         *        aggregate is true (e.g. "cat_output.csv", or "cat_rank_<N>.csv" for a per-rank MPI file).
+         * @param precision Significant digits applied uniformly when rendering all values to text.
+         * @param descriptors The full set of catchments (and their fields) this manager will write;
+         *        files are opened and headers written for all of them here, so the manager is ready to
+         *        receive data immediately after construction.
+         */
+        CatchmentCsvOutputMgr(bool aggregate, std::string output_root, std::string aggregated_filename,
```
PhilMiller: The two parameters `bool aggregate` and `string aggregated_filename` are really representing a single compound parameter - a sum type akin to `std::optional<std::string> aggregated_filename`. 

I maybe shouldn't care much, since the weird aggregated CSV format will probably fade into the sands of time once NetCDF is implemented.

include/utilities/output/CatchmentCsvOutputMgr.hpp
```diff
@@ -0,0 +1,114 @@
...
+                              int precision, std::vector<FeatureDescriptor> descriptors);
+
+        //! Convenience for per-feature (one file per catchment) output.
+        CatchmentCsvOutputMgr(std::string output_root, std::vector<FeatureDescriptor> descriptors,
+                              int precision = DEFAULT_PRECISION)
+            : CatchmentCsvOutputMgr(false, std::move(output_root), "", precision, std::move(descriptors)) {}
```
PhilMiller: I'm resistant to these sorts of convenience constructor overloads when the use cases is that instances are mostly coming from a single factory, rather than being instantiated by many independent bits of code.

src/utilities/output/CatchmentCsvOutputMgr.cpp
```diff
@@ -0,0 +1,197 @@
...
+#include <filesystem>
+#include <sstream>
+#include <iomanip>
+#include <stdexcept>
+#include <utility>
+
+using utils::CatchmentCsvOutputMgr;
+
+namespace {
+    //! Join the column output names with commas for a CSV header. CSV writes only each column's
```
PhilMiller, line 31: Out of curiosity, what is the `//!` comment format? I see it used inconsistently throughout this commit.

src/utilities/output/CatchmentCsvOutputMgr.cpp
```diff
@@ -0,0 +1,197 @@
...
+    //! output_name; the source_name, units, and other OutputField metadata are not emitted here.
+    std::string join_columns(const std::vector<utils::OutputField> &columns)
```
PhilMiller, line 33: This and `format_values` hard-code a comma as the delimiter, while the functions below that prepend catchment ID and time step take `delimiter` as a parameter.

These are all scoped to this anonymous namespace, so they're completely private to the implementation. The callers don't pass a delimiter argument, so I think the parameters should just be dropped, and comma hard-coded uniformly.

If you want to be **really** ridiculous, you could template the whole class on `char delimiter`, so that it could be instantiated as tab-separated, null-separated, or colon-separated or whatever.

include/utilities/output/CatchmentOutputsMgr.hpp
hellkite500: Under current implementation, any given catchment can override a formulation and configure it uniquely with a formulation spec that isn't the same as another catchment using the same formulation but with different configuration (is that a valid configuration? I don't know, but its possible...)

include/utilities/output/CatchmentCsvOutputMgr.hpp
```diff
@@ -0,0 +1,114 @@
...
+        const bool aggregate_;
+        const std::string output_root_;
+        const std::string aggregated_filename_;
+        const int precision_;
+        bool closed_ = false;
+
+        // Open output streams keyed by resolved file path. In per-feature mode each (formulation,
+        // catchment) has its own path; in aggregated mode every catchment of a formulation resolves
+        // to the same path and shares one stream.
+        std::unordered_map<std::string, std::shared_ptr<std::ofstream>> streams_;
```
PhilMiller: Might as well go full `std::filesystem::path`, and not deal in `string` at all when actually referring to assembled paths. The necessary hash and comparison operators are defined for it.

include/utilities/output/CatchmentCsvOutputMgr.hpp
hellkite500: I thought I fixed this, that must have slipped through my several rebases.

src/utilities/output/CatchmentCsvOutputMgr.cpp
hellkite500: Its another doxygen style comment format that has mixed use throughout some portions of the code...its similar to a block `/** */` style comment.  I've found it simpler to read for shorter one, maybe two line comments.  The larger block style header comments are nice when you have `@brief` and then longer documentation.

src/utilities/output/CatchmentCsvOutputMgr.cpp
hellkite500: I thought about making delimiter a configurable component, but figured it wasn't really needed right now.  Maybe setting things up to use `constexpr` delimiter uniformly and have a clean slate to add a configurable feature in the future is the best approach right now.

include/utilities/output/CatchmentOutputsMgr.hpp
hellkite500: In the latest refactor, an attributes map is attached to all `field`s, which we can later extend/configure to carry these kinds of attributes that an output manager may want to do something with.  That should come as a follow up to this PR.

test/realizations/catchments/Bmi_Fortran_Formulation_Test.cpp
hellkite500: This was a bit of testing legacy from the `precision` configuration, I think.  Since that functionality is now specifically the responsibility of the output manager, I pushed these tests to numerical precision.  Since `get_output_values_for_timestep` is now intended to produce numeric values (`vector<double>`) I think it is pertinent for the testing of that functionality to assert valid values at higher precision.

include/core/Layer.hpp
PhilMiller: This likely needs to be split to 
> `virtual ~Layer();`

here and 

> `Layer::~Layer() = default;`

in `Layer.cpp`, to avoid linker errors on hypothetical TUs that `#include Layer.hpp` but don't link against `ngen::outputs`, because they'll be missing the destructor for `CatchmentOutputsMgr`

include/realizations/catchment/Bmi_Module_Formulation.hpp
```diff
@@ -61,28 +61,46 @@ namespace realization {
...
-         * are calculated values for all variables.
-         *
-         * The default delimiter is a comma.
-         *
-         * Implementations will throw `invalid_argument` exceptions if data for the provided time step parameter is not
-         * accessible.  Note that, for this type, only the last processed time step is accessible, because formulations
-         * do not save results from previous time steps.  This also has the consequence of there being no valid set of
-         * arguments before a least one call to @ref get_response has been made.
+         * Only the last processed time step is accessible, because formulations do not save results from previous
+         * time steps; there is thus no valid time step before at least one call to @ref get_response has been made.
```
PhilMiller: That's probably actually "at least one call to @ref update has been made"

include/realizations/catchment/Catchment_Formulation.hpp
```diff
@@ -31,36 +30,26 @@ namespace realization {
...
              *
-             * @param timestep The time step for which data is desired.
-             * @param delimiter The value delimiter for the string.
-             * @return A delimited string with all the output variable values for the given time step.
+             * @param timestep The time step for which values are desired.
+             * @return The output values for @p timestep, one per output column.
              */
-            virtual std::string get_output_line_for_timestep(int timestep,
-                                                             std::string delimiter = DEFAULT_FORMULATION_OUTPUT_DELIMITER) = 0;
+            virtual std::vector<double> get_output_values_for_timestep(int timestep) = 0;
```
PhilMiller, line 52: Do we actually have a use case for maintaining the `int timestep` parameter? The only implementation present or foreseen requires callers to pass the current timestep as an argument, so there's no information being gained, just a weird cross-check.

include/realizations/catchment/Bmi_Formulation.hpp
```diff
@@ -234,6 +213,15 @@ namespace realization {
             output_header_fields = output_headers;
         }
 
+        /**
+         * The configured/derived output header (display) field names, positionally parallel to
+         * @ref get_output_variable_names -- header i is the output name for variable i. A concrete
+         * subclass uses these as each column's output_name when building @ref get_output_fields.
+         */
+        const std::vector<std::string> &get_output_header_field_names() const {
```
PhilMiller: The return type could be cut down to `boost::span<const std::string>` rather than the concrete `vector const&`

PhilMiller: (I'm going to make suggestions like this and `std::string_view` on any new code that's getting added, to start shifting the project's practice that direction)

include/realizations/catchment/Bmi_Module_Formulation.hpp
```diff
@@ -61,28 +61,46 @@ namespace realization {
...
+         * Throws `std::invalid_argument` if @p timestep is not the current (last processed) time step.
          *
          * @param timestep The time step for which data is desired.
-         * @return A delimited string with all the output variable values for the given time step.
+         * @return The output values for the given time step, one per output column.
+         */
+        std::vector<double> get_output_values_for_timestep(int timestep) override;
```
PhilMiller, line 77: If I were being unreasonable and obsessive about hypothetical performance, I would say that this should be `void get_output_values_for_timestep(int timestep, boost::span<double> values) override;`, with an added `int get_num_output_values()` to ensure the caller's buffer is big enough. That would be _massive_ overkill here, though.

PhilMiller: It is possible that on some systems, an unfortunate `malloc` implementation or constellation of usage could produce surprising overhead here, that would warrant the above suggestion, though.

Maybe it would just be good performance hygiene to adopt it proactively, and avoid having to think about the pitfalls.

src/NGen.cpp
```diff
@@ -524,15 +565,17 @@ int main(int argc, char* argv[]) {
...
         #if NGEN_WITH_NETCDF
-        nexus_outputs_mgr = std::make_shared<utils::PerFormulationNexusOutputMgr>(nexus_ids, formulation_ids, manager->get_output_root(), timesteps);
+        nexus_outputs_mgr = std::make_shared<utils::PerFormulationNexusOutputMgr>(nexus_ids, formulation_ids, manager->get_output_config().root, timesteps);
         #else
         throw std::runtime_error("NetCDF support required to use per-formulation nexus files.");
         #endif
 
         #endif
     }
     else {
```
PhilMiller: 1. Do we need to look at whether nexus output is enabled? If we're (eventually) driving t-route directly, then it won't need to be
2. Should this `else` be an explicit `format == CSV` check instead?

test/realizations/catchments/Bmi_Cpp_Formulation_Test.cpp
```diff
@@ -260,7 +262,18 @@ TEST_F(Bmi_Cpp_Formulation_Test, Initialize_output_header_fields) {
...
 
-    EXPECT_EQ(form.get_output_header_line(","), "custom_1,custom_2");
+    // Each field maps a source variable to its configured output (header) name; units come from the
+    // model (OUTPUT_VAR_1 -> "mm/s", OUTPUT_VAR_2 -> "m/s"), positionally aligned with the values.
+    const std::vector<utils::OutputField> fields = form.get_output_fields();
+    ASSERT_EQ(fields.size(), 2u);
+    EXPECT_EQ(fields[0].source_name, "OUTPUT_VAR_1");
+    EXPECT_EQ(fields[0].output_name, "custom_1");
+    ASSERT_TRUE(fields[0].units.has_value());
+    EXPECT_EQ(*fields[0].units, "mm/s");
```
PhilMiller, line 272: This can actually be simplified to `EXPECT_EQ(fields[0].units, "mm/s");`, at least as far as `optional<T> == T` goes. Google Test's macro may be confused by that.

doc/REALIZATION_CONFIGURATION.md
```diff
@@ -21,15 +21,38 @@ The Configuration is a key-value object and must contain these three first level
...
+  * `grouping` — how a domain's records are distributed across files (the file-*granularity* axis):
+    * `"per_feature"` (default) — one file per catchment / per nexus.
+    * `"per_formulation"` — the catchments of a formulation aggregate into one file with a leading `catchment_id` column (one file per formulation). Fewer, larger files, with columns uniform within each file by construction. (Nexus CSV aggregation is not yet implemented.)
+  * `rank_subdir` — whether an MPI rank's files are placed under a `rank_<N>/` subdirectory (the filesystem-*layout* axis, orthogonal to `grouping`; default `false`). Only affects distributed (multi-rank) runs — a no-op in serial — and reduces filesystem contention when many ranks share a directory. For `per_formulation` under MPI it is applied automatically regardless of this setting, since each rank must write its own file to avoid collisions.
+
+```
+"output": {
+    "root": "/path/to/output/",
+    "catchment": { "enable": true, "format": "csv", "grouping": "per_formulation", "rank_subdir": true },
+    "nexus":     { "enable": true, "format": "csv", "grouping": "per_feature",     "rank_subdir": true }
```
PhilMiller, line 41: Can `"nexus": { "enable": false, ... }` actually be honored by the code right now? It seemed like not

include/realizations/config/catchment_output.hpp
```diff
@@ -0,0 +1,58 @@
...
+                const Output& output_config,
+                std::vector<utils::FeatureDescriptor> descriptors,
+                const std::string& aggregated_filename = DEFAULT_AGGREGATED_FILENAME)
+        {
+            if (output_config.catchment.format != OutputFormat::csv) {
+                throw std::runtime_error("Catchment output: only the CSV format is currently supported.");
+            }
+            const bool aggregate = output_config.catchment.grouping == OutputGrouping::per_formulation;
+            return std::make_shared<utils::CatchmentCsvOutputMgr>(
+                aggregate, output_config.root, aggregated_filename, output_config.precision, std::move(descriptors));
```
PhilMiller: For the future structural evolution here, I'd prefer if the code for the CSV case were under an affirmative condition `format == OutputFormat::csv`, and then the exception were in an `else`

src/NGen.cpp
PhilMiller: Regardless of whether disabling nexus output is a desired thing, the `"enable"` key is documented for it, so it should be honored, or the documentation changed

include/utilities/output/CatchmentOutputsMgr.hpp
PhilMiller: Looking over the whole PR as it stands today, the only caller is in a test, and that's similarly covered by an `EXPECT_THROW` from a subsequent attempt at submitting data. I think we could drop this.

hellkite500: I'm not really sold either way, but this method *is* part of the nexus manager abstraction.  There is real potential to consolidate these two interfaces in the `ngen::output` library eventually, and I think they have pretty similar parity as of the latest state here.  Including/dropping `is_closed()` here is just a matter of how much parity we want between the two currently independently defined API's.

include/core/DomainLayer.hpp
hellkite500: This has been refactored and and domain and surface layers use the same base, with different instances created in main to drive the behavior differences between them via the same base API.

src/utilities/output/CatchmentCsvOutputMgr.cpp
hellkite500: Unfortunately C++ doesn't support optional reference types until C++ 26.  The boost optional does, but ultimately I ended up just simplifying these functions to directly write the `ofstream` in the receive function anyways and dropped this function.

include/utilities/output/CatchmentOutputsMgr.hpp
hellkite500: After looking at the pattern a little more, I decided that forcing construction to provide units and not have a default nullopt provided a better "caller makes explicit decision on units" pattern, and instead of extending that into these, I just dropped them.  They were useful for test code to construct test cases succinctly, but a small helper function in the test code keeps it readable and wraps the full constructor.

include/utilities/output/CatchmentCsvOutputMgr.hpp
hellkite500: This was mostly for simplified test construction, but dropping this along with the previously noted decision to make units a required arg make this marginally less useful, so I dropped it as well.

src/NGen.cpp
hellkite500: Thinking about this symmetry, nexus output is currently *always on*, but it is really only required when running *integrated routing* or when a routing run is planned post ngen.  I went ahead and made this optional the same way catchment output is, since there may be valid research and development cases for just looking at catchment output and not nexus output.  With an early stop error if routing is requested AND nexus output is disabled.

include/realizations/catchment/Bmi_Module_Formulation.hpp
hellkite500: I'm not inclined to touch this at the moment.  If/when a real system/case arises that shows the need, then adopting the pattern seems like a pretty straight forward change once this current refactor is merged.

test/realizations/catchments/Bmi_Cpp_Formulation_Test.cpp
hellkite500: The resulting print on a failure isnt useful.
The current form gives
```
  *fields[0].units
    Which is: "mm/s"
  "WRONG_UNIT"
```
on failure. 
```cpp
EXPECT_EQ(fields[0].units, "mm/s");
```
gives
```
  fields[0].units
    Which is: ("mm/s")
  "WRONG_UNIT"
    Which is: 0x102904541
```
options are extracting the expected value to a string (current form) or
```cpp
EXPECT_EQ(fields[0].units, std::string("mm/s"))
```
to clean up the failure messages.

include/utilities/output/CatchmentCsvOutputMgr.hpp
hellkite500: Fun fact, gcc8 doesn't have a hash implementation for the filesystem::path, so adopting this change required bumping the toolset in the CI runner gcc 12 (minimum requirement for gcc is now 9 by taking this approach...) Bumped the minimum version in the dependency docs as well.

src/core/SurfaceLayer.cpp
```diff
@@ -71,9 +71,13 @@ void ngen::SurfaceLayer::update_models(boost::span<double> catchment_outflows,
...
+        if (nexus_outputs_mgr) {
+            nexus_outputs_mgr->receive_data_entry(id, current_time_marker, contribution_at_t);
+        }
 
         //std::cout<<"\tNexus "<<id<<" has "<<contribution_at_t<<" m^3/s"<<std::endl;
     } //done nexuses
-    nexus_outputs_mgr->commit_writes();
+    if (nexus_outputs_mgr) {
+        nexus_outputs_mgr->commit_writes();
+    }
```
PhilMiller, line 82: Not for this PR, but I'm thinking that the nexus and catchments output managers referenced by other code should always exist, and they should be responsible for deciding what does or doesn't get passed along, and to where. Besides better separating concerns, that also would support things like multiple configured outputs hidden behind the single call from the simulation logic.

include/realizations/config/catchment_output.hpp
```diff
@@ -45,12 +46,15 @@ namespace realization
                 std::vector<utils::FeatureDescriptor> descriptors,
                 const std::string& aggregated_filename = DEFAULT_AGGREGATED_FILENAME)
         {
-            if (output_config.catchment.format != OutputFormat::csv) {
+            if (output_config.catchment.format == OutputFormat::csv) {
+                const bool aggregate = output_config.catchment.grouping == OutputGrouping::per_formulation;
+                return std::make_shared<utils::CatchmentCsvOutputMgr>(
+                    output_config.root,
+                    aggregate ? std::optional<std::string>(aggregated_filename) : std::nullopt,
```
PhilMiller, line 53: Tiny stylistic thing, that doesn't need to be fixed for this PR: that can be `std::make_optional(aggregated_filename)`

src/utilities/output/CatchmentCsvOutputMgr.cpp
```diff
@@ -157,10 +133,20 @@ void CatchmentCsvOutputMgr::receive_data_entry(const std::string &formulation_id
...
-    (*out) << catchment_output_data_row(aggregate_, catchment_id, data_time_marker.sim_time_index,
-                                        data_time_marker.time_stamp, format_values(values, precision_)) << "\n";
+    if( aggregated_filename_.has_value() ) {
+        (*out) << catchment_id << DELIMITER;
+    }
+    // Add time step/timestamp
+    (*out) << data_time_marker.sim_time_index << DELIMITER
+           << data_time_marker.time_stamp << DELIMITER;
+    // Add values, precision set on the stream at construction.
+    for (std::size_t i = 0; i < values.size(); ++i) {
```
PhilMiller, line 151: There's an edge case here when `values.size() == 0`, in that the line will end with `DELIMITER` following the `time_stamp`.

I was going to say we needn't care, but it's actually a simplification to fix it: drop `<< DELIMITER` just above, and write it unconditionally down here.

include/realizations/catchment/Bmi_Module_Formulation.hpp
```diff
@@ -61,28 +61,46 @@ namespace realization {
...
+            // (which defaults to the variable name, or is a configured alias).
+            const std::vector<std::string> &variables = get_output_variable_names();
+            boost::span<const std::string> headers = get_output_header_field_names();
+            std::vector<utils::OutputField> fields;
+            fields.reserve(variables.size());
+            for (std::size_t i = 0; i < variables.size(); ++i) {
+                std::optional<std::string> units;
+                try {
+                    units = get_bmi_model()->GetVarUnits(variables[i]);
+                } catch (const std::exception &) {
```
PhilMiller: To spare thinking too hard about the `throw`, how about
```suggestion
                } catch (...) {
```
Since we're not actually looking at the substance of the exception.

src/realizations/catchment/Bmi_Module_Formulation.cpp
```diff
@@ -16,27 +16,28 @@ namespace realization {
...
             static bool no_conversion_message_logged = false;
             if (!no_conversion_message_logged) {
                 no_conversion_message_logged = true;
                 logging::warning("Output variables do not have unit conversion. Capability not yet implemented in ngen.");
             }
 
-            std::string output_str;
-            for (const std::string& name : get_output_variable_names()) {
-                // Placeholder to request no conversion
+            const std::vector<std::string>& names = get_output_variable_names();
```
PhilMiller: As a teensy bit of future-evolution-proofing, could this be spelled `auto const& name = ...`? I don't want to later change that to return a `span<string>`, and then burn the benefit by constructing a temporary `vector` copy of the whole thing.

src/realizations/catchment/Bmi_Module_Formulation.cpp
```diff
@@ -381,15 +382,25 @@ namespace realization {
...
                     out_headers[i] = out_headers_json_list[i].as_string();
                 }
-                set_output_header_fields(out_headers);
+                if (get_output_variable_names().size() == out_headers.size()) {
+                    set_output_header_fields(out_headers);
+                }
+                else {
+                    std::cerr << "WARN: configured output headers have " << out_headers.size()
+                              << " fields, but there are " << get_output_variable_names().size()
+                              << " output variables" << std::endl;
```
PhilMiller, line 401: I'd *prefer* if this were a hard error, but understand doing it this way for now.

test/realizations/config/CatchmentOutput_Test.cpp
```diff
@@ -0,0 +1,139 @@
...
+        while (std::getline(in, line)) lines.push_back(line);
+        return lines;
+    }
+
+    // A unique, process-scoped output root (trailing slash) so concurrent test binaries don't collide
+    // over fixed names. Not created here -- the manager creates it, which is part of what we exercise.
+    std::string fresh_root() {
+        static int counter = 0;
+        std::filesystem::path d = std::filesystem::temp_directory_path()
+            / ("ngen_cat_factory_test_" + std::to_string(getpid())) / std::to_string(counter++);
```
PhilMiller: Do we care about tidying up the intermediate PID-specific directory as the test process exits?

src/utilities/output/CatchmentCsvOutputMgr.cpp
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    //! Build the CSV header line. When include_id is true a leading "catchment_id"
+    //! column is prepended (aggregated output, where one file holds many catchments);
+    //! otherwise the layout is the long-standing "Time Step,Time,<vars>". The column
+    //! order is kept identical to catchment_output_data_row so header and data rows
+    //! always agree.
+    std::string catchment_output_header_line(bool include_id,
+                                             const std::string &variable_header,
+                                             const std::string &delimiter = ",")
```
PhilMiller:
```suggestion
                                             const std::string &delimiter = DELIMITER)
```

PhilMiller:
I'm pretty happy with the design now. A few tiny nitpicks to address or not, and I'm fine with merging this and moving forward.

include/realizations/config/catchment_output.hpp
christophertubbs: This is for my own education, but, event though you _can_ use the shorter style that more mimics the `make_shared` calls, wouldn't it be a better idea to maintain the `std::optional<type>(value)` approach as a sort of preferred example since `make_optional(value)` has the potential to cause surprising and hard to track behavior due to decay deduction if/when optional declarations are used elsewhere in the code base? If the current form remains, I imagine it would be more likely that others would use this pattern as well.

src/utilities/output/CatchmentCsvOutputMgr.cpp
hellkite500: This requires an explicit conversion, so something like
```cpp
const std::string &delimiter = std::string(DELIMITER))
```
The other options is to simply make both these parameters `string_view`
```cpp
    std::string catchment_output_header_line(bool include_id,
                                             const std::string_view variable_header,
                                             const std::string_view delimiter = DELIMITER)
    {
        std::string line;
        if (include_id) {
            line += "catchment_id";
            line += delimiter;
        }
        line += "Time Step";
        line += delimiter;
        line += "Time";
        line += delimiter;
        line += variable_header;
        return line;
    }

}
```
which simply trades off the ability to use concatenation and has to be implemented using `+=` in the body.
I've applied this change to keep the consistency with using the `std::string_view` for the delimiter, and as a nod to "use the view type instead of the reference" mechanisms.

test/realizations/config/CatchmentOutput_Test.cpp
hellkite500: Looking a little closer, this mechanism is probably best supported by gtests environment setup/teardown
```
struct TempDirCleanup : ::testing::Environment {
        void SetUp() override    { std::filesystem::remove_all(pid_test_root()); }
        void TearDown() override { std::filesystem::remove_all(pid_test_root()); }
    };
    const auto* temp_dir_cleanup = ::testing::AddGlobalTestEnvironment(new TempDirCleanup);
```
so that the paths are cleaned once at startup and once at exit.  I think this pattern could (and should) be extended to some other places throughout the testing code that deals with tmp/intermediate disk writes, but that's a PR for another day.  Once this merges, an issue to consolidate across the test code and utilize a shared `filesystem` helper and environment would make sense.

I'll push this refactor here so we have something to work with during that consolidation.

include/realizations/catchment/Bmi_Module_Formulation.hpp
hellkite500: Applied

src/realizations/catchment/Bmi_Module_Formulation.cpp
hellkite500: Applied

PhilMiller (approved):

robertbartel (approved):
There are still things about this I'm uneasy with.  It still seems possible to add two catchment formulations under one manager/file that don't use the same output fields (if I've missed the guard for that, let me know).  Also, I think certain settings (e.g., a `per_formulation` grouping) will cause `rank_subdir` in the realization config to be silently ignored (the silently part is more of the problem).  And there is a little more being changed in one PR than I'd like, given the complexity and the somewhat messy existing config structure.

But I don't want to trump consensus at this point, or hold up something that is progress in the right direction.  And I know more work on this is coming.

hellkite500:
Thanks for the feedback!

>It still seems possible to add two catchment formulations under one manager/file that don't use the same output fields (if I've missed the guard for that, let me know).

I think this is a technically reachable edge case that could be identified during initialization and a fail fast approach implemented.  I'm inclined to open this as an issue and PR the fix in a single commit to avoid churning this current branch.

>Also, I think certain settings (e.g., a `per_formulation` grouping) will cause `rank_subdir` in the realization config to be silently ignored (the silently part is more of the problem).

This is true, and acknowldged in the realization configuration documentation as

"rank_subdir — … For per_formulation under MPI it is applied automatically regardless of this setting, since each rank must write its own file to avoid collisions."

> And there is a little more being changed in one PR than I'd like, given the complexity and the somewhat messy existing config structure.

Agreed, this took on a larger shape that intended at first, but lays out a better foundation for more modular work in the future.  The commit sequencing was intended to *help* the size issue, but may not have been as useful as I thought.

> But I don't want to trump consensus at this point, or hold up something that is progress in the right direction. And I know more work on this is coming.

I do have a few noted follow ups that this larger refactor and implementation will make easier to digest, so given consensus on the stability of this branch, I'm ok with merging, noting the edge case issue with a potential fix coming in after this.

PR #1000, opened 2026-07-08
refactor: netcdf cache offsetting and index manipulation logic

aaraney:
Refactor the netcdf cache offsetting and index manipulation functions. This hopefully cleans up the logic a bit and makes it a little easier to follow and reason about (please be the judge of that :)). This also fixes a bug when the total forcing time steps are not divisible by the cache line size.

I still need to add tests, but a first review would be great! Thanks!

### DIFF
--- a/src/forcing/NetCDFPerFeatureDataProvider.cpp
+++ b/src/forcing/NetCDFPerFeatureDataProvider.cpp
@@ -494,18 +494,86 @@ size_t NetCDFPerFeatureDataProvider::get_ts_index_for_time(const time_t &epoch_t
     }
 }
 
+namespace cache {
+    std::size_t page_count(std::size_t c_size, std::size_t line_size) {
+        return (c_size / line_size) + std::min<std::size_t>(1, c_size % line_size);
+    }
+
+    std::size_t page_p_idx(std::size_t c_idx, std::size_t line_size) {
+        return c_idx / line_size;
+    }
+
+    std::size_t page_c_idx(std::size_t c_idx, std::size_t line_size) {
+        std::size_t p_idx = page_p_idx(c_idx, line_size);
+        return p_idx * line_size;
+    }
+
+    std::size_t page_p_idx_to_c_idx(std::size_t p_idx, std::size_t line_size) {
+        return p_idx * line_size;
+    }
+
+    std::size_t page_entry_j_idx(std::size_t c_idx, std::size_t line_size) {
+        return c_idx - page_c_idx(c_idx, line_size);
+    }
+
+    std::size_t page_cache_line_size(std::size_t c_idx, std::size_t c_size, std::size_t line_size) {
+        std::size_t ps = page_count(c_size, line_size);
+        std::size_t p_idx = page_p_idx(c_idx, line_size);
+        assert(p_idx < ps);
+        if (p_idx < ps - 1) {
+            return line_size;
+        }
+        std::size_t ts = page_c_idx(c_idx, line_size);
+        return c_size - ts;
+    }
+
+    std::size_t page_entry_idx(std::size_t i_idx, std::size_t c_idx, std::size_t c_size, std::size_t line_size) {
+        std::size_t pls = page_cache_line_size(c_idx, c_size, line_size);
+        // NOTE: this is relative to the default line size
+        std::size_t pj_idx = page_entry_j_idx(c_idx, line_size);
+        return (pls * i_idx) + pj_idx;
+    }
+}
+
+
 double NetCDFPerFeatureDataProvider::get_value(const CatchmentAggrDataSelector& selector, ReSampleMethod m) 
 {
+    /*
+     * this cache is made up of pages.
+     * each page contains cache lines.
+     * a page is a flattened 2d array with dimensions catchment-id, time (i, j).
+     * connecting pages contiguously along cache lines (left-to-right), forms the time series.
+     * this implicit dimension is named the 'c' dimension.
+     * the 'c' dimension is the (page idx + page relative cache line j idx) * cache line size.
+     * 'c' is a forcing time step idx.
+     *
+     * each line, dim i, contains forcing data for a cache line sized time period for a single catchment.
+     * there are n cache lines in a page.
+     * n is the number of catchments a single ngen process is responsible for.
+     * pages are divided by _cache line size_ (index p).
+     * cache line size equates to n forcing time steps.
+     * a page only contains data for a single forcing variable.
+     *
+     * if the forcing time dim size is not divisible by the cache line size,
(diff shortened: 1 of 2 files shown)


src/forcing/NetCDFPerFeatureDataProvider.cpp
```diff
@@ -588,9 +588,9 @@ double NetCDFPerFeatureDataProvider::get_value(const CatchmentAggrDataSelector&
         }
         // Find all values in the current cache slice and push them onto raw_values
         while(value_idx >= cache_slice_start && 
-              value_idx < cache_slice_start + cache_slice_t_size &&
+              value_idx < cache_slice_start + adjusted_size &&
               value_idx <= idx2){
-            raw_values.push_back(cached->at((cat_idx * cache_slice_t_size) + (value_idx % cache_slice_t_size)));
+            raw_values.push_back(cached->at((cat_idx * adjusted_size) + (value_idx % cache_slice_t_size)));
```
PhilMiller: Not having thought this through fully yet, should the modulus still be `cache_slice_t_size`?

As an intermediate step to the very clean usage of `mdspan` in the far future where that's widely available, you could define a `span` of `cached` containing the source data, which can then be indexed from 0 to the appropriate length. That seems like it should be fine, since `value_idx` is incremented in each iteration, rather than advanced by some larger step, but the modulo makes me wonder if there's some circular buffer stuff going on that I'm missing.

aaraney: I don't think either is right. I think the particular testing data I was using was divisible by the cache line size in the right way where this worked as expect.

Long story short, it was easier to refactor the code into something a little easier to follow than fix the bug.

aaraney:
for me: see #998 for testing inspiration

edit: and #922?

PhilMiller:
This all would be much easier to work with if we had [`mdspan`](https://github.com/kokkos/mdspan) available. Standardized in C++23, but implemented back to C++14.

@hellkite500 @robertbartel Do you think we'd be OK importing a dependency on that header-only library? Is it worth running this by Chris and Donald? Does it help that the reference implementation came from another US government agency, with strict provenance and review processes on their open-source releases?

src/forcing/NetCDFPerFeatureDataProvider.cpp
```diff
@@ -511,58 +579,50 @@ double NetCDFPerFeatureDataProvider::get_value(const CatchmentAggrDataSelector&
...
 
-    const std::size_t read_len = idx2 - idx1 + 1;
+    const std::size_t read_len = c_idx2 - c_idx1 + 1;
 
     std::vector<double> raw_values;
     raw_values.reserve(read_len);
 
-    std::size_t idx1_cache_slice_start = idx1 - (idx1 % cache_slice_t_size);
-    std::size_t time_idx = idx1 % cache_slice_t_size;
+    std::size_t cache_line_size = cache_slice_t_size;
```
PhilMiller, line 606: I think this may still be incorrect for the case of a single page. I.e. a very short run, in which there are fewer time steps than the chunking granularity that will determine `cache_slice_t_size`

aaraney: If I follow correctly, I think this saves us?

https://github.com/aaraney/ngen/blob/463b86ba06b07718f6e8f8a226b14f88adf77984/src/forcing/NetCDFPerFeatureDataProvider.cpp#L810-L812
```c++
    if (num_times && cache_slice_t_size > num_times){
        cache_slice_t_size = num_times;
    }
```

test/forcing/NetCDFPerFeatureDataProvider_Test.cpp
```diff
@@ -220,6 +222,155 @@ TEST_F(NetCDFPerFeatureDataProviderTest, TestForcingDataRead)
...
+
+    static double cellValue(std::size_t cat, std::size_t t) {
+        return 100.0 * cat + t;
+    }
+
+    // Creates a forcing file with n_cats catchments ("cat-0".."cat-N"), n_times
+    // hourly timesteps starting at kStartEpoch, and one variable "temp" (units K)
+    // holding cellValue(cat, t). time_chunk == 0 leaves the variable contiguous.
+    std::string makeForcing(const std::string& name, std::size_t n_cats, std::size_t n_times, std::size_t time_chunk) {
+        std::string path = "./" + name;
```
aaraney, line 244: Let me know if we'd rather write this somewhere else (e.g. temp dir).

aaraney:
@hellkite500, @PhilMiller: I added some tests. Please have a look when you get a chance. Thanks!

src/forcing/NetCDFPerFeatureDataProvider.cpp
PhilMiller: I can believe it does. Hopefully this case is covered by a test.

test/forcing/NetCDFPerFeatureDataProvider_Test.cpp
PhilMiller: In a temp (sub-)directory would be good. It should also include data from the current-running test process (e.g. PID) to ensure non-collision with other instances of the test that may be running concurrently, or that may have left behind debris.

I commented about this in some more detail on another recent PR dealing with output file writing (by Nels or Bobby). Let me know if you need help finding it. Maybe the bits in question can be extracted to a shared helper and re-used here.

test/forcing/NetCDFPerFeatureDataProvider_Test.cpp
```diff
@@ -220,6 +222,155 @@ TEST_F(NetCDFPerFeatureDataProviderTest, TestForcingDataRead)
...
+{
+    const std::size_t n_cats = 3, n_times = 50, chunk = 24; // pages: 24, 24, 2
+    auto path = makeForcing("nc_cache_straddle.nc", n_cats, n_times, chunk);
+    NetCDFPerFeatureDataProvider provider(path, kStartEpoch, kStartEpoch + n_times * kStride, utils::getStdErr());
+
+    double tol = 1e-9;
+
+    // timesteps 23..24 straddle the boundary between pages 0 and 1
+    CatchmentAggrDataSelector two("cat-1", "temp", kStartEpoch + 23 * kStride, 2 * kStride, "K");
+    EXPECT_NEAR(provider.get_value(two, data_access::MEAN),
```
PhilMiller, line 353: Given the focus of these test cases on the cache bits, any reason to exercise `MEAN` as well, rather than simplifying things by just using `SUM` throughout?

PhilMiller:
I'm good with the code and the coverage of the testing. What's the team's sense of how picky we want to be about temporary files in the test code? If the conclusion is "not very", then we can just merge this, and tidy it up later.

PhilMiller (approved):

PR #1003, opened 2026-07-14
Delete never-used mdarray and mdframe

PhilMiller:
These aimed at being nice abstractions for various forms of multi-dimensional time-series data that we'd handle. However, they never ended up finding use in the rest of the codebase. So, they're just more code we're carrying and testing, without added value.

## Removals

- `mdarray` and `mdframe`, along with their tests

## Testing

1. CI

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows project standards (link if applicable)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

PR #1006, opened 2026-08-11
Single-process multithreaded parallelization of catchment simulation

PhilMiller:

## Additions

- CMake: add an option `NGEN_WITH_THREADING` to enable this behavior

## Removals

-

## Changes

- `Layer::update_models`: run the catchment simulations in a `std::for_each`, possibly in parallel if `NGEN_WITH_THREADING` is enabled

## Testing

1.

## Notes

-

## Todos

-

## Checklist

1.

- [ ] Linux


src/core/Layer.cpp
```diff
@@ -29,7 +29,9 @@ void ngen::Layer::update_models(boost::span<double> catchment_outflows,
     // in this timestep (mirrors SurfaceLayer).
     utils::time_marker current_time_marker(
         output_time_index, simulation_time.get_current_epoch_time(), current_timestamp);
-    for(const auto& id : processing_units) {
+    auto b = begin(processing_units);
+    auto e = end(processing_units);
+    std::for_each(std::execution::par, b, e, [&](const std::string& id) {
```
PhilMiller: One pitfall to note here that I find from reading descriptions of the parallel algorithms specs is that if there's an uncaught exception below this point in the call stack (e.g. one of the BMI formulations fails for whatever reason), it will `std::terminate()` the process, and not let us handle it at a higher level.

If we want to add fine-grained resilience logic for operational purposes, we'd need to add `catch` logic here for anything we intend to handle, and record it somehow to address outside the parallel execution.

PhilMiller: OK, I need to read the docs more carefully. The above bit about `terminate()` maybe only applies to `par_unseq` (i.e. vectorized) execution, with `par` execution gathering up and re-`throw`ing an `exception_list` of everything that was thrown and caught inside.

christophertubbs:
`AppleClang` for C++17 doesn't support parallel execution policies, so that's just going to break on mac. Don't know how much of a deal breaker the lack of mac support is since we don't have the mac population that we used to.

What's the deal with the following?

```  /home/runner/work/ngen/ngen/include/utilities/mdframe/variable.hpp:227:33: error: ‘constexpr boost::static_visitor<R>::static_visitor() [with R = boost::variant<int, float, double>]’ is protected within this context
    227 |         auto visitor = std::bind(
        |                        ~~~~~~~~~^
    228 |             visitors::mdarray_at<SupportedTypes...>{},
        |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    229 |             std::placeholders::_1,
        |             ~~~~~~~~~~~~~~~~~~~~~~
    230 |             index
        |             ~~~~~                
    231 |         );
        |         ~                        
```

PhilMiller:
I have no idea what's going on with the mdframe thing, but since it's been unused since it was introduced, we could [drop it](https://github.com/NOAA-OWP/ngen/pull/1003)

PhilMiller:
> AppleClang for C++17 doesn't support parallel execution policies, so that's just going to break on mac. Don't know how much of a deal breaker the lack of mac support is since we don't have the mac population that we used to.

That's disappointing, but easy enough to address. I've got the bits for optional compilation in CMake already, I just haven't wired it through to say `std::execution::seq` when it's disabled.

christophertubbs:
I think you can force gcc rather than AppleClang. It's just not super obvious.

I'm rerunning the tests on #1003 . If they pass, I'll go ahead and merge. If not, I'll bug you to fix that one. That should fix the mdframe stuff since it's all about deleting the problematic content.

christophertubbs:
Why the change from C++20 to C++17?

PhilMiller:
> Why the change from C++20 to C++17?

The move to C++20 would have been beyond the reasonable scope of this PR, and wasn't really necessary. See Layer.cpp line 72 for the only use I wanted.

christophertubbs:
> The move to C++20 would have been beyond the reasonable scope of this PR, and wasn't really necessary. See Layer.cpp line 72 for the only use I wanted.

Gotcha - C++20 wasn't the problem, the engineering workflow of making a ton of disconnected changes was.

PhilMiller:
Two reasons this is "single-process" right now:
1. MPI requires that if we're going to have multiple threads at all, we must replace the call to `MPI_Init` with `MPI_Init_thread` with an appropriate argument describing how threads will interact with MPI calls
2. The thread-safety modifications in #1014 addressed thread-safety of `HY_PointHydroNexus`, but not its `NGEN_WITH_MPI` counterpart `HY_PointHydroNexusRemote`

PhilMiller:
Testing this approach has gotten hung up on the various available model codes all doing things that cause them to fail when multiple instances are called concurrently in a single process.

Noah-OWP-Modular, for instance, uses hard-coded unit numbers for several files of input parameters. These race between instances making open, read, and close calls.

Something is causing cfe to report `BMI_FAILURE` during initialization as well. It has at least a few global variables that could be causing trouble.

PhilMiller:
I did some exploration with locking around initialization of individual BMI models, and found that I could get some speedup with extra threads if I also added code to pre-load the init config files in each thread before taking the lock and calling into the model. This was on Ursa, with the init config files on the Lustre-backed `/scratch4`, so probably the worst case for getting at those files.

This is messy, but it shows that there is speed-up potential if we make the models suitably thread-safe.

PhilMiller:
Ugh, this runs into serious issues when using a module like Noah-OWP-Modular, which is 
* implemented in Fortran
* `open()`s and reads config files beyond the top-level initconfig named by the argument to BMI `initialize()`

Fortran has a rule against multiple 'units' [1] being simultaneously connected to a single file. The runtime of both GNU and Intel Fortran actually checks for this, even when all such operations are read-only.

I'll still benchmark this, with the PET+CASAM formulation, just to see what's possible, but this presents a serious barrier to production usage.

[1] Think "file descriptors" if you know C/UNIX stuff

PR #957, opened 2026-04-25
BMI Formulation Serialization and Deserialization implemented via a `bmi::protocol`

hellkite500:
BMI State Serialization and Deserialization Protocol.
This is an opt-in convention that lets a BMI model participate in ngen's state checkpoint/restore workflow by exposing four reserved BMI variables (`ngen::serialization_create`, `_free`, `_size`, `_state`). The engine side is a pair of sibling protocols

- `NgenSerializationProtocol` (save)
- `NgenDeserializationProtocol` (restore)

registered under `NgenBmiProtocols` on every BMI formulation. Records share a single length-prefixed archive file keyed by a new engine-level `compound_id()` per formulation, so a multi-feature, multi-submodule simulation can checkpoint into one file and restore any entity by id. Scale work, MPI-agnostic path templating, and full cross-language test-model support (C / C++ / Fortran / Python) are included.

See `doc/BMI_SERIALIZATION_PROTOCOL.md` for the model-author-facing spec and `test/utils/bmi/` for end-to-end coverage.

## Additions

- `SerializationRecord` on-disk format (length-prefixed boost archives)
  with truncation tolerance and a `MAX_RECORD_ARCHIVE_BYTES` sanity cap
- `NgenSerializationProtocol` (save) and `NgenDeserializationProtocol`
  (restore), with metadata-only `check_support`, `is_fatal`-driven
  escalation at `initialize()`, and the usual frequency gating
- `CheckpointIndex` — process-global per-path lazy-built index turning
  restore from O(features × records) into O(records + features); exposed
  `clear_checkpoint_indexes()` for memory-bounded callers
- `compound_id()` accessor on `Bmi_Formulation` with multi-submodule
  3-part key injection (`<feature>.<n>:<submodule-mtn>:<multi-mtn>`)
- `realization::config::GlobalConfigKey` + `apply_config` —
  realization-level inheritable config blocks with per-formulation override
- `utilities::resolve_path_tokens` + `{{rank}}`/`{{pid}}`/`{{host}}`/`{{date}}`
  template support so per-rank/host/date paths stay an engine concern
  (the protocol itself is MPI-agnostic)
- `Bmi_Py_Adapter` `uint8`/`int8`/`byte` dispatch branch for opaque-byte
  marshaling across the Python language boundary
- BMI serialization protocol support in all four test models
  (test_bmi_c, test_bmi_cpp, test_bmi_fortran, test_bmi_py — pickle-based)
[...]

PR #1009, opened 2026-08-11
add a compound_id which can compose a feature identifier along with with module identifiers into a single id

hellkite500:
Introduce an engine-level compound identity for BMI formulation instances. Designed as a unique per-formulation key.

Default composition for a single-BMI formulation:

    "<this->id>:<model_type_name>"
    e.g. "cat-1:bmi_c_cfe"

For a Bmi_Multi_Formulation submodule, the parent injects a three-part compound via `set_compound_id()` inside `init_nested_module`, BEFORE the submodule's `create_formulation()` runs — so any engine-level consumer that reads `compound_id()` during submodule construction sees the full key:

    "<this->id>:<submodule-mtn>:<multi-mtn>"
    e.g. "cat-1.0:bmi_c_cfe:bmi_multi_noahowp_cfe"

The submodule's `this->id` is already `<catchment>.<index>` from Bmi_Multi_Formulation's existing identifier construction; we append its own `model_type_name` (peeked from the config map so it's available pre-create_formulation) and the enclosing multi's `model_type_name`.

`this->id` is not mutated — it stays the feature identifier used by current code paths (output naming, forcing lookup, logs). The compound is a read-only view of the formulation's place in the realization, built from fields already in hand.

This is used in #957, and was extracted as an independent feature from that effort. 

## Additions

- a formulation compound_id_ member and getter/setters

## Checklist

### DIFF
--- a/include/realizations/catchment/Bmi_Formulation.hpp
+++ b/include/realizations/catchment/Bmi_Formulation.hpp
@@ -140,6 +140,43 @@ namespace realization {
         }
 
         /**
+         * Get this formulation instance's compound identity.
+         *
+         * Default composition for a single-BMI formulation:
+         *   "<this->id>:<model_type_name>"             e.g. "cat-1:bmi_c_cfe"
+         *
+         * For a Bmi_Multi_Formulation submodule, `Bmi_Multi_Formulation`
+         * injects a three-part compound via `set_compound_id()` before
+         * `create_formulation()` runs:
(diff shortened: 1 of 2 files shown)


PhilMiller:
I don't especially like the "set from above" pattern, but it's common in our formulation architecture, so this isn't the place to litigate it.

That said, it will pretty readily allow us to accommodate in-process ensemble members and the like down the line.

Overall, I think the arrangement here is sound enough, and we need this or something like it to move the serialization work forward.

PhilMiller (approved):

PR #1010, opened 2026-08-11
Extend existing clang format rules

hellkite500:
This can be take or leave, but these were the rules used to apply the styling commit in #957


christophertubbs:
I was prepared to rubber stamp, but single line `if` statements make my skin crawl. I also stick to prepending class members in member functions with `this->`, so my opinion only matters so much in this arena 😄 

You get an auto-rubber stamp from me if that line goes mysteriously missing.

PR #1011, opened 2026-08-12
correct off-by-one in end-of-run sentinel

hellkite500:
NgenMassBalance::run treated `ctx.current_time_step == ctx.total_steps` as the end-of-run sentinel when frequency == -1. The Context contract puts current_time_step in [0, total_steps - 1] (end-of-step counter), so the comparison gated against an out-of-range value and the negative-frequency branch was dead at every in-range counter value.

The original comparison was authored with begin-of-step semantics in mind, where K == total_steps would name the boundary before the would-be Nth step (which is never computed). Mixing the two perspectives made the branch unreachable in practice.

Existing frequency_end tests passed against the buggy comparison because they manually constructed contexts with current_time_step == total_steps, matching the bug rather than the contract. They're rewritten to drive in-range; frequency_end_single_step, frequency_end_large_n, and frequency_negative_other_than_minus_one cover the sentinel edges.

### DIFF
--- a/test/realizations/catchments/Bmi_C_Formulation_Test.cpp
+++ b/test/realizations/catchments/Bmi_C_Formulation_Test.cpp
@@ -883,15 +883,18 @@ TEST_F(Bmi_C_Formulation_Test, check_mass_balance_frequency) {
     double mass_error;
     mass_error += 10; // Force a mass balance error above tolerance
     get_friend_bmi_model(formulation)->SetValue(OUTPUT_MASS_NAME, &mass_error); //
+    // Context contract: current_time_step is drawn from [0, total_steps - 1].
+    // For a 3-step run with frequency=2, indices 0 and 2 fire (0 % 2 == 0,
+    // 2 % 2 == 0) and index 1 is skipped.
     //Check initial mass balance -- should error which indicates it was propoerly checked
     //per frequency setting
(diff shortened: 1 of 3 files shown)


test/realizations/catchments/Bmi_C_Formulation_Test.cpp
```diff
@@ -592,15 +592,18 @@ TEST_F(Bmi_C_Formulation_Test, check_mass_balance_frequency) {
     double mass_error;
     mass_error += 10; // Force a mass balance error above tolerance
     get_friend_bmi_model(formulation)->SetValue(OUTPUT_MASS_NAME, &mass_error); //
+    // Context contract: current_time_step is drawn from [0, total_steps - 1].
+    // For a 3-step run with frequency=2, indices 0 and 2 fire (0 % 2 == 0,
+    // 2 % 2 == 0) and index 1 is skipped.
     //Check initial mass balance -- should error which indicates it was propoerly checked
```
PhilMiller:
```suggestion
    //Check initial mass balance -- should error which indicates it was properly checked
```
May as well fix this typo while you're at it

PhilMiller (dismissed):

hellkite500:
Applied the suggested typo fix.  Suggest simply squashing this PR and keeping the first commit's message.

PhilMiller (approved):

PhilMiller:
Rebased and force-pushed to run the tests afresh. Will then merge.

PR #1013, opened 2026-08-14
Simulation_Time thread safety

PhilMiller:
Make `Simulation_Time::get_timestamp` non-mutating, and use a thread-safe date/time formatting function.

## Testing

1. Local run of test suite
2. Confirmed output of identical `current_timestamp` values from `SurfaceLayer` before and afterward
3. CI

## Checklist

### DIFF
--- a/include/simulation_time/Simulation_Time.hpp
+++ b/include/simulation_time/Simulation_Time.hpp
@@ -106,10 +106,11 @@ class Simulation_Time
      * @brief Accessor to the current timestamp string
      * @return current_timestamp
      */ 
-    std::string get_timestamp(int current_output_time_index)
+    std::string get_timestamp(int current_output_time_index) const
     {
-        // "get" method mutates state!
-        current_date_time_epoch = start_date_time_epoch + current_output_time_index * output_interval_seconds;
+        if (start_date_time_epoch + current_output_time_index * output_interval_seconds != current_date_time_epoch) {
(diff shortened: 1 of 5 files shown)


PhilMiller:
This change may be a little bit touchy, and I'd like to get confirmation that the output in various forms is not negatively impacted. I printed values myself, and they seem to be the same, but I don't want to trust just my own analysis.

include/simulation_time/Simulation_Time.hpp
```diff
@@ -106,19 +106,20 @@ class Simulation_Time
      * @brief Accessor to the current timestamp string
      * @return current_timestamp
      */ 
-    std::string get_timestamp(int current_output_time_index)
+    std::string get_timestamp(int current_output_time_index) const
     {
-        // "get" method mutates state!
-        current_date_time_epoch = start_date_time_epoch + current_output_time_index * output_interval_seconds;
```
robertbartel, line 112: While I very much like this, I'm still trying to reason through whether it will cause a problem for `NgenSimulation::get_timestamp_for_step`.  Pretty sure the latter isn't used anywhere right now, but potential future uses could be a problem ... maybe.  I suppose the solution there might be to just clean it up entirely.

PhilMiller: It looks like that function is actually unused. Maybe I can just delete it in this PR to clear up the possibility of concern?

robertbartel (approved):

PR #1014, opened 2026-08-14
Make all code called under Layer::update_models thread-safe

PhilMiller:
This is the set of changes necessary to safely run the loop over catchment formulations in multiple threads.

I manually walked down the call graph to identify these changes, relying on the earlier `const` additions and other changes in #1012 and #1013 to cut off inspections.

In runs not using multi-threading, the changes in this PR add two kinds of overhead
* Memory footprint of the mutexes and such
* Execution time overhead of un-contended atomic operations to increment, lock, and unlock

I believe both should be negligible for our purposes.

This is separate from a PR introducing the mechanism for actual parallel execution, because that should come with greater scrutiny of the design choice and performance impacts.

## Changes

- Add `std::mutex` members to the nexus and catchment CSV output classes, and the logging namespace
- Lock on those mutexes when accessing shared state from member functions of those classes
- Catchment CSV output's stream lookup was modified to potentially use more than a shared mutex (e.g. per-stream, or picked from a lock pool)
- Atomic accumulation of catchment outflows
- Turn a boolean flag for whether a particular warning has been printed once into an atomic flag

## Testing

1. CI
2. I provisionally added multithreading as seen in #1006 and ran with ThreadSanitizer, to observe a clean run. Thus, I believe this PR is comprehensive in its coverage.

## Checklist

## Testing Notes

Actually applying ThreadSanitizer (TSan) usefully was a bit of a challenge. The issue is that the implementation of `std::for_each(std::execution::par, ...)` is rather janky in libstdc++ (as provided by GCC), and so is GCC's implementation of TSan. The library depends on the open-sourced version of Intel TBB, but silently falls back to a serial implementation if that is not linked into the final binary. GCC does not properly instrument `atomic_thread_fence` operations for TSan, while TBB uses such operations heavily for synchronization. Thus, a build of the ngen code with GCC, TSan, and naive packaged TBB results in a huge mass of false positives from TBB itself.

To obtain meaningful testing, eventually reporting no issues, I did the following:
* Built TBB from source using Clang with TSan enabled
* Built ngen with Clang and TSan enabled
* Added linkage with `-ltbb` to for the `ngen` executable

### DIFF
--- a/src/core/Layer.cpp
+++ b/src/core/Layer.cpp
@@ -18,8 +18,6 @@ void ngen::Layer::update_models(boost::span<double> catchment_outflows,
                                 std::unordered_map<std::string, int> const& nexus_indexes,
                                 int current_step)
 {
-    std::mutex accumulation_mutex;
-
     //std::cout<<"Output Time Index: "<<output_time_index<<std::endl;
     if(output_time_index%1000 == 0) std::cout<<"Running timestep " << output_time_index <<std::endl;
     std::string current_timestamp = simulation_time.get_timestamp(output_time_index);
@@ -56,12 +54,7 @@ void ngen::Layer::update_models(boost::span<double> catchment_outflows,
 #if NGEN_WITH_ROUTING && NGEN_WITH_ROUTING_TROUTE_BMI
         int results_index = catchment_indexes.at(id);
         // XXX: This is currently accumulating in meters of depth, which may not be desirable
-        {
-            // XXX: Switch when we can use C++20
-            //std::atomic_ref(catchment_outflows[results_index]) += response;
-            std::lock_guard g(accumulation_mutex);
-            catchment_outflows[results_index] += response;
-        }
+        std::atomic_ref(catchment_outflows[results_index]) += response;
 
 #endif // NGEN_WITH_ROUTING && NGEN_WITH_ROUTING_TROUTE_BMI
         if (catchment_output_mgr) {
--- a/src/core/Layer.cpp
+++ b/src/core/Layer.cpp
@@ -18,6 +18,8 @@ void ngen::Layer::update_models(boost::span<double> catchment_outflows,
                                 std::unordered_map<std::string, int> const& nexus_indexes,
                                 int current_step)
 {
+    std::mutex accumulation_mutex;
+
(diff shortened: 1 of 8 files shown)


PhilMiller:
If we're *really* concerned about the overheads from this, I could conditionalize all of the stuff seen here in one way or another. Rather than literally wrapping every such line, I could define a `using ngen_mutex = null_mutex` vs `using ngen_mutex = std::mutex`, and make `ngen_mutex::[un]lock` be no-ops, such that `std::lock_guard` would be trivial. That seems like overkill for probably-not-a-problem.

src/core/Layer.cpp
```diff
@@ -54,7 +54,8 @@ void ngen::Layer::update_models(boost::span<double> catchment_outflows,
 #if NGEN_WITH_ROUTING && NGEN_WITH_ROUTING_TROUTE_BMI
         int results_index = catchment_indexes.at(id);
         // XXX: This is currently accumulating in meters of depth, which may not be desirable
-        catchment_outflows[results_index] += response;
+        std::atomic_ref(catchment_outflows[results_index]) += response;
```
PhilMiller, line 57: Eek, I need to be careful - this is a C++20-ism. I noted that in the rolled-up PR before I started picking it apart

PhilMiller: g++ is actually fine with this, even in GCC 11.4 as shipped in Ubuntu 22.04 used in CI. Clang errors about it specifically, though, reporting that it's a C++20 feature being used in a C++17 compilation.

Operational conformance also means that we need to be strict about our language version usage until we get the go-ahead to upgrade to C++20, and actually do so.

PhilMiller: The nicest alternative of `atomic<double>` is also a C++20-ism.

I'm just going to switch it to a lock on a local mutex, since I don't expect this to be the common path right now anyway.

PhilMiller: We've got C++20 now, and available compilers and standard library implementations seem to be good enough to support this, so I'll use it as it's both more convenient, more expressive, and potentially more performant.

christophertubbs (approved):
I don't see anything objectionable.

PR #1015, opened 2026-08-18
CatchmentCsvOutputManager: Avoid hash<std::filesytem::path> compatibility pitfall

PhilMiller:
Switch use of an `unordered_map<std::filesystem::path, stream*>` to `map<path, stream*>` to avoid the need for a hash function for `std::filesystem::path`. This is an issue because C++17 inadvertently omitted that requirement, and compilers' standard libraries conforming to C++17 didn't implement it until several versions later. For instance, GCC 9 supports full C++17, but the hash isn't implemented until 11.4.0, and is missing even in 11.2.0.

Cf:
https://cplusplus.github.io/LWG/issue3657
And the very last table row here:
https://en.cppreference.com/cpp/compiler_support/17

## Testing

1. Local build
2. CI

## Notes

- I think the hypothetical performance impact of this should be negligible, especially since we should move away from CSV-per-catchment output when we care about performance anyway.

## Checklist

### DIFF
--- a/docker/ngen.dockerfile
+++ b/docker/ngen.dockerfile
@@ -7,9 +7,10 @@ RUN dnf update -y \
     && dnf install -y --allowerasing tar git gcc-toolset-12 make cmake udunits2-devel coreutils \
     && dnf clean all
 
-# Rocky 8's system compiler is GCC 8, whose libstdc++ lacks complete C++17 <filesystem>
-# support (notably std::hash<std::filesystem::path>). Build with the gcc-toolset-12 SCL
-# toolchain instead, and point the runtime loader at its libstdc++.
+# Rocky 8's system compiler is GCC 8, which doesn't support C++20 and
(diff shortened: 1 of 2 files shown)


PhilMiller:
I may also simplify out the corresponding bits in the Docker file and the dependencies documentation.

PhilMiller:
Ahmad ran into this on the NSF Anvil cluster, which provides the not-good-enough GCC 11.2.0.

hellkite500:
> I may also simplify out the corresponding bits in the Docker file and the dependencies documentation.

We should update these to drop the note about the hash limitation since it is no longer a blocker.  GCC 9 still becomes the minimum for the std::filesystem implementation, and I don't see any real performance considerations that are going to matter for the typical use of this.

hellkite500 (approved):
With the formal move to c++ 20, this may not really be a needed move, but it also isn't going to hurt in any way that I can see right now.

PR #1017, opened 2026-08-20
Support auxiliary hydrofabric attribute tables as source for catchment attributes

robertbartel:
Some hydrofabric files (or, supplemental files) may be created that contain per-divide model params for different formulations (e.g., from regionalization tools).  These may be in tables named something other than `divide-attributes` - the previous standard - or even multiple tables.  This PR adds configuration and functional capabilities to join data from these auxiliary tables with catchment features data from `divides` table.  This then makes the auxiliary data available to the existing mechanisms for setting formulation module parameter values via the realization config; i.e., with the `model_params` `"source": "hydrofabric"` syntax.  For example:

```json
"model_params": {
    "bexp": { "source": "hydrofabric", "from": "reg.b" }
}
```

The PR introduces a new, optional, top-level key for the realization config: `auxiliary_hydrofabric_attributes`.  It can be used as follows:

```json
"auxiliary_hydrofabric_attributes": [                     
    {
        "table": "cfexnom__v___regn_jul26_cfexnom_clust_test2_gower_agglomerative_k15",
        "alias": "reg",              // optional; prefix defaults to the table name
        "file": "path/to/other.gpkg",// optional; defaults to the CLI catchment data file
        "key_column": "divide_id",   // optional; this is the default used to join table to `divides`
        "required": false            // optional; this is the default
    }
]
```

Tables are expected to be in the catchment data hydrofabric file provided on the command line, but optionally another file can be provided via the config.

A configured table must exist.  Rows only must exist if `"required": true` within that table's entry in `"auxiliary_hydrofabric_attributes"`; if `false`, params will be skipped for missing-row catchments.


aaraney:
This is something i've wanted for a while! This is great!

Having just glanced at the semantics, I was a little confused that the `source` remains `hydrofabric` when pulling from an auxiliary. What was your thinking in keeping `source: hydrofabric` over `source: reg` / the full table name?

robertbartel:
The design works by joining the aux data to divide data from the hydrofabric.  So, we are appending something to the data already exposed to things when `source: hydrofabric` is used.  In that sense, it isn’t a different source.  

At a higher level, I was anticipating the/a common use case would be for the aux tables to actually be included in the hydrofabric file (largely because I was working with an example file where this was the case).  In that situation also, `source: hydrofabric` seemed natural and appropriate.

robertbartel:
Converting to draft for the moment.  The design of #960 (and potentially further refinements to it; see #1021) will likely impact the details of this significantly.  Some initial review is fine, but full formal review should probably wait until #960 is settled.

PhilMiller:
I haven't looked at the code yet, but from the description, am I right to understand that these values must still be in the same file as the hydrofabric itself? I think we can pretty clearly see to a future where that won't be desirable - Hydrofabric v.N is finalized, and then we're doing various runs to generate auxiliary parameter values associated with it. We should expect to reference the hydrofabric from a read-only file, and load other data that references it from elsewhere (e.g. another SQLite file).

PhilMiller:
That "multiple SQLite file" thing would even still allow full database query functionality:
https://stackoverflow.com/questions/6824717/sqlite-how-do-you-join-tables-from-different-databases

PhilMiller:
Oops, I'm apparently illiterate today. Your description showed a "file: other.gpkg" line, and I totally missed it.

JoshCu:
As a stepping stone to updating all the models to allow full configuration via bmi, could this be paired with something like fuse to replace the per catchment config files?

The way I'm generating all the files at the moment is just a bunch of templates that look like this https://github.com/CIROH-UA/NGIAB_data_preprocess/blob/modular-configs/modules/data_sources/config/catchment/cfe.ini where the placeholders match the column names in my attribute table in the gpkg.
I'm basically just doing this
```python
with sqlite3.connect(hydrofabric) as conn:
    df = pandas.read_sql_query(
        """
        SELECT
        d.areasqkm,
        d.lengthkm,
        da.*
        FROM divides AS d
        JOIN 'divide-attributes' AS da ON d.divide_id = da.divide_id
        """,
        conn,
    )
template = Path(file.cfe_template).read()
for _,row in df.iterrows():
    Path(output_folder / f"{row[id]}.ini").write(template.format(**row))
```
It would be great if I could give ngen table names and template paths and have it fake a bunch of `.data_root/config/cat_config/<table_name>/<cat_id>.<extension>` files for each of the models.

If the performance isn't awful it would be nice to cut down on the ~50M config files used per day in the datastream :smile:

robertbartel:
As discussed above, this is in _Draft_ status while #960 and #1021 get sorted out.  Originally, this work did not account for any of the v4 work, since it isn't committed to _master_ yet.  The current PR branch still reflects that.

However, in a different branch -  [i/aux-attrs/on-refined-hf40](https://github.com/robertbartel/ngen/tree/i/aux-attrs/on-refined-hf40) - I have reworked the design for this aux attribute functionality to be based on the v4 hydrofabric support, including the latest requested refinements (i.e., #1021).   

For now, keeping the PR branch as is to reflect what is in _master_, but if/when the other PRs get settled, I'll update this branch to reflect the redesign.  In the mean time, _i/aux-attrs/on-refined-hf40_ can be used directly if desired.

robertbartel:
> As a stepping stone to updating all the models to allow full configuration via bmi, could this be paired with something like fuse to replace the per catchment config files?

I personally feel very uneasy about going down this road with the current design and building things around BMI.  

BMI can be unspecific in certain places, but the documentation on the `initialize` function makes it pretty clear that BMI expects models to read inputs from configuration files.  I think we already use a bit of a generous interpretation of the docs to justify also setting model param values directly.  But it fits because those capabilities supplement the standard BMI functions, but do not replace them.   

If ngen _needs_ to stop loading models from config files - and I get why loading millions of configs isn’t ideal - then I strongly believe we should frame the discussion as one about integrating via something other than BMI.  Because (short of getting the BMI spec itself changed), that _is_ what such a change does.  I'm not necessarily saying that's a great idea or something we should prioritize, but I think it's important we avoid the inconsistency of calling something "BMI" if it goes against the spec.

PR #1019, opened 2026-08-20
Move to require C++20

PhilMiller:
C++20 adds a variety of features that will make our development more convenient and the resulting code more reliable and performant.

E.g.
* `std::span`
* Ranges
* Container heterogeneous lookup
* Container `contains()` members
* Floating-point atomics
* `std::atomic_ref`
* String `begins_with` and `ends_with`
* Text formatting with `std::format`
* Calendar and Time Zone support in `std::chrono`

The latter two unfortunately aren't implemented until later compiler/library releases, so we may need to hold off on adoption for a bit.

## Changes

- Top-level CMake: set C++ standard from 17 to 20
- Code tweaks to fix up deprecated and ill-formed code that no longer compiles
- Documentation updates to match the necessary compiler versions
- CI: Add jobs using Clang 14 on Ubuntu 22.04 and GCC 13 on macOS 15, to get more cross-version compatibility coverage

## Testing

1. AppleClang and GCC locally and in CI workflows
2. Clang and GCC on Ubuntu 22.04 in CI workflow

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows project standards (link if applicable)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

PR #1021, opened 2026-08-21
Draft refinements to HF v4 support

robertbartel:
Introduces a new library strictly for hydrofabric classes, with version, schema versioning, readers and detectors.  Separates and generalize geopackage reading, and then has hydrofabric readers use (when available).

This PR could be considered as a continuation of or addendum to #960.  The branch here contains that work plus some additional commits in response to requested changes.   One could review and approve (effectively, if not literally due to the _Draft_ status) this branch as an indication that #960, once fast-forwarded with the additional commits here, was approved. 

However, because of the significant difference between even just this and #960, it may be preferred to hold here until the other is merged independently, and then take this up under a totally separate scope.


include/geopackage/GeoPackageReader.hpp
```diff
@@ -0,0 +1,85 @@
...
+ */
+class GeoPackageReader
+{
+  public:
+    /**
+     * Take ownership of an open database.
+     *
+     * @param[in] db Open GeoPackage database to read from
+     */
+    explicit GeoPackageReader(sqlite::database db);
```
PhilMiller, line 33: Given the ownership transfer, should this be `sqlite::database&&`?

include/hydrofabric/GeoPackageHydrofabricReader.hpp
```diff
@@ -0,0 +1,231 @@
...
+
+/**
+ * Common behavior of the hydrofabric v4 readers.
+ *
+ * Both v4 variants renamed the nexus identifier columns (`id`/`toid` to `nexus_id`/`nexus_toid`),
+ * so nexus translation lives here, as does the summary warning about divides left without a
+ * downstream reference. What the variants do not share is how a divide reaches its nexus, which is
+ * left to attribute_divide_toid() in the concrete subclasses.
+ */
+class V4GeoPackageHydrofabricReader : public GeoPackageHydrofabricReader
```
PhilMiller: Slight naming inconsistency between `V2_GeoPackage..` and `V4GeoPackage...` - please pick one.

test/CMakeLists.txt
```diff
@@ -163,6 +167,37 @@ ngen_add_test(
...
+########################## Hydrofabric Unit Tests
+# The GeoJSON implementation needs no SQLite, so it is testable in a build without it -- which is
+# also the build where the factory's refusal of GeoPackage paths is the only path available.
+ngen_add_test(
+    test_hydrofabric
+    OBJECTS
+        hydrofabric/GeoJSONHydrofabricReader_Test.cpp
+    INCLUDES
+        # FileChecker, which nothing this target links exposes on its own.
+        ${PROJECT_SOURCE_DIR}/include/utilities
```
PhilMiller, line 182: This seems like a mis-use of the just-added `INCLUDES` parameter, vs exposing `FileChecker` as (part of) a utilities library target.

src/hydrofabric/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+add_library(hydrofabric STATIC
```
PhilMiller, line 1: This library is maybe mis-named, since the contents are "hydrofabric reading/loading", not hydrofabric stuff in general (which is kinda what the `HY_*` classes in `core` are)

src/hydrofabric/GeoPackageHydrofabricSchema.cpp
```diff
@@ -0,0 +1,152 @@
...
+    const std::vector<std::string>& divides_columns)
+{
+    if (nexus_columns.empty()) {
+        throw std::runtime_error(
+            "hydrofabric detect_version: no nexus table"
+        );
+    }
+
+    const bool has_nexus_id =
+        std::find(nexus_columns.begin(), nexus_columns.end(), "nexus_id")
```
PhilMiller: I'm not asking for a change here, but I'll just note that with [C++20](#1019), we can spell this pattern
```
std::ranges::find(nexus_columns, "nexus_id") != nexus_columns.end()
```

Probably better to just wait for C++23 
```
std::ranges::contains(nexus_columns, "nexus_id")
```

PhilMiller: Alternately, if these were instead `std::set`, then we'd have `set::contains` from C++20 today

include/geopackage/ngen_sqlite.hpp
```diff
@@ -146,13 +146,18 @@ class database
     //! @param table name of table
     //! @return true if table does exist
     //! @return false if table does not exist
-    bool contains(const std::string& table);
+    bool contains(const std::string& table) const;
+
+    //! List the column names of a table
+    //! @param table name of table
+    //! @return column names in declaration order, empty if the table does not exist
+    std::vector<std::string> columns(const std::string& table) const;
```
PhilMiller: Is there anywhere in here that needs the returned collection to be indexed? If not, let's just return `set` instead of `vector` here, and make other code simpler (and marginally faster)

src/hydrofabric/GeoPackageHydrofabricSchema.cpp
```diff
@@ -0,0 +1,152 @@
...
+    // throw its detailed error if the schema doesn't match any known version,
+    // rather than reporting it as some version it isn't.
+    return detect_version(nexus_db, divides_db);
+}
+
+HydrofabricVersion detect_hydrofabric(const sqlite::database& db) {
+    return detect_hydrofabric(db, db);
+}
+
+std::string get_layer_id_column(const HydrofabricVersion version, const std::string& layer,
```
PhilMiller, line 89: I notice that neither this nor its callers check that the returned column name actually exists in the target `layer`, except in one case. So, if that were incorrect, it would lead to a confusing error from the subsequent code in the caller that queries a non-existent column.

PhilMiller:
I've gone through the source changes here (but not the tests) as a whole stand-alone PR, and it all reads well to me. I'd generally be in favor of merging this in whole (vs #960 as an increment) if there isn't some functional shortcoming I'm missing.

include/hydrofabric/GeoPackageHydrofabricReader.hpp
```diff
@@ -0,0 +1,231 @@
...
+     *
+     * @param[in] layer Hydrofabric layer about to be read
+     * @throws std::runtime_error if a required column is absent
+     */
+    virtual void check_required_columns(const std::string& layer) const;
+
+    /**
+     * Translate a freshly read divides collection into the property names ngen consumes.
+     *
+     * @param[in,out] divides Divides collection to update in place
```
hellkite500, line 94:
```suggestion
     * @param[in,out] divides collection to update in place
```

include/hydrofabric/GeoPackageHydrofabricReader.hpp
```diff
@@ -0,0 +1,231 @@
...
+     * @throws std::runtime_error if a row cannot be translated
+     */
+    virtual void fixup_divides(geojson::FeatureCollection& divides) const = 0;
```
hellkite500: This function name and the function description are not intuitively connected at all...at least not to me...

include/hydrofabric/GeoPackageHydrofabricReader.hpp
```diff
@@ -0,0 +1,231 @@
...
+
+    /**
+     * Translate a freshly read nexus collection into the property names ngen consumes.
+     *
+     * @param[in,out] nexus Nexus collection to update in place
```
hellkite500, line 108:
```suggestion
     * @param[in,out] nexus collection to update in place
```

include/hydrofabric/GeoPackageHydrofabricReader.hpp
```diff
@@ -0,0 +1,231 @@
...
+     * @throws std::runtime_error if a row cannot be translated
+     */
+    virtual void fixup_nexus(geojson::FeatureCollection& nexus) const = 0;
```
hellkite500: Same as `fixup_divides`, the name/description confuse me

include/hydrofabric/GeoPackageHydrofabricReader.hpp
```diff
@@ -0,0 +1,231 @@
...
+
+    void check_required_columns(const std::string& layer) const override;
+
+    void fixup_nexus(geojson::FeatureCollection& nexus) const override;
+
+    /**
+     * Alias each divide's downstream nexus to "toid", then warn once with a count of the divides
+     * that ended up without one, rather than once per divide.
+     *
+     * @param[in,out] divides Divides collection to update in place
```
hellkite500, line 174:
```suggestion
     * @param[in,out] divides collection to update in place
```

include/hydrofabric/GeoPackageHydrofabricReader.hpp
```diff
@@ -0,0 +1,231 @@
...
+     */
+    void fixup_divides(geojson::FeatureCollection& divides) const override;
+
+    /**
+     * Attribute "toid" on one divide, however this variant expresses it.
+     *
+     * Leaving "toid" unset marks the divide terminal, which is what an absent downstream reference
+     * means to every consumer downstream of the loader.
+     *
+     * @param[in,out] properties Property map for the divide
```
hellkite500, line 185: this param pattern repeats in a not as clear way here, but the "properties Property" is odd reading, but its not clear to me what the intent of this is (it seemed more obvious with the Nexus/Catchment ones I noted above...but now I'm not sure all around)

hellkite500: actually the id below kinda explains it, its the function parameter name followed by the description.  You can probably ignore this and other comments, its just a form I'm not used to reading.  A separator ( . or : maybe?) could help readability perhaps...This is now a style nit though

include/hydrofabric/GeoPackageHydrofabricSchema.hpp
```diff
@@ -0,0 +1,135 @@
...
+std::string get_layer_id_column(HydrofabricVersion version, const std::string& layer, const sqlite::database& db);
+
+/**
+ * Precompute the divide_id -> flowpath_toid map used to synthesize "toid" on V4_0_BETA1 divides
+ * rows, which carry no native toid and instead reference their downstream nexus indirectly via
+ * flowpath_id -> flowpaths.flowpath_toid. Built once up front so the per-row loop can do a
+ * hashtable lookup instead of N queries.
+ *
+ * Returns an empty map outside of (V4_0_BETA1, "divides") -- V4_0 reads flowpath_toid directly and
+ * needs no synthesis -- or when the GPKG has no `flowpaths` table (logs a WARN unless NGEN_QUIET).
```
hellkite500: The previous doc string uses `(warns unless NGEN_QUIET)` whereas this one says `(logs a WARN unless NGEN_QUIET)`.  These should probably get unified

include/hydrofabric/HydrofabricReaderFactory.hpp
```diff
@@ -0,0 +1,65 @@
...
+ * inspected, yielding the release it describes, or UNRECOGNIZED when it turns out not to be a
+ * hydrofabric at all.
+ *
+ * @param[in] catchment_path Path to the file holding the divides
+ * @param[in] nexus_path Path to the file holding the nexuses; may be the same file
+ * @return HydrofabricVersion identified for the pair, or UNRECOGNIZED if the paths hold
+ *         GeoPackages that are not a hydrofabric
+ * @throws std::runtime_error if either file cannot be opened, if the two are in different formats,
+ *         if a GeoPackage's schema matches no known release, or if they are GeoPackages and
+ *         GeoPackage support was not built in
```
hellkite500, line 31: "GeoPackage support was not built in" should likely name SQLITE, and perhaps even the CMAKE var directly here to be clear that geo package support comes simply from sqlite, and nothing more.

src/hydrofabric/GeoPackageHydrofabricReader.cpp
```diff
@@ -0,0 +1,271 @@
...
+  , nexus_reader_(std::move(nexus_reader))
+  , version_(version)
+{}
+
+const geopackage::GeoPackageReader& GeoPackageHydrofabricReader::reader_for(const std::string& layer) const
+{
+    if (layer == "nexus" && nexus_reader_.has_value()) {
+        return nexus_reader_.value();
+    }
+    return divides_reader_;
```
hellkite500: Should this fall through to divides_reader_ for bad layer names?

src/hydrofabric/GeoPackageHydrofabricReader.cpp
```diff
@@ -0,0 +1,271 @@
...
+void require_column(
+    const geopackage::GeoPackageReader& reader,
+    const std::string& layer,
+    const std::string& column,
+    const std::string& version
+)
+{
+    const std::vector<std::string> columns = reader.db().columns(layer);
+    if (std::find(columns.begin(), columns.end(), column) == columns.end()) {
+        throw std::runtime_error(
```
hellkite500, line 32: This could be a rather large refactor, and isn't stopping for this initial work, but I think a discussion around `throw` and error handling at various levels of the library/integration is worth having at some point.

src/hydrofabric/GeoPackageHydrofabricReader.cpp
```diff
@@ -0,0 +1,271 @@
...
+{
+    return version_;
+}
+
+void GeoPackageHydrofabricReader::check_required_columns(const std::string& /* layer */) const
+{}
+
+geojson::GeoJSON GeoPackageHydrofabricReader::read_divides(const std::vector<std::string>& ids)
+{
+    check_required_columns("divides");
```
hellkite500: Consider the layer names/keys as named `constexpr` values in the respective classes/namespace

src/hydrofabric/GeoPackageHydrofabricReader.cpp
```diff
@@ -0,0 +1,271 @@
...
+    const HydrofabricVersion version
+)
+  : GeoPackageHydrofabricReader(std::move(divides_reader), std::move(nexus_reader), version)
+{}
+
+void V4GeoPackageHydrofabricReader::check_required_columns(const std::string& layer) const
+{
+    if (layer == "nexus") {
+        const geopackage::GeoPackageReader& reader = reader_for(layer);
+        require_column(reader, layer, "nexus_id", "v4");
```
hellkite500: Now I'm getting a little lost on the "v4" "v4.0" and "v4.0beta1" differentiation

src/hydrofabric/HydrofabricReaderFactory.cpp
```diff
@@ -0,0 +1,226 @@
...
+            );
+            #else
+            // Unreachable: detect_hydrofabric() refuses GeoPackage paths in this build.
+            throw std::runtime_error("SQLite3 support required to read GeoPackage files.");
+            #endif
+        case HydrofabricVersion::UNRECOGNIZED:
+            break;
+    }
+
+    throw std::runtime_error(
```
hellkite500, line 215: This error implies geopackage/layer semantics in the more general "make hydrofabric reader"

src/partitionGenerator.cpp
```diff
@@ -395,6 +393,34 @@ void read_arguments(int argc, char* argv[],
     if (error) exit(-1);
 }
 
+// Attach a SentinelFeature named "wb-<label>-<feature id>" as the
```
hellkite500: I think I mentioned this before in the other review, but noting here again that this should probably be a fundamental responsibility of the hy features network class which the the graph is built upon

hellkite500 (commented):
I still need to go through the test code in more detail, but +1 for the `fixture_builders` on that front!

I generally like this over the mixed geopackage/hydrofabric approach.  I think the schema detection complicates the implementations a bit, and hopefully we can get to a point where the geopackage contains version metadata we can rely on in the detection steps.

include/hydrofabric/GeoPackageHydrofabricReader.hpp
hellkite500: See my comments in the review below, this can largely be ignored for now...

PhilMiller:
As another note of modern C++ adoption, consider taking string parameters as [`std::string_view`](https://en.cppreference.com/cpp/string/basic_string_view) rather than `std::string` or `std::string const&`

include/geopackage/GeoPackageReader.hpp
robertbartel: This is at the level of nuance where I often overthink things in C++.  

My initial reaction:  by-value is simpler with negligible costs difference since the type is is move-only.  But conventions, implications from the approach about the design, and other factors related to good practices might come into play here.    

I'm open to changing if you like.  Since you asked rather than suggested, I'll wait for confirmation.

include/hydrofabric/GeoPackageHydrofabricReader.hpp
robertbartel: The convention is to use the `_` as a replacement for `.` in the version substring component of names, and that was applied consistently, at least among the concrete types; i.e., starting with either `V2_2` or `V4_0`.  

I think the problem really is that I didn't effectively differentiate the name of the abstract superclass for the two different v4 readers:  `V4GeoPackageHydrofabricReader`.  I'll change that to `AbstractV4GeoPackageHydrofabricReader`.

test/CMakeLists.txt
robertbartel: Kinda, but this is not inconsistent with other use of _FileChecker_.  It probably does need to be fixed as you suggest, just not in this PR.

src/hydrofabric/CMakeLists.txt
robertbartel: This might come down to:  what is a _hydrofabric_ exactly?  

We have often use the term to refer to the spatial domain being modeled by the data in hydrofabric files, but I think it might really be _that particular_ representation of the spatial domain; i.e., the data and files themselves.  ngen then has its own internal data model for the spatial domain.  The two representations align well, since they both are based on _HY_Features_, but they are not exactly the same thing:  e.g., ngen has no _Divide_ class.  

If they really are different things, then it makes sense to have an ngen `hydrofabric` package scoped to just working with the domain in that representation.  At the moment, I think that is just reading and retrieving the data to provide to ngen's own data model.

include/geopackage/ngen_sqlite.hpp
robertbartel: I don't think so.  I'll make that change.

include/geopackage/GeoPackageReader.hpp
PhilMiller: Ah, ok, I wasn't sure that `sqlite::database` was move-only. Taking it by value is fine. Taking it by r-value reference would just be a re-affirmation of that fact

src/hydrofabric/GeoPackageHydrofabricSchema.cpp
robertbartel: Fair point.  I'll make some adjustments to how that's handled.

include/hydrofabric/GeoPackageHydrofabricReader.hpp
robertbartel: Fair enough.  I'll rename them to `normalize_*` and try to improve the docstring a bit.

include/hydrofabric/GeoPackageHydrofabricSchema.hpp
robertbartel: Sure, I can tweak that.

include/hydrofabric/HydrofabricReaderFactory.hpp
robertbartel: Hmm ... in the code and comments doesn't feel like the right place to reference CMake build vars.  Given that, I don't really think this needs to be changed.  If we ever really needed to make that distinction, wouldn't SQLite support be the wrong one to emphasize here anyway?

src/hydrofabric/GeoPackageHydrofabricReader.cpp
robertbartel: Eh, no, it probably would be better to throw an error here.  I'll change that.

src/hydrofabric/GeoPackageHydrofabricReader.cpp
robertbartel: Specific labels for each version are defined, so we can switch to using those.

src/hydrofabric/HydrofabricReaderFactory.cpp
robertbartel: Hmmm ... so, kind of, but also, in a redundant manner that has the benefit of being more user-friendly.  

Right now, you cannot arrive at that line without the error message being true, but this isn't instantly obvious from the code.  Not sure whether you noticed it or not @hellkite500, but I had to go remind myself.   This suggests to me that having those detail in the message will be valuable.  

I see a few potential directions we could go in to retain this value while cleaning up the design, all of which also imply making appropriate updates to the `identify` function:
1. include another optional `reason` property in _Identified_ that contains a message on why something was identified as `UNRECOGNIZED`, and then use that in place of the current error text
    * Also, probably remove the other throws in `identify` and instead set the `reason` property
    * Thinking about it, if there is an `UNRECOGNIZED` version value, I'm not sure it makes sense to ever have `identify` fail to return
2. Remove the `UNRECOGNIZED` version type entirely, so every bad case throw an error immediately, and just replace this error with a generic future-proofing "no case implemented for type" error
3. Add another _HydrofabricVersion_ value specifically for geopackage files that don't match any known hydrofabric schema, rather than just lumping them in as the generic `UNRECOGNIZED` (we have, after all, recognized something about them)

Thoughts?

src/hydrofabric/GeoPackageHydrofabricReader.cpp
robertbartel: Sure, I can change these.

src/partitionGenerator.cpp
robertbartel: Sure, I think that's fairly easy to move.

robertbartel:
@hellkite500 and @PhilMiller, I think I've addressed all your comments, most with changes, though some wanting more discussion.  When you can, please go through and assess what conversations can be marked resolved, and make comments on any others.

One note: I didn't got back and update things to use `string_view`, Phil.  I can, but it seems like the type of thing you were ok leaving out for now, and also something maybe worth a dedicated effort later (along with some other modernization).

include/core/network.hpp
```diff
@@ -388,6 +388,21 @@ namespace network {
...
 
     };
+
+    /**
+     * @brief Attach a SentinelFeature downstream of every terminal feature of a given type.
+     *
+     * Every non-sentinel feature in @p collection whose id type prefix satisfies @p matches_type
+     * and that has no destination feature yet gains a destination SentinelFeature named
+     * "wb-<label>-<feature id>", and the new sentinels are added to @p collection. The "wb-"
+     * prefix is what makes a sentinel partition as a catchment.
```
PhilMiller, line 398: There's a typo, but I'm not 100% sure on the correction

```suggestion
     * "wb-<label>-<feature id>", and the new sentinels are added to @p collection. The "wb-"
     * prefix is what marks a sentinel partition as a catchment.
```

src/hydrofabric/GeoPackageHydrofabricSchema.cpp
```diff
@@ -0,0 +1,171 @@
...
+    joined << "[";
+    bool first = true;
+    for (const std::string& column : columns) {
+        if (!first) joined << ", ";
+        joined << column;
+        first = false;
+    }
+    joined << "]";
+    return joined.str();
+}
```
PhilMiller, line 32:
```
src/partitionGenerator.cpp:#include <boost/algorithm/string/join.hpp>
src/partitionGenerator.cpp:    using boost::algorithm::join;
```
There's a library routine we already use that does this.

src/hydrofabric/GeoPackageHydrofabricSchema.cpp
```diff
@@ -0,0 +1,171 @@
...
+    const std::set<std::string>& nexus_columns,
+    const std::set<std::string>& divides_columns)
+{
+    if (nexus_columns.empty()) {
+        throw std::runtime_error(
+            "hydrofabric detect_version: no nexus table"
+        );
+    }
+
+    if (nexus_columns.count(NEXUS_ID_COLUMN) > 0) {
```
PhilMiller, line 46:
```suggestion
    if (nexus_columns.contains(NEXUS_ID_COLUMN)) {
```
https://en.cppreference.com/cpp/container/set/contains

test/hydrofabric/HydrofabricReaderFactory_Test.cpp
```diff
@@ -0,0 +1,274 @@
...
+#include "HydrofabricReaderFactory.hpp"
+#include "FileChecker.h"
+#include "fixture_builders.hpp"
+
+namespace fixtures = ngen::hydrofabric::fixtures;
+
+namespace {
+
+//! Copy a file byte for byte, so a fixture can be given a name that says nothing about its format.
+std::string copy_as(const std::string& source, const std::string& name)
```
PhilMiller, line 17: Is this not just `std::filesystem::copy(source, name, std::filesystem::copy_options::overwrite_existing)`?

test/hydrofabric/HydrofabricReaderFactory_Test.cpp
```diff
@@ -0,0 +1,274 @@
...
+    } catch (const std::runtime_error& e) {
+        EXPECT_NE(std::string(e.what()).find("one format"), std::string::npos)
+            << "exception message: " << e.what();
+    }
+}
+
+// Format comes from the file's own header, so these cover the names an extension test got wrong.
+
+// The case that matters most, because ngen produces it: loading a subdivided hydrofabric under MPI
+// appends a rank suffix to both data paths, leaving a GeoPackage named `....gpkg.0`.
```
PhilMiller, line 217: This seems to be describing a shortcoming of ngen, but it's good to know that's handled.

PhilMiller:
Unless I'm missing something about the bigger context in which this change is to be made, I'd be in favor of merging this in whole.

